PR #1, opened 2023-02-17
GLM prototype with simulation

ahwillia:


billbrod:
Notes/questions for basis:
1. Need reference for `RaisedCosineBasis`
2. Do we need one for `OrthExponentials`?
3. Is my docstring for `Basis.transform` correct? I think we need more details about `x`, it's not clear to me when users would set it.
4. Do we want to call this `transform` (rather than, `generate_basis_funcs`)? I know the goal is to match scikit learn API, but it's not really transforming any data -- it's most often called without supplying the argument, and even if it *is* supplied, I don't think it's data to be transformed, more like sample points for the bases, no? (that way we could also change `x` to something more descriptive)
5. Any restriction on values for `x` in `RaisedCosineBasis.transform (like the other two)? 
6. Need a better description of `order` for `MSpline`

billbrod:
GLM: 
- why does `score` ignore the last time point of the prediction?
- I tried to get all the functions using the same code for predicting / scoring (to make it easier to change or extend, as necessary), but simulate uses a slightly different einsum. I thought converting from `einsum('nbt,nbj -> nt')` to `einsum('nbt,nbj -> n')` would just require a `sum()`, but it doesn't look like it (either I forget what einsum does or it's a precision issue). is there a way to combine here, or will we just need to pay attention and remember to change one whenever we change the other.
- What's the difference between `simulate` and `predict`? docstrings should be clear as to when users want one vs the other

neurostatslib/glm.py
```diff
@@ -1,116 +1,324 @@
...
-        subkeys = jax.random.split(random_key, num=num_timesteps)
+        subkeys = jax.random.split(random_key, num=n_timesteps)
 
         def scan_fn(spikes, key):
-            X = convolve_1d_basis(B, spikes)
-            # X.shape == (num_neurons x num_basis_funcs x 1)
-            log_pred = jnp.einsum("nb,nbj->n", jnp.squeeze(X), Ws)
-            new_spikes = jax.random.poisson(key, jnp.exp(log_pred))
+            # (n_neurons, n_basis_funcs, n_timebins)
+            X = convolve_1d_basis(
```
ahwillia, line 314: Could this be replaced by `self.predict(...)`

billbrod:
Couple questions on checks for basis functions: 
- should the input `sample_pts` (to `Basis.gen_basis_funcs`, previously `transform`) have to be of length `window_size`? Currently, I raise a warning if not, but can change that to an exception
- RaisedCosineBasis raises an exception if its bases don't uniformly tile the space or if any of them are all zeros, which happens if its `sample_pts`  was generated with the wrong `n_basis_funcs` -- that's what we want, right?

ahwillia:
LGTM!

PR #15, opened 2023-06-02
Changes to SplineBasis

BalzaniEdoardo:
Here I edited the class MSplineBasis. In particular,

- I created a class SplineBasis with __init__ and a method to generate knots
    -  the classes MSplineBasis, BSplineBasis and Cyclic_BSplineBasis inherits SplineBasis but have different "gen_basis_funcs" methods

- Added a few test for the B-splines, making sure they can be defined for multiple params configurations, and that they respect some assumptions

### DIFF
--- a/neurostatslib/basis.py
+++ b/neurostatslib/basis.py
@@ -214,30 +214,7 @@ class OrthExponentialBasis(Basis):
                      axis=1)
         ).T
 
-
-class MSplineBasis(Basis):
-    """M-spline 1-dimensional basis functions.
-
-    Parameters
-    ----------
-    n_basis_funcs
-        Number of basis functions.
-    window_size
-        Size of basis functions.
-    order
-        Order of the splines used in basis functions. Must lie within [1,
-        n_basis_funcs]. The m-splines have ``order-2`` continuous derivatives
-        at each interior knot. The higher this number, the smoother the basis
-        representation will be.
-
-
-    References
-    ----------
-    .. [1] Ramsay, J. O. (1988). Monotone regression splines in action.
-       Statistical science, 3(4), 425-441.
-
-    """
-
+class SplineBasis(Basis):
     def __init__(self, n_basis_funcs: int, window_size:int,
                  order: int = 2):
         super().__init__(n_basis_funcs, window_size, (0, window_size))
@@ -270,7 +247,7 @@ class MSplineBasis(Basis):
                 "than `n_basis_funcs` parameter."
             )
 
-        assert (perc_low >= 0)  & (perc_high <= 1), "Specify low and high percentile (perc_low, perc_high) as float between 0 and 1"
+        assert (perc_low >= 0) & (perc_high <= 1), "Specify low and high percentile (perc_low, perc_high) as float between 0 and 1"
         assert (perc_low < perc_high), "perc_low must be < perc_high. "
 
         # clip to avoid numerical errors in case of percentile numerical precision close to 0 and 1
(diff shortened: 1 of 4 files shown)


prototests/test_splineBasis.py
```diff
@@ -0,0 +1,63 @@
+import pytest
+from neurostatslib.basis import BSplineBasis, Cyclic_BSplineBasis
+import numpy as np
+
+np.random.seed(10**5)
```
ahwillia, line 5: I like to feed the seed into the test function like

```python
@pytest.mark.parametrize('seed', [111, 222, 333])
def test_me(seed):
    rng = np.random.RandomState(seed)
    random_numbers = rng.randn(100)
```

prototests/test_splineBasis.py
```diff
@@ -0,0 +1,63 @@
...
+
+par_grid = []
+for n_basis in range(1, 10):
+    for order in range(1, n_basis):
+        for window_size in [11,100,1000]:
+            par_grid.append({'n_basis':n_basis, 'order':order, 'window_size':window_size})
+samp = np.random.normal(size=1000)
```
ahwillia, line 12: I think there should be a way to re-format this so that all the code execution happens inside test functions. Having this executed outside a test function seems a bit messy to me.

prototests/test_splineBasis.py
```diff
@@ -0,0 +1,63 @@
...
+    assert (all( np.abs(eval_basis[:, sel].sum(axis=0)) == 0))
+
+
+par_grid = []
+for n_basis in range(4, 12):
+    for order in range(2, 6):
+        if (n_basis < 2*(order-1)) or (n_basis < order + 2):
+            continue
+        for window_size in [11,100,1000]:
+            par_grid.append({'n_basis':n_basis, 'order':order, 'window_size':window_size})
```
ahwillia, line 45: See comment above about reformatting so that everything happens inside a test function

neurostatslib/basis.py
```diff
@@ -281,14 +303,201 @@ def gen_basis_funcs(self, sample_pts: NDArray) -> NDArray:
...
+            basis_eval[i-reps, in_sample] = splev(sample_pts[in_sample], (knots, id_basis[i], self.order - 1), der=der)
+
+        # # check sum equal 1 (B-spline are supposed to sum to 1)
+        # assert(np.abs(basis_eval.sum(axis=0) - 1).max() < 1e-6)
+        return basis_eval
+
+
+
+class Cyclic_BSplineBasis(BSplineBasis):
+    """Cyclic M-spline 1-dimensional basis functions.
```
ahwillia, line 415: Docstring says `M-spline` here but we are implementing B-splines

neurostatslib/basis.py
```diff
@@ -281,14 +303,201 @@ def gen_basis_funcs(self, sample_pts: NDArray) -> NDArray:
...
+
+    Parameters
+    ----------
+    n_basis_funcs
+        Number of basis functions.
+    window_size
+        Size of basis functions.
+    order
+        Order of the splines used in basis functions. Must lie within [1,
+        n_basis_funcs]. The m-splines have ``order-2`` continuous derivatives
```
ahwillia, line 425: Again, docstring says `m-spline` here but should be `b-spline`

neurostatslib/basis.py
```diff
@@ -281,14 +303,201 @@ def gen_basis_funcs(self, sample_pts: NDArray) -> NDArray:
...
```
ahwillia, line 332: `m-spline` in the docstring should be `b-spline`

ahwillia:
Nice work. Right now it adds a decent number of lines to the codebase. Do you think it might be possible to re-use some of the common code between the cyclic and non-cyclic B-spline classes? For example, make them sub-classes of an abstract BSpline class? If its not possible no worries, I just want the implementation to be as simple as possible from the standpoint of long-term maintenance.

prototests/test_splineBasis.py
BalzaniEdoardo: I changed completely how the test is set up, let me know if you like this way of structuring it more

prototests/test_splineBasis.py
BalzaniEdoardo: this is happening inside a test class now

prototests/test_splineBasis.py
BalzaniEdoardo: same thing

PR #17, opened 2023-06-09
Development: restructure the repository

BalzaniEdoardo:
This PR aims to reconfigure the repository's folder structure and introduce a basic CI that runs tests and linters.

Changes Made:
- Added pyproject.toml for installing.
- Removed setup.py.
- Added tox.ini file for facilitating testing/linting on multiple OS.
- Added .github/workflows/ci.yml for basic CI.
- Added .github/workflows/connect.yml to SSH to GitHub runners for debugging when manually triggered.
- Set all of the developers as code owners (globally for now).
- Moved the code to src/neurostatslib.

The basis.py code will be substantially rewritten, so the differences between this development version and the main are not important. Once this PR goes through, I will start following the git-flow, creating a branch from development that will be used to restructure the basis function class.


.github/CODEOWNERS
```diff
@@ -0,0 +1 @@
+@BalzaniEdoardo @billbrod @ahwillia @gviejo
```
billbrod, line 1: Confused about this codeowners file: my understanding is that these files should be structured like `path @user`, where `user` is then the owner for every file at `path` (as in [the docs](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-code-owners)). Thus, I was expected this to be `* @BalzaniEdoardo`, meaning Edoardo needs to approve every PR to this repo. I think there's something weird here where it thinks Edoardo is the path. Also, in general, I thought we had agreed to have one owner per repo (else, we'd all have to review every PR), potentially making different people "subowners" (e.g., making me the owner of `ci.yml` or `.readthedocs.yml` or Guillaume the owner of `tox.ini`)

.github/workflows/ci.yml
```diff
@@ -0,0 +1,51 @@
+name: ci
+
+on:
+#  push:
+#    branches:
+#      - folder_struct_and_setup
```
billbrod, line 6: delete these if not using them

.github/workflows/connect.yml
```diff
@@ -0,0 +1,27 @@
+name: connect
```
billbrod, line 1: add a comment here to describe what this workflow is for and when you'd use it? (same for `ci.yml`)

src/neurostatslib/basis.py
```diff
@@ -0,0 +1,621 @@
+"""Bases classes.
```
billbrod, line 1: why is this file double the length of the old `basis.py`? looks like it contains things that are in the `edoardo` branch -- the new spline bases. those should be removed so that we keep the two changes separate.

src/neurostatslib/glm.py
```diff
@@ -1,13 +1,17 @@
+"""GLM core module
```
billbrod, line 1: are the only differences here the result of running the linters?

src/neurostatslib/utils.py
```diff
@@ -1,26 +1,29 @@
+"""Utility functions for data pre-processing
```
billbrod, line 1: are the only changes here the result of linters?

tests/test_BSpline.py
```diff
@@ -0,0 +1,74 @@
+# content of ./test_parametrize.py
+import pytest
```
billbrod, line 2: should be removed from this PR for now

tests/test_glm_synthetic.py
```diff
@@ -49,26 +49,26 @@
...
 fit_pred = fitted_model.predict(spike_data)
+#
+# fig, axes = plt.subplots(2, 1)
+# axes[0].plot(onp.arange(nt), spike_data[0])
+# axes[0].plot(onp.arange(ws, nt + 1), sim_pred[0])
+# axes[0].plot(onp.arange(ws, nt + 1), fit_pred[0])
+# axes[1].plot(onp.arange(nt), spike_data[1])
+# axes[1].plot(onp.arange(ws, nt + 1), sim_pred[1])
+# axes[1].plot(onp.arange(ws, nt + 1), fit_pred[1])
+# plt.show()
```
billbrod, line 60: why are these lines commented out?

tests/test_glm_synthetic.py
```diff
@@ -16,7 +16,7 @@
     n_basis_funcs=5,
     window_size=ws
 )
-sim_pts = nsl.sample_points.raised_cosine_log(5, ws)
+sim_pts = jnp.clip(neurostatslib.sample_points.raised_cosine_log(5, ws),0,100)
```
billbrod, line 19: this shouldn't be necessary with the change in `raised_cosine_log`, right?

tests/test_glm_synthetic_single_neuron.py
```diff
@@ -15,7 +15,7 @@
     n_basis_funcs=5,
     window_size=ws
 )
-sim_pts = nsl.sample_points.raised_cosine_log(5, ws)
+sim_pts = jnp.clip(neurostatslib.sample_points.raised_cosine_log(5, ws),0,100)
```
billbrod, line 18: same comments as `test_glm_synthetic`:
 - why are the lines commented out?
 - this clip shouldn't be necessary, right?

tests/test_glm_synthetic_single_neuron.py
```diff
@@ -1,12 +1,12 @@
 import jax
 import jax.numpy as jnp
 import numpy as onp
-import neurostatslib as nsl
 from neurostatslib.glm import GLM
 from neurostatslib.basis import RaisedCosineBasis
 import matplotlib.pyplot as plt
 jax.config.update("jax_platform_name", "cpu")
 jax.config.update("jax_enable_x64", True)
+import neurostatslib
```
billbrod, line 9: what's the matter with `import neurostatslib as nsl`?

tests/test_glm_synthetic_single_neuron.py
```diff
@@ -1,12 +1,12 @@
 import jax
```
billbrod, line 1: also, this script is basically the same as `test_glm_synthetic.py`. I'm thinking maybe we combine them: make a function that runs this, with an option to set the number of neurons to include. but that doesn't belong in this PR

billbrod (changes requested):
In addition to my comments, I don't understand the structure of `pyproject.toml` and `tox.ini`. If we can document why we included the sections we did and what they mean in the template repo, I'm fine, but just want to highlight the same issues here.

.github/CODEOWNERS
BalzaniEdoardo: with this syntax, I am setting up everyone as code owner. I suggested in the comments that we can make it more specific later but since we did not discuss the specifics yet I went for this temporary solution. If you prefer I can set myself as the only owner for now or add you guys as specific script owner as you are suggesting

src/neurostatslib/basis.py
BalzaniEdoardo: I mention this in the PR request comment, the basis class will be rebuild from scratch, so ignore any diffs. (this version works, doesn't brake the code but still will be re-structured)

src/neurostatslib/glm.py
BalzaniEdoardo: yes

tests/test_glm_synthetic.py
BalzaniEdoardo: all this test code will need to be re-written. I'll move visualization scripts into the scripts we will use for the documentation

tests/test_glm_synthetic.py
BalzaniEdoardo: now yes, anyways, this code which is more for visualization will be moved to a dedicated documentation script.

tests/test_glm_synthetic_single_neuron.py
BalzaniEdoardo: as I said before, this will be moved

tests/test_glm_synthetic_single_neuron.py
BalzaniEdoardo: sounds good

src/neurostatslib/basis.py
BalzaniEdoardo: reverted to the main and created a specific branch to tackle the basis function issue

tests/test_glm_synthetic.py
BalzaniEdoardo: reverted tests to the main version. tests will be reformatted when the functions that they are pointing to will be modified

tests/test_glm_synthetic_single_neuron.py
BalzaniEdoardo: for now I kept it consistent with the main

tests/test_glm_synthetic_single_neuron.py
BalzaniEdoardo: same as above, I kept it consistent with jax

.github/CODEOWNERS
BalzaniEdoardo: edited, for now I set myself as codeowner, we can be more specific  once we agree on who does what on this repo

.github/workflows/connect.yml
BalzaniEdoardo: done

PR #18, opened 2023-06-13
Restructuring repo

BalzaniEdoardo:
this PR aims to conform the repo structure to the ccn-template


.github/workflows/ci.yml
```diff
@@ -0,0 +1,48 @@
...
+     branches:
+       - main
+       - development
+
+jobs:
+  tox:
+    strategy:
+      matrix:
+        os: [ubuntu-latest] #[ubuntu-latest, macos-latest, windows-latest]
+        python-version: ['3.10'] #['3.8', '3.9', '3.10']
```
billbrod, line 17: just curious: we'll want to test multiple versions and OS eventually -- are you thinking we just wait to do that until a bit further along?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,48 @@
...
+    runs-on: ${{ matrix.os }}
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
+        with:
+          ref: ${{ github.ref }}
```
billbrod, line 23: I'm not sure this is actually necessary -- I don't have this in `plenoptic`, and the branch of the PR is checked out each time. It's unclear to me what it does

.github/workflows/connect.yml
```diff
@@ -0,0 +1,31 @@
...
+    steps:
+      - name: Checkout                 # Step to checkout the repository
+        uses: actions/checkout@v3
+        with:
+          ref: ${{ github.ref }}       # Use the current Git reference
+
+      - name: Set up Python            # Step to set up Python
+        uses: actions/setup-python@v4  # Use v4 for compatibility with pyproject.toml
+        with:
+          python-version: '3.10'       # Specify the Python version
```
billbrod, line 22: it's totally possible that we have an issue with only a single version of python, so probably want to have the same build matrix as `ci.yml`

CCN-letterFoot.png
billbrod, line 1: I know that the images are in the repo root in our template, but I feel like they should be in `docs/assets` or something. Don't want to clutter up the root too much...

about.md
billbrod, line 1: what is supposed to be in this file? it's empty in our template as well.

docs/examples/README.md
billbrod, line 1: I don't think the examples / stubs belong in this repo -- we can copy them as starters from template when we want to.

docs/gen_ref_pages.py
billbrod, line 1: related to conversation in [ccn template PR](https://github.com/flatironinstitute/ccn-template/pull/6), it's not clear to me when this is necessary. is this useful now?

index.md
billbrod, line 1: I think this is supposed to be in `docs/`, right? and probably shouldn't be empty -- should at least be a placeholder saying docs are under development

.readthedocs.yaml
billbrod, line 1: Is the link with readthedocs working? I don't see it in the status checks -- I can set it up, if you'd like, or you can try and practice.

mkdocs.yml
billbrod, line 1: This currently has some issues: `index.md`, `CONTRIBUTING.md` and `about.md` are all outside the `docs/` structure (this is my problem, if you copied this directly from my ccn template PR).

I can fix this if you'd like

tests/test_glm_runs.py
```diff
@@ -0,0 +1,61 @@
...
+import numpy as np
+
+class DimensionMismatchError(Exception):
+    """Exception raised for dimension mismatch errors."""
+
+    def __init__(self, message):
+        self.message = message
+        super().__init__(self.message)
+
+def test_setup_Mbasis():
```
billbrod, line 13:
```suggestion
def test_setup_MSplineBasis():
```

tests/test_glm_runs.py
```diff
@@ -0,0 +1,61 @@
...
+    """
+    Minimal test for MBasis definition.
```
billbrod, line 15:
```suggestion
    Minimal test for MSplineBasis definition.
```

tests/test_glm_runs.py
```diff
@@ -0,0 +1,61 @@
...
+        if spike_basis_matrix.shape[0] != n_basis:
+            raise DimensionMismatchError(f"The output basis matrix has {spike_basis_matrix.shape[1]} time points, while the number of basis specified is {n_basis}. They must agree.")
+
+
+        if spike_basis_matrix.shape[1] != window:
+            raise DimensionMismatchError(f"The output basis basis matrix has {spike_basis_matrix.shape[1]} window size, while the window size specified is {window}. They must agree.")
+
+    return
+
+if __name__ == "__main__":
```
billbrod, line 45: I'm confused about this ifmain block -- my understanding of `pytest` is that it runs all the `test_` functions within a `test_*py` script, so what's this doing?

tests/test_glm_synthetic.py
```diff
@@ -0,0 +1,110 @@
...
+
+    W = onp.empty((2, 5, 2))
+    for i, j in itertools.product(range(nn), range(nn)):
+        W[i, :, j] = w0 if (i == j) else w1
+
+    simulated_model = GLM(B)
+
+    return
+
+if __name__ == '__main__':
```
billbrod, line 44: same quesiton about ifmain as the other test file

tests/test_glm_runs.py
billbrod: Specifically, when I run `pytest`, two tests are run, one from each file. And it's not the ifmain block

tests/test_glm_synthetic_single_neuron.py
```diff
@@ -0,0 +1,64 @@
+import jax
+import jax.numpy as jnp
+import numpy as onp
+import neurostatslib as nsl
+from neurostatslib.glm import GLM
+from neurostatslib.basis import RaisedCosineBasis
+import matplotlib.pyplot as plt
+
+if __name__ == '__main__':
```
billbrod, line 9: this does not get run by pytest, same thing as above

tox.ini
```diff
@@ -0,0 +1,55 @@
+[tox]
+isolated_build = True
+envlist = py38, py39, py310
+
+[testenv]
+deps = -e .
```
billbrod, line 6: I'm still confused about the  relationship between this and pyproject.toml's `[dev]` optional dependency -- maybe your new comments will clarify. But are both necessary?

billbrod (changes requested):
Variety of questions throughout. Bigger picture:
- I'm assuming the differences in the `.py` files moved into `src/` are because of the linters? and thus ignoring them.
- is the link with readthedocs working?
- I can fix the docs and tests issues, if you'd like

tox.ini
billbrod: Answering myself: no, they're not. We'll use `extras = dev` to make sure tox uses the correct optional dependencies.

docs/index.md
```diff
@@ -0,0 +1,7 @@
+# neurostatslib 
+A toolbox of statistical analysis for neuroscience. 
+
+## Disclaimer
+Please note that this package is currently under development. While you can download and test the functionalities that are already present, it's important to be aware that the code stability and systematic testing cannot be guaranteed at this stage. 
+
+[README.md](..%2FREADME.md)
```
billbrod, line 7: This, unfortunately doesn't work as a way of linking to the readme. See some of the discussion [here](https://github.com/mkdocs/mkdocs/discussions/3062) for why, but basically because README isn't included in the `docs/` folder. So, unfortunately, I think a direct url is the way to do this:
```suggestion
See our [README](https://github.com/flatironinstitute/generalized-linear-models/blob/main/README.md) for more info.
```

We're discussing the relationship between docs/index and readme in [this issue](https://github.com/flatironinstitute/ccn-template/issues/16)

billbrod (changes requested):
All these changes look good to me, but some of my older comments are still unaddressed.

docs/gen_ref_pages.py
billbrod: Okay, understand this now -- I agree it's necessary, but let's add a link to the ccn template notes page for more info?

.readthedocs.yaml
billbrod: I see it now

CCN-letterFoot.png
billbrod: Can discuss this more in [ccn template](https://github.com/flatironinstitute/ccn-template/issues/17)

.github/workflows/ci.yml
BalzaniEdoardo: yes, exactly. for now, since it's early days I am testing on single os

.github/workflows/ci.yml
BalzaniEdoardo: after digging into this it seems that ref: ${{ github.ref }} is of no use with this action; it would behave in the same way if I omit it. I am removing it

.github/workflows/connect.yml
BalzaniEdoardo: I'll set the ci to run on multiple python versions

tests/test_glm_runs.py
BalzaniEdoardo: turned everything in a def test_ so that it is run as a test.
I had to comment the plotting otherwise pytest would not terminate until figures were closed

docs/index.md
BalzaniEdoardo: thanks for this

tests/test_glm_runs.py
```diff
@@ -58,4 +57,5 @@ def test_setup_Mbasis():
     model.fit(spike_data)
     model.predict(spike_data)
     key, subkey = jax.random.split(key)
-    X = model.simulate(subkey, 20, spike_data[:, :100])
\ No newline at end of file
+    X = model.simulate(subkey, 20, spike_data[:, :100])
+    return
```
billbrod, line 61: is return necessary for tests?

tests/test_glm_runs.py
billbrod: test name should still be changed

tests/test_glm_synthetic.py
```diff
@@ -86,25 +86,25 @@ def test_set_up_glm():
...
+    #         B.T @ simulated_model.spike_basis_coeff_[i, :, j],
+    #         label="true"
+    #     )
+    #     axes[i, j].plot(
+    #         B.T @ fitted_model.spike_basis_coeff_[i, :, j],
+    #         label="est"
+    #     )
+    #     axes[i, j].axhline(0, dashes=[2, 2], color='k')
+    # axes[-1, -1].legend()
+    # plt.show()
```
billbrod, line 110: what are these commented out lines? they should either be in a test, deleted, or moved into an issue if it's something we'll eventually want but aren't sure why now.

commented out things get lost and forgotten

tests/test_glm_synthetic_single_neuron.py
```diff
@@ -44,21 +44,21 @@
...
+    #     B.T @ simulated_model.spike_basis_coeff_[0, :, 0],
+    #     label="true"
+    # )
+    # ax.plot(
+    #     B.T @ fitted_model.spike_basis_coeff_[0, :, 0],
+    #     label="est"
+    # )
+    # ax.axhline(0, dashes=[2, 2], color='k')
+    # ax.legend()
+    # plt.show()
```
billbrod, line 64: same comment about commented out lines

docs/gen_ref_pages.py
```diff
@@ -0,0 +1,31 @@
+"""Generate the code reference pages and navigation."""
```
billbrod, line 1:
```suggestion
"""Generate the code reference pages and navigation.

See [CCN template repo](https://ccn-template.readthedocs.io/en/latest/notes/03-documentation/) for why.
"""
```

tests/test_glm_runs.py
billbrod: Ah, yeah the way to handle this is just to call `plt.close('all')` at the end of the test, [e.g.](https://github.com/LabForComputationalVision/plenoptic/blob/main/tests/test_display.py#L102)

billbrod (changes requested):
Almost there! Just a handful of small changes.

tests/test_glm_synthetic.py
billbrod: see earlier comment for plots in tests: I think the problem is `plt.show()`. that does not belong in a test and instead, tests should end with `plt.close('all')`

in this case, I don't think the plots are actually necessary for the tests. but it might be worth including at least one to have a working example for later on?

tests/test_glm_runs.py
BalzaniEdoardo: locally, it doesn't seem to move to the next line (basically it halts the execution and never gets to plt.close).
I figured out a solution: you need to set 'Agg' as the backend, which doesn't start an event loop like qt does

tests/test_glm_synthetic.py
BalzaniEdoardo: fixed it changing beckend for matplotlib, removed the commented parts

tests/test_glm_synthetic_single_neuron.py
BalzaniEdoardo: as above

tests/test_glm_runs.py
BalzaniEdoardo: nope, I'll remove it (but generally I like a return statement, it makes it clear when the function ends, especially if your last line of code is more indented then the function/method body, like if you have some if statement)

tests/test_glm_synthetic_single_neuron.py
```diff
@@ -30,35 +36,30 @@ def test_glm_fit():
...
-    # ax.axhline(0, dashes=[2, 2], color='k')
-    # ax.legend()
-    # plt.show()
\ No newline at end of file
+    fig, ax = plt.subplots(1, 1)
+    ax.plot(onp.arange(nt), spike_data[0])
+    ax.plot(onp.arange(ws, nt + 1), sim_pred[0])
+    ax.plot(onp.arange(ws, nt + 1), fit_pred[0])
+    
+    #plt.close('all')
```
billbrod:
```suggestion
```

tests/test_glm_synthetic_single_neuron.py
```diff
@@ -30,35 +36,30 @@ def test_glm_fit():
...
+        label="true"
+    )
+    ax.plot(
+        B.T @ fitted_model.spike_basis_coeff_[0, :, 0],
+        label="est"
+    )
+    ax.axhline(0, dashes=[2, 2], color='k')
+    ax.legend()
+    
+    #plt.close('all')
```
billbrod:
```suggestion
    plt.close('all')
```

billbrod (approved):
looks good!

PR #23, opened 2023-06-30
Restructuring of the basis module

BalzaniEdoardo:
Hi everyone!
Here a couple of notes on the PR

## Summary of the content
I was able to start the PR for the basis module restructuring just in time.
The PR contents are quite a lot:
1. `neurostatslib.basis' module have been completely restructured, while it contains the same basis function that are in main (RaisedCosineBasisLog/Linear, OrthExponentialBasis, MSplineBasis)
2. testing the basis module in `test/test_basis.py` (basis class instantiation, expected input/output, and checked that exception raised when the input did not conform the method requirements, test that all the non-abstract classes implement the abstract methods).
3. Added **Developer Notes** to the documentation with an introduction page and a note on the `basis` module
4. Added a couple of example scripts in 'docs/examples' that show the user how to define, combine and plot the basis functions
5. Added a plugin that renders Latex in markdown (I wrote a couple of equation in the example scripts.

## Additional comments
It looks like mkdocstrings does not fully support the Numpy docstrings format yet, see [here](https://mkdocstrings.github.io/python/usage/docstrings/numpy/?h=numpy). I don't like the rendering of the documentation as it is now. I suggest to change to one of the supported style now that the number of docstrings is relatively small.


docs/developers_notes/basis_module.md
```diff
@@ -0,0 +1,130 @@
...
+
+### The Public Method `evaluate`
+
+The `evaluate` method checks input consistency and evaluates the basis function at some sample points. It accepts one or more numpy arrays as input, which represent the sample points at which the basis will be evaluated, and performs the following steps:
+
+1. Checks that the inputs all have the same sample size `N`, and raises a `ValueError` if this is not the case.
+2. Checks that the number of inputs matches what the basis being evaluated expects (e.g., one input for a 1-D basis, N inputs for an N-D basis, or the sum of N 1-D bases), and raises a `ValueError` if this is not the case.
+3. Checks that the output of the method does not exceed `gb_limit` GB of memory size. This size limit is specified at class initialization.
+4. Calls the `_evaluate` method on the input, which is the subclass-specific implementation of the basis set evaluation.
+5. Returns a numpy array of dimension `N x number of bases`, with each basis element evaluated at the samples.
```
billbrod, line 47:
```suggestion
5. Returns a numpy array of shape `(N, n_basis_funcs)`, with each basis element evaluated at the samples.
```

two changes: use numpy shape convention (clearer than x), and change to `n_basis_funcs`, which is an argument we accept. 

Feels a bit weird to mix "mathy" (`N`) and "computer-sciencey" (`n_basis_funcs`) conventions, I'd vote for being consistently computer-sciencey, except in equations.

docs/developers_notes/basis_module.md
```diff
@@ -0,0 +1,130 @@
...
+4. Calls the subclass-specific `_evaluate` method.
+5. Returns both the sample grid points and the evaluation output at each grid point.
+
+### Abstract Methods
+
+The `neurostatslib.basis.Basis` class has the following abstract methods, which every non-abstract subclass must implement:
+
+1. `_evaluate`: Evaluates a basis over some specified samples.
+2. `_get_samples`: Returns a tuple of equidistant samples (as flat numpy arrays) for each dimension of the basis function. These samples are used to form a grid over which the basis is evaluated.
+3. `_check_n_basis_min`: Checks the minimum number of basis functions required. This requirement can be specific to the type of basis.
```
billbrod, line 67: Looks like for many of the subclasses, this just checks that the value is positive. Is it worth implementing this in the superclass and telling developers they can overwrite it themselves?

docs/developers_notes/basis_module.md
```diff
@@ -0,0 +1,130 @@
...
+
+
+## The `AdditiveBasis` Class
+
+This class extends the abstract `Basis` class and represents a set of basis functions that are combined additively. It overrides the `_evaluate`, `_get_samples`, and `_check_n_basis_min` methods with specific implementations for additive basis functions.
```
billbrod, line 72: needs a longer description of what it means to combine bases additively or a link to an example

docs/developers_notes/basis_module.md
```diff
@@ -0,0 +1,130 @@
...
+
+### Overridden Methods
+
+1. `_evaluate`: For an `AdditiveBasis` object, this method should evaluate each basis function at the specified samples and then add the results together.
+2. `_get_samples`: This method should return equidistant samples over the domain of each basis function. The samples should then be summed together.
+3. `_check_n_basis_min`: This method checks whether the number of basis functions meets the minimum requirement for an additive basis set.
+
+## The `MultiplicativeBasis` Class
+
+This class extends the `Basis` abstract class and represents a set of basis functions that are combined multiplicatively. It overrides the `_evaluate`, `_get_samples`, and `_check_n_basis_min` methods with specific implementations for multiplicative basis functions.
```
billbrod, line 82: needs a longer description of what it means to combine bases multiplicatively or a link to an example

docs/developers_notes/basis_module.md
```diff
@@ -0,0 +1,130 @@
...
+
+### Overridden Methods
+
+1. `_evaluate`: For a `MultiplicativeBasis` object, this method should evaluate each basis function at the specified samples and then multiply the results together.
+2. `_get_samples`: This method should return equidistant samples over the domain of each basis function. The samples should then be multiplied together.
+3. `_check_n_basis_min`: This method checks whether the number of basis functions meets the minimum requirement for a multiplicative basis set.
+
+## The `SplineBasis` and `MSplineBasis` Classes
+
+The `SplineBasis` class is an abstract subclass of `Basis` that represents a spline basis function. The `MSplineBasis` class is a concrete subclass of `SplineBasis` that implements a specific type of spline basis, the M-spline.
```
billbrod, line 92: needs a longer description or a link to what a "spline basis function" means. and what's shared among spline bases that makes it worth having their own subclass.

docs/developers_notes/basis_module.md
```diff
@@ -0,0 +1,130 @@
...
+
+### Overridden Methods for `MSplineBasis`
```
billbrod, line 94: does SplineBasis do anything? should be described

docs/developers_notes/basis_module.md
```diff
@@ -0,0 +1,130 @@
...
+
+1. `_evaluate`: For an `MSplineBasis` object, this method should evaluate the M-spline at the specified samples.
+2. `_get_samples`: This method should return equidistant samples over the domain of the M-spline.
+3. `_check_n_basis_min`: This method checks whether the number of M-spline basis functions meets the minimum requirement.
+
+## The `RaisedCosineBasis`, `RaisedCosineBasisLinear`, and `RaisedCosineBasisLog` Classes
+
+The `RaisedCosineBasis` class is an abstract subclass of `Basis` that represents a raised cosine basis function. The `RaisedCosineBasisLinear` and `RaisedCosineBasisLog` classes are concrete subclasses of `RaisedCosineBasis` that implement specific types of raised cosine basis functions.
```
billbrod, line 102: needs a longer description or a link to what a "raised cosine basis function" means (with reference, I think). Also, what's shared among them that makes it worth having their own subclass, and how they differ.

E.g., "The two concrete subclasses have the same `_evaluate` function but differ in their `_get_samples` method"

docs/developers_notes/basis_module.md
```diff
@@ -0,0 +1,130 @@
...
+
+### Overridden Methods for `RaisedCosineBasisLinear` and `RaisedCosineBasisLog`
+
+1. `_evaluate`: For these objects, this method should evaluate the raised cosine function (with either linear or logarithmic spacing) at the specified samples.
+2. `_get_samples`: This method should return equidistant samples over the domain of the raised cosine function.
+3. `_check_n_basis_min`: This method checks whether the number of raised cosine basis functions meets the minimum requirement.
+
+## The `OrthExponentialBasis` Class
+
+The `OrthExponentialBasis` class extends the `Basis` abstract class and represents an orthogonal exponential basis set. It overrides the `_evaluate`, `_get_samples`, and `_check_n_basis_min` methods with specific implementations for orthogonal exponential basis functions.
```
billbrod, line 112: needs a longer description or a link to what a "orthogonal exponential basis set" means.

docs/developers_notes/basis_module.md
```diff
@@ -0,0 +1,130 @@
...
+## Developer Guidelines
+
+Developers are welcome focus on either develop non-abstract classes for new basis function types
+as we as add additional checks at evaluation, as well as improve the documentation and readability
+of the code. What we won't advise is to alter the abstract-classes structure, which may affect
+the overall module and how each of the objects interacts with each other.
+
+Feel free however to work on any piece of code you fell you can improve. More importantly, remember
+to thoroughly test all your classes and functions, and provide clear, detailed comments in your code. 
+This will not only help others use your library, but also facilitate future maintenance and development.
```
billbrod, line 130: These lines are generic and belong in the intro note

docs/examples/plot_1D_basis_function.py
```diff
@@ -0,0 +1,73 @@
...
+# Plotting the Basis Elements
+# -----------------------
+
+# Plot the evaluated basis elements
+plt.figure()
+plt.title(f"M-spline basis with {eval_basis.shape[0]} elements")
+plt.plot(eval_basis.T)
+plt.show()
+
+# %%
```
billbrod, line 53: I think it's worth adding a section introducing `evaluate_on_grid`  here: the point of that method is to simplify the generation of sample points, right?

docs/examples/plot_2D_basis_function.py
```diff
@@ -0,0 +1,154 @@
...
+## Defining a 2D Basis Function
+
+To model the response of a neuron to the position of an animal in an arena, we need a 2D model for stimuli.
+In this case, the stimuli are the 2D coordinates (x, y) representing the animal's position at each time step.
+
+We can model a response function for multi-dimensional stimuli as the external product of two 1D basis functions.
+The response function approximation for the 2D case would be:
+
+$$
+f(x, y) \\approx \sum_{ij} \\alpha_{ij} \, a_i (x) b_j(y).
```
billbrod, line 15: why is this approximate?

docs/examples/plot_2D_basis_function.py
```diff
@@ -0,0 +1,154 @@
...
+y_coord = np.linspace(0, 50, 1000)
+
+# Evaluate the product basis at the x and y coordinates
+eval_basis = prod_basis.evaluate(x_coord, y_coord)
+
+# Output the number of elements and samples of the evaluated basis, 
+# as well as the number of elements in the original 1D basis objects
+print(f"Product of two 1D splines with {eval_basis.shape[0]} "
+      f"basis element and {eval_basis.shape[1]} samples:\n"
+      f"\t- a_basis had {a_basis._n_basis_funcs} elements\n\t- b_basis had {b_basis._n_basis_funcs} elements.")
```
billbrod, line 51: I think it's worth plotting the separate 1D bases first

docs/examples/plot_2D_basis_function.py
```diff
@@ -0,0 +1,154 @@
...
+# Basis Evaluation and Plotting
+# ------------------------------------
+# We evaluate the product of two 1D basis objects which requires two inputs: 
+# one for the x-coordinate and one for the y-coordinate.
+# The number of elements of the product basis is the product of the elements of the 1D basis.
+
+# Define a trajectory with 1000 time-points representing the recorded trajectory of the animal
+T = 1000
+x_coord = np.linspace(0, 10, 1000)
+y_coord = np.linspace(0, 50, 1000)
```
billbrod, line 42: Why have them go to different maximum values? Unless there's a specific reason that we explain, make them identical.

docs/examples/plot_2D_basis_function.py
```diff
@@ -0,0 +1,154 @@
...
+# To plot a 2D basis set, we evaluate the basis on a grid of points over the basis function domain.
+# We use the `evaluate_on_grid` method of neurostatslib.basis for this.
+
+X, Y, Z = prod_basis.evaluate_on_grid(200, 200)
+
+# Setup a 3D plot
+fig, ax = plt.subplots(subplot_kw={"projection": "3d"})
+
+# Plot only the support
+Z[Z == 0] = np.nan
```
billbrod, line 65: This is opaque to me -- why do this?

docs/examples/plot_2D_basis_function.py
```diff
@@ -0,0 +1,154 @@
...
+ax.plot_surface(X, Y, Z[50], cmap="viridis", alpha=0.8)
+ax.plot_surface(X, Y, Z[83], cmap="rainbow", alpha=0.8)
+ax.plot_surface(X, Y, Z[125], cmap="inferno", alpha=0.8)
```
billbrod, line 68: Why these three? are we just picking three at random? Then we should say that. also why different cmaps?

docs/examples/plot_2D_basis_function.py
```diff
@@ -0,0 +1,154 @@
...
+import numpy as np
+import matplotlib.pyplot as plt
+import neurostatslib as nsl
+
+# Define 1D basis objects
+a_basis = nsl.basis.MSplineBasis(n_basis_funcs=20, order=3)
+b_basis = nsl.basis.MSplineBasis(n_basis_funcs=10, order=2)
+
+# 2D basis function as the product of the two 1D basis objects
+prod_basis = a_basis * b_basis
```
billbrod, line 30: Since the later header says "Additive basis", the one for this section should say "Multiplicative basis" (or similar)

Also, should explain why someone would do this. E.g., "A multiplicative basis is an efficient way of generating a simple representation of x,y position in physical or visual space. etc"

docs/examples/plot_2D_basis_function.py
```diff
@@ -0,0 +1,154 @@
...
+      f"basis element and {eval_basis.shape[1]} samples:\n"
+      f"\t- a_basis had {a_basis._n_basis_funcs} elements\n\t- b_basis had {b_basis._n_basis_funcs} elements.")
+
+# %%
+# 2D Basis Elements Plotting
+# -----------------------
+# To plot a 2D basis set, we evaluate the basis on a grid of points over the basis function domain.
+# We use the `evaluate_on_grid` method of neurostatslib.basis for this.
+
+X, Y, Z = prod_basis.evaluate_on_grid(200, 200)
```
billbrod, line 59: worth explaining what X,Y,Z are

docs/examples/plot_2D_basis_function.py
```diff
@@ -0,0 +1,154 @@
...
+
+# Setup a 3D plot
+fig, ax = plt.subplots(subplot_kw={"projection": "3d"})
+
+# Plot only the support
+Z[Z == 0] = np.nan
+ax.plot_surface(X, Y, Z[50], cmap="viridis", alpha=0.8)
+ax.plot_surface(X, Y, Z[83], cmap="rainbow", alpha=0.8)
+ax.plot_surface(X, Y, Z[125], cmap="inferno", alpha=0.8)
+plt.title(f"Product basis with {eval_basis.shape[0]} elements")
```
billbrod, line 69: what does the shape refer to?

docs/examples/plot_2D_basis_function.py
```diff
@@ -0,0 +1,154 @@
...
+# %%
+# Define and Evaluate a 2D Additive Basis
+# ---------------------------------------
+# Just like in the multiplicative case, we can create an additive basis function as the sum of two 1D basis functions.
+# We approximate the response function for multi-dimensional stimuli as the sum of two 1D basis functions.
+# For the 2D case, we can approximate our response function as follows:
+# $$
+# f(x,y) \\approx \alpha_i \cdot a_i(x) + \beta_j \cdot b_j(y)
+# $$
+# Here, we simply add two basis objects, 'a_basis' and 'b_basis', together to get the additive basis.
```
billbrod, line 81: why would someone use an additive basis? something like "Where a mulitiplicative basis is good for representing space, an additive basis is useful for ..."

docs/examples/plot_2D_basis_function.py
```diff
@@ -0,0 +1,154 @@
...
+
+X, Y, Z = additive_basis.evaluate_on_grid(200, 200)
+
+fig, ax = plt.subplots(subplot_kw={"projection": "3d"})
+
+# Only plot the support
+Z[Z == 0] = np.nan
+ax.plot_surface(X, Y, Z[5], cmap="viridis", alpha=0.8)
+ax.plot_surface(X, Y, Z[25], cmap="inferno", alpha=0.8)
+plt.title(f"Additive basis with {eval_basis.shape[0]} elements")
```
billbrod, line 107: same comments as for the multiplicative basis: explain X,Y,Z; why are you inserting NaNs; why plot those two values; what does shape mean?

docs/examples/plot_2D_basis_function.py
```diff
@@ -0,0 +1,154 @@
...
+c_basis = nsl.basis.RaisedCosineBasisLinear(n_basis_funcs=8)
+
+# Multiply or add basis of different types
+prod_basis = b_basis * c_basis
+add_basis = b_basis + c_basis
+
+print(f"Product of two 1D splines of different type with {eval_basis.shape[0]} "
+      f"basis element and {eval_basis.shape[1]} samples:\n"
+      f"\t- b_basis is of type {b_basis.__class__}\n\t- c_basis is of type {c_basis.__class__}")
+
```
billbrod, line 127: It is not clear to me how the code here relates to the text. Is this an angular basis? I think what you're showing is that the two combined bases don't need to be the same subclass, but that's not clear. I think that can just be said (doesn't need its own block and example code).

I think the way to do that is with [admonitions](https://squidfunk.github.io/mkdocs-material/reference/admonitions/#supported-types)! The `info` type probably. That's supported by [mkdocs gallery](https://smarie.github.io/mkdocs-gallery/generated/gallery/plot_0_sin/)

docs/examples/plot_2D_basis_function.py
```diff
@@ -0,0 +1,154 @@
...
+# %%
+# N-Dimensional Basis
+# -------------------
+# If you need to model an N-dimensional response function, you can combine
+# N 1D basis objects. However, be aware that the dimension of the evaluated basis function
+# will increase exponentially with the number of dimensions.
+# For example, evaluating a product of $N$ 1D bases with $T$ samples and $K$ basis element,
+# will output a $K^N \times T$ matrix.
```
billbrod, line 135: I think this should:
- describe why you might want higher dimensions (e.g., crossing head-direction with position in space)
- "be aware that the dimension of the evaluated basis function will increase exponentially with the number of dimensions" is a very abstract sentence. I think it should be followed with "; this means your basis will use exponentially more memory, so be careful! We'll raise an error if we think you'll go over the `gb_limit` set at basis initialization (XXGB by default"

docs/examples/plot_2D_basis_function.py
```diff
@@ -0,0 +1,154 @@
...
+b_basis = nsl.basis.RaisedCosineBasisLinear(n_basis_funcs=n_basis)
+c_basis = nsl.basis.RaisedCosineBasisLinear(n_basis_funcs=n_basis)
+
+prod_basis_3 = a_basis * b_basis * c_basis
+samples = np.linspace(0, 1, T)
+eval_basis = prod_basis_3.evaluate(samples, samples, samples)
+
+print(f"Product of three 1D splines results in {prod_basis_3._n_basis_funcs} "
+      f"basis elements.\nEvaluation output of shape {eval_basis.shape}")
+
```
billbrod, line 150: plot some views of these? has to be lower d slices, but still probably useful

docs/examples/plot_2D_basis_function.py
billbrod, line 1: Given that you include ND bases here,  probalby worth renaming to "plot higher dimensional basis function" or similar

billbrod (changes requested):
I'm doing two commits, one today on the docs, and then one when I'm back on basis.py and tests. 

I'm excited to for this refactor, I think it will make the basis objects much more usable.

For this one, higher-level thoughts: 

- There are some problems with how the docs render. I put suggestions for how to fix it on the relevant files but the a bigger picture question looking into this brought up for me is: how do we want to render the code references section? If you accept my suggestions, clicking on the "code references" header will just bring up the literate-nav showing all the different modules, which is fine but not that informative (and redundant with the navigation in the side panel). matplotlib, for example, has a bit more detail, which is manually written: [docs](https://matplotlib.org/stable/api/index.html), [source](https://github.com/matplotlib/matplotlib/blob/main/doc/api/index.rst). Don't need to get that all straightened out now, I created an issue to talk about it (#25)
- We're currently lacking examples showing what the different bases are and when you'd use them. That doesn't need to be in this PR (it's already big enough), but then we need an issue so we remember to add it eventually.
- `basis_module.md`: There's a lot of repetition in the description of the over-written methods (e.g., "`_get_samples`: This method should return equidistant samples over the domain of the raised cosine function."). Instead, they should specify what makes them different from the others or we should say in one place "The standard `_get_samples` is linearly-spaced points between 0 and `n_samples`" and then refer to that throughout. Right now, it's unclear that basically all of them have the same `_get_samples` *except* for `RaisedCosineBasisLog`
- For the 2D basis examples: I have some notes asking about why someone would use additive vs. multiplicative basis. That doesn't need to go there, but it does need to go somewhere. If we're not adding it in this PR, we need an issue for it.
- How does the GB limit get determined for Additive and Multiplicative bases? It looks like it's just the limit from the first basis, which feels weird to me. But I guess the idea is that all bases initialized by a given user will have the same limit, corresponding to how much of their RAM they want to use? Worth making that explicit somewhere

For mkdocstrings and numpydocs:
- it looks like only the "Methods" section is unsupported: [this page](https://mkdocstrings.github.io/griffe/docstrings/#sections) on the parsers shows full support of all other sections. I would remove that section from the docstring -- in the API reference created for documentation, the methods will all be rendered separately anyway, and for users I think it can also be a bit much.
- I prefer numpydocs over google, but it's up to you which you want to do. 
- I agree the rendering right now isn't great, but I think that's a mkdocs / mkdocstrings problem (not a docstring style problem), and [can](https://mkdocstrings.github.io/python/usage/customization/?h=css) be [configured](https://mkdocstrings.github.io/python/usage/)

docs/developers_notes/basis_module.md
```diff
@@ -0,0 +1,130 @@
...
+├─ Abstract Subclass RaisedCosineBasis
+│   │
+│   ├─ Subclass RaisedCosineBasisLinear
+│   │
+│   └─ Subclass RaisedCosineBasisLog
+│
+└─ Class OrthExponentialBasis
+```
+
+We've used abstract classes to ensure that any basis object inheriting from the superclass will implement the abstract methods of the superclass. This guarantees that if the inputs and outputs of those methods conform to the requirements specified by the abstract superclass, any new basis class implementation will integrate seamlessly with the basis objects already implemented in the library.
```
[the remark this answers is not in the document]
billbrod: Or I guess they'll interact with them via `+` and `*`, but still

docs/developers_notes/basis_module.md
```diff
@@ -0,0 +1,130 @@
...
+Additionally, a user can combine multiple objects that are subclasses of Basis, creating a higher-dimensional basis set object. Basis objects can be combined using the `__add__` and `__mul__` methods, which allow for the following syntax:
+
+```
+basis_add = basis_a + basis_b # calls basis_a.__add__(basis_b)
+
+basis_add = basis_a * basis_b # calls basis_b.__mul_(basis_b)
+```
+
+## The Class `neurostatslib.basis.Basis`
+
```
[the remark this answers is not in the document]
billbrod: and that we want developers to call `super().__init__()` in all subclasses (specifying the arguments they should pass as well)

docs/examples/plot_2D_basis_function.py
billbrod: in case people aren't familiar with meshgrid. maybe link to something about how this is how you represent something evaluated on a 2d surface in python --  I find this slightly confusing everytime I come across it.

docs/examples/plot_2D_basis_function.py
billbrod: also, why approximate?

src/neurostatslib/basis.py
```diff
@@ -11,315 +11,937 @@
...
+
+        Returns
+        -------
+        basis_funcs : (number of basis, number of samples)
+            Raised cosine basis functions
+
+        """
+
+        if any(sample_pts[0] < -1e-12) or any(sample_pts[0] > 1 + 1e-12):
+            raise ValueError(f"Sample points for RaisedCosine basis must lie in [0,1]!")
```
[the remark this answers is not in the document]
BalzaniEdoardo: thanks!

docs/developers_notes/intro.md
```diff
@@ -0,0 +1,33 @@
...
+
+Please note that these notes assume a certain level of programming knowledge. Familiarity with Python, object-oriented programming, and the NumPy, Scipy and Jax libraries would be beneficial when reading these notes.
+
+## Navigating the Developer Notes
+
+The Developer Notes are organized into sections, each dedicated to a different module or class within the `neurostatslib` library. Each section provides an overview of the class or module, a description of its role and functionality within the library, and a detailed guide to its methods and attributes.
+
+## Interact with us
+
+If you're considering contributing to the library, first of all, welcome aboard! As a first step, we recommend that you read the [`CONTRIBUTING.md`](https://github.com/flatironinstitute/generalized-linear-models/blob/main/CONTRIBUTING.md) guidelines. These will help you understand how to interact with other contributors and how to submit your changes.
```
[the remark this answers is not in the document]
BalzaniEdoardo: thanks, the issue was needed here!

docs/developers_notes/basis_module.md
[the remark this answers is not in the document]
BalzaniEdoardo: the + * interaction is what I meant, but I see your point

docs/developers_notes/basis_module.md
```diff
@@ -0,0 +1,130 @@
...
+### The Public Method `evaluate`
+
+The `evaluate` method checks input consistency and evaluates the basis function at some sample points. It accepts one or more numpy arrays as input, which represent the sample points at which the basis will be evaluated, and performs the following steps:
+
+1. Checks that the inputs all have the same sample size `N`, and raises a `ValueError` if this is not the case.
+2. Checks that the number of inputs matches what the basis being evaluated expects (e.g., one input for a 1-D basis, N inputs for an N-D basis, or the sum of N 1-D bases), and raises a `ValueError` if this is not the case.
+3. Checks that the output of the method does not exceed `gb_limit` GB of memory size. This size limit is specified at class initialization.
```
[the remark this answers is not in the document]
BalzaniEdoardo: it is approximate the size of an array of size N as  (N x size of the array elements).
for example, for the size of an array of 1000 numpy.int64, you would get 8000 bytes instead if you generate the array and you call sys.getsizeof(array) you would get 8112 bytes on my machine
(112 bytes is the size of the instantiated empty array that you would get with sys.getsizeof(np.array([],dtype=np.int64))).

The key is the approximation is accurate enough given that the overlay of the numpy & jax array is very small

docs/developers_notes/basis_module.md
```diff
@@ -0,0 +1,130 @@
+# The Basis Module
+
+The `neurostatslib.basis` module provides objects that allow users to construct and evaluate basis functions of various types. The classes are hierarchically organized as follows:
+
+```
+Abstract Class Basis
+|
+├─ Abstract Subclass AdditiveBasis
+│
+├─ Abstract Subclass MultiplicativeBasis
```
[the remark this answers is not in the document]
BalzaniEdoardo: Makes sense

docs/developers_notes/basis_module.md
BalzaniEdoardo: can you help me with this? 

I understand what is happening mechanistically with this basis, it is a basis of decaying exponential with different decaying rates that gets passed through a numerical orthonormalization (something like using Gram–Schmidt method). I am less sure of when could be useful and specific applications of it

docs/examples/plot_2D_basis_function.py
BalzaniEdoardo: I am thinking about the problem as that of identifying a non-linear response function that lives in some Hilbert space of some sort. In that case, by having a finite basis set we are likely approximating the function.

for example: a basis for L2 functions is necessarily countable infinite, otherwise it will be incomplete.

docs/developers_notes/basis_module.md
billbrod: We should talk with @ahwillia about this, I believe he came up with this basis set.

docs/developers_notes/basis_module.md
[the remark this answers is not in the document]
billbrod: Gotcha, that makes sense (and I think is worth describing somewhere, in docs or docstring).

But in general, like I think I've mentioned, I'm uncomfortable with this 3rd step. I don't think I'm aware of another scientific python package that implements this check? I.e., you can easily create a torch tensor, numpy array, or xarray that's too big for memory, and I'm assuming there's a good reason for that. My guess is that this step could cause it to break in weird ways, and it's not really doing what the user wants --- they'd want to make sure *their whole analysis pipeline* (or at least the GLM portion that's written in our package) doesn't exceed memory (rather than just the basis matrix), but there's no way we can guarantee that. 

Ultimately, it's up to you, but I think we should take the fact that this doesn't exist in other scientific packages, despite being easy to approximate and relevant, as a strong warning sign (happy to be proven wrong if this is somewhere!)

billbrod:
Broader point about the memory warning in the basis function that I pulled out of a smaller comment: in general, like I think I've mentioned, I'm uncomfortable with the basis checking the approximate memory usage against some limit and raising an exception if it goes above it. I'm not aware of another scientific python package that implements this check? I.e., you can easily create a torch tensor, numpy array, or xarray that's too big for memory, and I'm assuming there's a good reason for that. My guess is that this step could cause it to break in weird ways, and it's not really doing what the user wants --- they'd want to make sure *their whole analysis pipeline* (or at least the GLM portion that's written in our package) doesn't exceed memory (rather than just the basis matrix), but there's no way we can guarantee that. 

It's ultimately up to you, but I think we should take the fact that this doesn't exist in other scientific packages, despite being easy to approximate and relevant, as a strong warning sign (happy to be proven wrong if this does exist somewhere!)

docs/examples/plot_2D_basis_function.py
BalzaniEdoardo: I added a comment about it. It's a visualization issue, if i don't set nan outside the support it will plot the surface over all the grid and the overlap will make the basis function less visible. I choose non overlapping basis for this reason. The other thing we can do is to plot in different subplots, but I like the fact of visualizing the basis in the same frame

docs/examples/plot_2D_basis_function.py
BalzaniEdoardo: for point one: added an example
for point two: i will remove gb limit, so for now i added a warning saying that it may cause memory error

docs/examples/plot_2D_basis_function.py
BalzaniEdoardo: I added a plot of a couple of slices

docs/examples/plot_2D_basis_function.py
BalzaniEdoardo: changed the name

docs/developers_notes/intro.md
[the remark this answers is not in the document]
BalzaniEdoardo: done

docs/developers_notes/intro.md
```diff
@@ -0,0 +1,33 @@
...
+
+If you have any questions or need further clarification on any of the topics covered in these notes, please don't hesitate to reach out to us. You can do so via the [discussion](https://github.com/flatironinstitute/generalized-linear-models/discussions/landing) forum on GitHub.
```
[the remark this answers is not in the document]
BalzaniEdoardo: initialized a minimal welcome to the discusison announcement. I model it on the plenoptic one

docs/developers_notes/intro.md
```diff
@@ -0,0 +1,33 @@
...
+- **Contributors**: If you wish to contribute to the `neurostatslib` project, the Developer Notes can provide a solid foundation of understanding, helping to ensure that your contributions align with the existing structure and design principles of the library.
+
+- **Advanced Users**: While the primary focus of these notes is on development, they might also be of interest to advanced users who want a deeper understanding of the library's functionality. 
+
+Please note that these notes assume a certain level of programming knowledge. Familiarity with Python, object-oriented programming, and the NumPy, Scipy and Jax libraries would be beneficial when reading these notes.
+
+## Navigating the Developer Notes
+
+The Developer Notes are organized into sections, each dedicated to a different module or class within the `neurostatslib` library. Each section provides an overview of the class or module, a description of its role and functionality within the library, and a detailed guide to its methods and attributes.
+
```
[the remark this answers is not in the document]
BalzaniEdoardo: added the notation

src/neurostatslib/basis.py
```diff
@@ -11,315 +11,902 @@
...
+            spacing.
+
+        Returns
+        -------
+        basis_funcs : (number of basis, number of samples)
+            Raised cosine basis functions
+
+        """
+
+        if any(sample_pts[0] < -1e-12) or any(sample_pts[0] > 1 + 1e-12):
```
billbrod, line 660: feels weird to hardcode the epsilon like this -- it should depend on the precision of the dtype right?

src/neurostatslib/basis.py
```diff
@@ -11,315 +11,902 @@
...
+
+    """
+
+    def __init__(self, n_basis_funcs: int) -> None:
+        super().__init__(n_basis_funcs)
+
+    def _transform_samples(self, sample_pts: NDArray) -> NDArray:
+        """
+        Linearly map the samples from [0,1] to the the [0, (number of basis - 1) * pi]
+        Parameters
```
billbrod, line 715: missing blank newline

src/neurostatslib/utils.py
```diff
@@ -109,3 +109,45 @@ def plot_spike_raster(
     )
     ax.set(yticks=[], xlim=[0, spike_data.shape[1]])
     return ax
+
+
+def row_wise_kron(A: jnp.array, C: jnp.array, jit=False, transpose=True) -> jnp.array:
+    """
+    Compute the row-wise Kronecker product between two matrices using JAX.
```
billbrod, line 116: can we link to what the Kronecker product means?

prototests/test_splineBasis.py
```diff
@@ -0,0 +1,73 @@
...
+def pytest_generate_tests(metafunc):
+    # called once per each test function
+    funcarglist = metafunc.cls.params[metafunc.function.__name__]
+    argnames = sorted(funcarglist[0])
+    metafunc.parametrize(
+        argnames, [[funcargs[name] for name in argnames] for funcargs in funcarglist]
+    )
+
+
+class TestBspline:
```
billbrod, line 14: there currently is no Bspline, right? so this doesn't belong here -- why doesn't it make the tests fail?

billbrod: I think this folder is left over from earlier, should be deleted (copy contents to an issue or keep a copy locally if it will be useful later)

tests/utils_testing.py
billbrod, line 1: currently, it's not obvious why these live in a separate file. is the idea that these will be reused for e.g., the GLM tests?

They could all live within `conftest.py`, so we only have one utility file.

billbrod (changes requested):
I've got some suggestions for how to improve things in `basis.py`, but in general I like how this looks. I've got more questions about tests, I think we should sit down and talk through them.

- Created an issue for the return type annotation (#27 ) -- we should be able to fix this problem by just adding `:` as necessary.
- There are several places (e.g., `MSpline._evaluate`, `MSpline._get_samples`), where you take the first argument of a tuple (e.g., `sample_pts[0]` or `n_samples[0]`) -- why isn't it a for loop through the tuple? I assume this is related to how Additive/MultiplicativeBasis work. But from reading it, it looks like you allow for multiple args to be passed and then only use the first one, which seems weird
- Most of the basis functions use the same `_check_n_basis_min` and `_get_samples`, so it seems like those should be moved to the superclass and the dev notes updated to say "you may overwrite these methods, if the default doesn't work (see RaisedCosineLog) for an example)"
- We probably want to limit what bases are easily available. One way to do that is to define `__all__` at the top of the `basis.py` to include concrete bases and then add (see [here](https://learn.scientific-python.org/development/patterns/exports/)):
```
def __dir__() -> list[str]:
    return __all__
```
- `OrthExponentialBasis` seems to not work, the following results in a ValueError: 
```
bs = nsl.basis.OrthExponentialBasis(5, np.fromiter([2,2,2,2,2], float))
bs.evaluate_on_grid(5)
```
- In several places (e.g., OrthExponential's `decay_rates`, the `evaluate` calls), we require an array, where what I think we really want is an array-like. That is, we want to allow people to pass lists as well, which we can then transform to an array as necessary.
- In general, I find these tests really hard to understand. The tests should be unit tests, testing one thing at a time. But most of the tests runs once, and do a for loop over all basis classes. Instead, each test should operate on a single basis, and the pytest parameterize allows us to loop over them: see [docs](https://docs.pytest.org/en/4.6.x/example/parametrize.html#parametrizing-tests), [lesson](https://intersect-training.org/testing-lesson/05-refactor-for-testing/index.html#pytestmarkparametrize). Right now, if the tests would fail for two bases, we'd have no way of knowing (it would exit on the first one, never reaching the second).
- The tests also often test multiple functionalities and multiple arguments in. Both of those are better suited to being split up into individual calls.
- You disable pytest's capture at several points -- why? It makes the output really unwieldy.

tests/utils_testing.py
BalzaniEdoardo: I think of conftest as something containing:
1) fixtures
2) configurations
3) custom hooks

The `utils_testing.py` is not related, contains functions that can be called by the tests which inspect the modules; it feels a bit different then the other confs

prototests/test_splineBasis.py
BalzaniEdoardo: removed

src/neurostatslib/utils.py
BalzaniEdoardo: added a link to wikipedia, if you have better online research feel free to change

src/neurostatslib/basis.py
```diff
@@ -345,47 +932,58 @@ def mspline(x: NDArray, k: int, i: int, T: NDArray):
...
-    # spikes = np.random.RandomState(123).randn(nt) > 2.5
-    # fig, ax = plt.subplots(1, 1, sharex=True)
-    # X = convolve_1d_basis(
-    #     basis.transform(), spikes
-    # )
-    # ax.plot(spikes, color='k')
-    # ax.plot(
-    #     np.arange(basis.window_size - 1, nt),
-    #     X[0].T
+    import matplotlib.pyplot as plt  # type: ignore
```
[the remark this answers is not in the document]
BalzaniEdoardo: do we still need an if __name__=="__main__" with the instantiation of a few basis?

src/neurostatslib/basis.py
```diff
@@ -1,234 +1,435 @@
...
 
+from neurostatslib.utils import row_wise_kron
+
+__all__ = [
+    "MSplineBasis",
+    "RaisedCosineBasisLinear",
+    "RaisedCosineBasisLog",
+    "OrthExponentialBasis",
+    "AdditiveBasis",
+    "MultiplicativeBasis",
```
BalzaniEdoardo, line 21: I included `AdditiveBasis` and `MultiplicativeBasis` because otherwise it will interfere with the behavior of inspect (like inspect.get_members). Basically when you list the concrete classes you'll miss add and mult.

billbrod: not sure I understand your comment -- when I type `nsl.basis` and hit tab, I also see `AdditiveBasis` and `MultiplicativeBasis`. You're saying there's no way to avoid that and still have `inspect.get_members` work in the way you rely on in your tests?

src/neurostatslib/basis.py
```diff
@@ -11,315 +11,902 @@
...
+        -------
+            NDArray
+            The generated basis functions.
+
+        Raises
+        ------
+        NotImplementedError
+            If the subclass does not implement the _evaluate method.
+        """
+        # checks on input and outputs
```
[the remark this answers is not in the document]
billbrod: still think this

src/neurostatslib/basis.py
```diff
@@ -11,315 +11,902 @@
...
 
         """
-        if len(sample_pts) != self.window_size:
+        self._check_input_number(n_samples)
+        n_basis: int = np.prod(n_samples, dtype=int)
+        self._check_full_model_matrix_size(n_basis)
+
+        # get the samples
+        sample_tuple = self._get_samples(*n_samples)
+        Xs = np.meshgrid(*sample_tuple, indexing="ij")
```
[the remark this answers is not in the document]
billbrod: okay, this all makes sense to me, agree it's much cleaner. still think it's worth mentioning in the `Notes` section of the docstring or in `Returns`. want to prevent someone who's very used to `meshgrid` getting tripped up from the reversal.

src/neurostatslib/basis.py
```diff
@@ -135,51 +113,31 @@ def evaluate_on_grid(self, *n_samples: int) -> Tuple[NDArray, ...]:
...
-        if eval_basis.shape[0] > eval_basis.shape[1]:
-            warnings.warn(
-                "Basis set number exceeds the number of samples! "
-                "Consider reducing the number of basis or increase sample size."
-            )
-
     def _check_input_number(self, xi: Tuple) -> None:
         """
-        Check that the number of inputs provided by the user matches the number of inputs that the Basis object requires.
+        Check that the number of inputs provided by the user matches the number of inputs required.
```
billbrod, line 140: I think we talked about possibly changing this function name? to make it clearer that it's about the dimensionality of the basis and input. what did we decide?

src/neurostatslib/basis.py
```diff
@@ -652,16 +553,21 @@ def _evaluate(self, *sample_pts: NDArray) -> NDArray:
...
 
+        Raises
+        ------
+        ValueError
+            If the sample provided do not lie in [0,1].
         """
-
-        if any(sample_pts[0] < -1e-12) or any(sample_pts[0] > 1 + 1e-12):
+        if any(sample_pts < -np.finfo(sample_pts.dtype).resolution) or any(
+            sample_pts > 1 + np.finfo(sample_pts.dtype).resolution
```
billbrod, line 565: returning to this: where had you run into an issue with this? feels weird to check like this instead of just against 0/1, so I want to better understand where this came up

src/neurostatslib/basis.py
```diff
@@ -845,71 +738,114 @@ def _check_n_basis_min(self) -> None:
...
+            raise ValueError(
+                "Two or more rate are repeated! Repeating rate will result in a "
+                "linearly dependent set of function for the basis."
             )
 
-    def _evaluate(self, *sample_pts: NDArray) -> NDArray:
+    def _check_sample_range(self, sample_pts: NDArray):
         """
-        Generate basis functions with given spacing.
+        Check if the sample points are all positive.
```
billbrod, line 763: why is this the only basis that this runs on? don't they all require sample pts to be in some domain?

src/neurostatslib/basis.py
```diff
@@ -933,10 +869,6 @@ def mspline(x: NDArray, k: int, i: int, T: NDArray):
 
 
 if __name__ == "__main__":
```
billbrod, line 871: if name/main block should be deleted

tests/conftest.py
```diff
@@ -1,86 +1,23 @@
 import numpy as np
 import pytest
+
 import neurostatslib.basis as basis
 
+
 def pytest_generate_tests(metafunc):
```
billbrod, line 7: can we get a docstring describing what this does?

tests/test_basis.py
```diff
@@ -1,203 +1,1024 @@
...
+            )
+
+    @pytest.mark.parametrize("n_basis_funcs", [-1, 0, 1, 3, 10, 20])
+    def test_minimum_number_of_basis_required_is_matched(self, n_basis_funcs):
+        """
+        Verifies that the minimum number of basis functions required (i.e., 2) is enforced.
+        """
+        raise_exception = n_basis_funcs < 2
+        if raise_exception:
+            with pytest.raises(ValueError):
```
billbrod, line 102: I recommend using `pytest.raises(ValueError, match='contents of error message')` when checking that a specific exceptions is raised (see [here](https://docs.pytest.org/en/6.2.x/assert.html#assertions-about-expected-exceptions)), to make sure you got the valueerror you expected

src/neurostatslib/basis.py
```diff
@@ -11,315 +11,902 @@
...
-                "window_size -- are you sure that's what you want?"
+                "Basis set number exceeds the number of samples! "
+                "Consider reducing the number of basis or increase sample size."
             )
 
-    def check_in_support(self, x: NDArray):
-        """Check whether x lies within support.
+    def _check_input_number(self, xi: Tuple) -> None:
+        """
+        Check that the number of inputs provided by the user matches the number of inputs that the Basis object requires.
```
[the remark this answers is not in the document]
billbrod: this is still present?

src/neurostatslib/basis.py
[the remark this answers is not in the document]
billbrod: no, don't think so -- demonstrating the use is what the docs and tutorials are for, we're not expecting anyone to run these as scripts.

billbrod (changes requested):
This is looking good. Most of my comments are now small things, like typos. I also have some comments from before that haven't been addressed -- I tried to resolve all those that were addressed, so you should hopefully be able to find them easily. Couple of bigger picture points here:

- With `evaluate` and `evaluate_on_grid` defined in the base class, the docstring is consistent for all concrete basis classes. This makes it generic and a little confusing -- in particular, `evaluate_on_grid` says it accepts multiple ints for `n_samples`, but if you try:
   ```
   bs = nsl.basis.MSplineBasis()
   bs.evaluate_on_grid(5,5)
   ```
   which the docstring suggests should be okay, you get an exception. Not sure if there's a good way around this, but wanted to flag it.
- you document private attributes for several bases. I feel like we should document only public ones, but that some of these should be made public: `order`, `n_basis_funcs` (but `_n_input_samples` can stay private)
- Some of the `check_n_basis_min` docstrings say "Most of the basis work with at least 1 element, but some such as the RaisedCosineBasisLog requires a minimum of 2 basis to be well defined.", which should be removed (I assume it's copy-paste from the abstract superclass)
- for your tests, not really sure what using the `self.cls` syntax gets you -- why not just explicitly use the name of the basis class?

tests/conftest.py
BalzaniEdoardo: right now it is not used right now because I wrote the test using pytest.mark.parametrize instead. I kept it mostly to remember that it exists as an option because you can use to handle complex parametrizations and you can do a lot of cool things with it:
https://docs.pytest.org/en/latest/how-to/parametrize.html#basic-pytest-generate-tests-example

BalzaniEdoardo:
>     * for your tests, not really sure what using the `self.cls` syntax gets you -- why not just explicitly use the name of the basis class?

self.cls is used in the meta-test to make sure that every concrete class has its own Test class implemented.

docs/examples/plot_2D_basis_function.py
BalzaniEdoardo: done, I did the same for the 2D case. I think now it should be clearer

docs/examples/plot_2D_basis_function.py
BalzaniEdoardo: edited so tat the range is the same

BalzaniEdoardo:
>  I'm still seeing check_input_number in basis.py, is that what you intended?

Check input number check that an N-inputs are passed to N-dim basis, so that's a check we must provide

docs/examples/plot_ND_basis_function.py
```diff
@@ -17,30 +17,37 @@
...
-- $[b_1 (\mathbf{y}), ..., b_h (\mathbf{y}) ]$ for $\mathbf{y}$.
+- $ [ a_0 (\mathbf{x}), ..., a_{k-1} (\mathbf{x}) ] $ for $\mathbf{x}$
+- $[b_0 (\mathbf{y}), ..., b_{h-1} (\mathbf{y}) ]$ for $\mathbf{y}$.
 
 These basis functions can be combined in the following ways:
 
-1. **Addition:** If we assume that the response function can be adequately described as the sum of the two components, the function is defined as:
+1. **Addition:** If we assume that there is no interaction between the stimuli, the response function can be adequately described by the sum of the individual components. The function is defined as:
    $$
    f(\mathbf{x}, \mathbf{y}) \\approx \sum_{i} \\alpha_{i} \, a_i (\mathbf{x})  + \sum_{j} \\beta_j b_j(\mathbf{y}).
```
billbrod, line 27:
```suggestion
   f(\mathbf{x}, \mathbf{y}) \\approx \sum_{i=0}^{k-1} \\alpha_{i} \, a_i (\mathbf{x})  + \sum_{j=0}^{h-1} \\beta_j b_j(\mathbf{y}).
```

src/neurostatslib/basis.py
```diff
@@ -37,15 +37,15 @@ class Basis(abc.ABC):
 
     Attributes
     ----------
-    _n_basis_funcs : int
+    n_basis_funcs : int
         Number of basis functions.
     _n_input_samples : int
```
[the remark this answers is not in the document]
billbrod: it shows up in a couple other docstrings, but I'll just note it here

tests/conftest.py
billbrod: oh if it's not used, we should just remove it. I agree that's neat but it will still be in the git history if you need to refer back to it, or you can paste it into an issue or something if you think it will be used in the future.

the docstring is good though!

src/neurostatslib/basis.py
billbrod: still curious

billbrod (changes requested):
There are a couple of things from earlier comments that got lost in the shuffle:
- I think `evaluate` should mention in the docstring somewhere that it's running some checks and will raise exceptions if they fail (whether to use the `Raises` section or not is up to you, I'd double-check what other packages do)
- I think `evaluate_on_grid` docstring should mention the check it runs and that the meshgrid it generates uses the non-default indexing

Then there are a couple comments from the last review as well. Almost there!

docs/developers_notes/basis_module.md
```diff
@@ -45,24 +45,23 @@ This method performs the following steps:
...
+2. `_check_n_basis_min`: Checks the minimum number of basis functions required. This requirement can be specific to the type of basis.
 
 ## Contributors Guidelines
 
 ### Implementing Concrete Basis Objects
 To write a usable (i.e., concrete, non-abstract) basis object, you
 
 - **Must** inherit the abstract superclass `Basis`
-- **Must** define the `_evaluate`, `_check_n_basis_min`, and `_get_samples` methods with the expected input/output format, see [Code References](../../reference/neurostatslib/basis/) for the specifics.
+- **Must** define the `_evaluate` and `_check_n_basis_min` methods with the expected input/output format, see [Code References](../../reference/neurostatslib/basis/) for the specifics.
```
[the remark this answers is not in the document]
BalzaniEdoardo: I'll add a note. it would be easier for a basis to linearly map [0,1] to any finite interval so that one doesn't have to re-implement _get_samples, but requiring this seems too much

billbrod (approved):
Looks good to me! :tada:

PR #32, opened 2023-08-02
Convolution 1d

BalzaniEdoardo:
## Summary

- `utils.convolve_1d_trials` : parallelized convolution in valid mode. 
- `utils.nan_pad_conv`: add nans at the beginning of the trial (causal filters), end of the trials (anti-causal filters) or on both sides (acausal filter) to the output of `utils.convolve_1d_trials`  
- `utils.pad_last_dimension`: pads the last dimension of a jax.ndarray of any shape with a constant
- `docs/examples/plot_examples_covolution.py`: minimal example for the usage of the functions, with plots


src/neurostatslib/utils.py
```diff
@@ -4,44 +4,243 @@
...
+
+    if any(k == 0 for trial in time_series for k in trial.shape) | (
+        len(time_series) == 0
+    ):
+        raise ValueError("trials_time_series should not contain empty trials!")
+
+    # Check window size
+    ws = len(basis_matrix[0])
+    if any(trial.shape[1] < ws for trial in time_series):
+        raise ValueError(
```
gviejo, line 101: Should it raise a warning or an error. Could it be possible to just remove the trial in this case? 
In pynapple,  there is drop_short_intervals to solve this issue before calling the GLM. But I was wondering if it should just go on especially if it's only a small percentage of intervals that are too shorts.

gviejo (approved):

src/neurostatslib/utils.py
```diff
@@ -4,44 +4,263 @@
...
 import jax.numpy as jnp
 import matplotlib.pyplot as plt
 import numpy as np
-from numpy.typing import ArrayLike
+from numpy.typing import ArrayLike, NDArray
 
-# Broadcasted 1d convolution operations.
-# [[n x t],[w]] -> [n x (t - w + 1)]
+# Same trial duration
+# [[r x t x n], [w]] -> [r x (t - w + 1) x n]
```
billbrod, line 16: Nitpick: I find `x` confusing when writing the dimensionality like this. It's fine if it's clear to you, but I tend to go with `,` to match numpy convention, so `[[r, t, n], [w]]`

src/neurostatslib/utils.py
```diff
@@ -4,44 +4,263 @@
 from __future__ import annotations
 
 from functools import partial
-from typing import Optional, Union
+from typing import Iterable, List, Optional, Union
```
billbrod, line 7: Just a note: we could consider only supporting 3.10 and up, and get the easier type annotations. Since we're still developing right now...

src/neurostatslib/utils.py
```diff
@@ -4,44 +4,263 @@
...
+    Parameters
     ----------
-    basis_matrix : (n_basis_funcs, window_size)
-        Matrix holding 1d basis functions.
-    time_series : (n_neurons, n_timebins)
-        Matrix holding multivariate time series.
+    iterable :
+    Array-like object containing numpy or jax.numpy NDArrays.
+    expected_dim :
+    Number of expected dimension for the NDArrays.
```
billbrod, line 42:
```suggestion
    iterable :
        Array-like object containing numpy or jax.numpy NDArrays.
    expected_dim :
        Number of expected dimension for the NDArrays.
```
docstring not being parsed correctly

src/neurostatslib/utils.py
```diff
@@ -4,44 +4,263 @@
...
+
+    basis_matrix = jnp.asarray(basis_matrix)
+    # check input size
+    if basis_matrix.ndim != 2:
+        raise ValueError("basis_matrix must be a 2 dimensional array-like object.")
+    elif 0 in basis_matrix.shape:
+        raise ValueError(
+            "Empty basis_matrix provided. "
+            f"The shape of basis_matrix is {basis_matrix.shape}!"
+        )
```
billbrod, line 94: Does this come up often?

src/neurostatslib/utils.py
```diff
@@ -4,44 +4,263 @@
...
+
+    # Check window size
+    ws = len(basis_matrix)
+    if any(trial.shape[0] < ws for trial in time_series):
+        raise ValueError(
+            "Insufficient trial duration. The number of time points in each trial must "
+            "be greater or equal to the window size."
+        )
+
+    # Check if all trials have the same duration # check if needed
```
billbrod, line 122: are these two hashes from two comments being merged? then should delete second one

src/neurostatslib/utils.py
```diff
@@ -4,44 +4,263 @@
...
+    same_dur = time_series.ndim == 3 if isinstance(time_series, jnp.ndarray) else False
+
+    if same_dur:
+        # Trials have the same duration. # call conv instead of corr
```
billbrod, line 126: are these two hashes from two comments being merged? then should delete second one

src/neurostatslib/utils.py
```diff
@@ -4,44 +4,263 @@
...
+    if any(trial.shape[0] < ws for trial in time_series):
+        raise ValueError(
+            "Insufficient trial duration. The number of time points in each trial must "
+            "be greater or equal to the window size."
+        )
+
+    # Check if all trials have the same duration # check if needed
+    same_dur = time_series.ndim == 3 if isinstance(time_series, jnp.ndarray) else False
+
+    if same_dur:
```
billbrod, line 125:
```suggestion
    if isinstance(time_series, jnp.ndarray) and time_series.ndim == 3 :
```
this feels cleaner to me, but your original way is fine too, just a little confusing to understand

src/neurostatslib/utils.py
```diff
@@ -4,44 +4,263 @@
...
+    Parameters
+    ----------
+    array:
+        The array to be padded.
+    axis:
+        The axis to be padded.
+    window_size:
+        The window size to determine the padding.
+    filter_type:
+        The type of convolution, default is 'causal'. It must be one of 'causal', 'acausal', or 'anti-causal'.
```
billbrod, line 157: the way to do this with type annotation is `Literal['causal', 'acausal', 'anti-causal']` (need to import `Literal` from `typing`)

src/neurostatslib/utils.py
```diff
@@ -4,44 +4,263 @@
...
+    constant_values:
+        The constant values for padding, default is jnp.nan.
```
billbrod, line 159: don't need to mention default, since it's in signature (otherwise, easy to change it and forget to update it here)

src/neurostatslib/utils.py
```diff
@@ -4,44 +4,263 @@
...
+
+    padding_settings = {
+        "causal": (window_size, 0),
+        "acausal": ((window_size - 1) // 2, window_size - 1 - (window_size - 1) // 2),
+        "anti-causal": (0, window_size),
+    }
+
+    pad_width = (
+        ((0, 0),) * axis
+        + (padding_settings[filter_type],)
```
billbrod, line 182: should raise a more informative error message if `filter_type` isn't one of the expected values.

src/neurostatslib/utils.py
```diff
@@ -4,44 +4,263 @@
...
+    Add NaN padding to convolution trials based on the convolution type.
+
+    Parameters
+    ----------
+    conv_trials:
+        A 4D array-like of trials to be padded. Each trial has shape (n_neurons, n_basis_funcs, n_timebins_trial).
+    window_size:
+        The window size to determine the padding.
+    filter_type: str, optional
+        The type of convolution, by default 'causal'. It must be one of 'causal', 'acausal', or 'anti-causal'.
```
billbrod, line 203: same note about `Literal`

tests/test_convolution_1d.py
```diff
@@ -0,0 +1,194 @@
...
+
+
+class Test1DConvolution:
+    @pytest.mark.parametrize("basis_matrix", [np.zeros((n, n)) for n in [0, 1, 2]])
+    @pytest.mark.parametrize("trial_count_shape", [(1, 30, 2), (2, 10, 20)])
+    def test_basis_matrix_type(self, basis_matrix, trial_count_shape: tuple[int]):
+        vec = np.ones(trial_count_shape)
+        raise_exception = any(k == 0 for k in basis_matrix.shape)
+        if raise_exception:
+            with pytest.raises(ValueError, match=r"Empty basis_matrix provided\. "
```
billbrod, line 14: doesn't `r` make it a raw string? why use that here?

tests/test_convolution_1d.py
```diff
@@ -0,0 +1,194 @@
...
+    )
+    def test_check_dim(self, iterable, filter_type):
+        raise_exception = any(trial.ndim != 3 for trial in iterable)
+        if raise_exception:
+            with pytest.raises(ValueError, match="conv_trials must be an iterable of 3D arrays"):
+                utils.nan_pad_conv(iterable, 3, filter_type)
+        else:
+            utils.nan_pad_conv(iterable, 3, filter_type)
+
+    @pytest.mark.parametrize("filter_type", ["causal", "acausal", "anti-causal", ""])
```
billbrod, line 116: probably don't need to test the allowed `filter_types` here, because otherwise this is identical to `test_check_dim`

tests/test_convolution_1d.py
```diff
@@ -0,0 +1,194 @@
...
+        if raise_exception:
+            with pytest.raises(ValueError, match="window_size must be a positive integer!"):
+                utils.nan_pad_conv(iterable, window_size, "anti-causal")
+        else:
+            padded = utils.nan_pad_conv(iterable, window_size, "causal")
+            for trial in padded:
+                print(trial.shape, window_size)
+            assert all(np.isnan(trial[:window_size]).all() for trial in padded), "Missing NaNs at the " \
+                                                                                       "beginning of the array!"
+            assert all(not np.isnan(trial[window_size:]).any() for trial in padded), "Fund NaNs at the " \
```
billbrod, line 143:
```suggestion
            assert all(not np.isnan(trial[window_size:]).any() for trial in padded), "Found NaNs at the " \
```

tests/test_convolution_1d.py
```diff
@@ -0,0 +1,194 @@
...
+                                                                                      "end of the array!"
+            assert all(padded[k].shape[0] == iterable[k].shape[0] - 1 + window_size for k in range(len(padded))), \
+                "Size after padding doesn't match expectation. Should be T + window_size - 1."
+
+    @pytest.mark.parametrize("iterable",
+                             [[np.zeros([2, 5, 4]), np.zeros([2, 6, 4])]]
+                             )
+    @pytest.mark.parametrize("window_size", [0, 1, 2, 3, 5, 6])
+    def test_padding_nan_anti_causal(self, window_size, iterable):
+        raise_exception = (not isinstance(window_size, int)) or (window_size <= 0)
```
billbrod, line 153: not actually testing what happens if `window_size` is not an `int`, probably should

docs/gen_ref_pages.py
```diff
@@ -23,9 +23,24 @@
...
+        this_module_path = Path("src") / path.parent.name
+        module_index = ""
+        for module_scripts in sorted(this_module_path.rglob("*.py")):
+            if "__init__" in module_scripts.name:
+                continue
+            module_index += f"* [{module_scripts.name.replace('.py', '')}]" \
+                            f"({module_scripts.name.replace('.py', '.md')})\n"
+
+        with mkdocs_gen_files.open(full_doc_path, "w") as fd:
+            fd.write(module_index)
```
billbrod, line 40: what's with this new version? what does it do that the other doesn't? would like a comment explaining that.

billbrod: (so that I know whether to use this myself at some point!)

billbrod (changes requested):
Just some small changes requested here, this looks good otherwise

- What's the difference between `nan_pad_conv` and `pad_dimension`? Do we expect users to always call `nan_pad_conv` and never call `pad_dimension`? Then it should be hidden. Otherwise, docstring should be more clear as to when to use one vs. the other. (can use `Related` section in docstring)

src/neurostatslib/utils.py
BalzaniEdoardo: It should not, but I see somebody filtering data with the wrong parameters and ending up with an empty basis. very rare probably

src/neurostatslib/utils.py
BalzaniEdoardo: fixed it

src/neurostatslib/utils.py
BalzaniEdoardo: removed comments

src/neurostatslib/utils.py
BalzaniEdoardo: I checked and tested the logic, if the conversion to array worked, I have already checked that time_series has 3 dimensions. I simplified the if close to just isinstance

src/neurostatslib/utils.py
BalzaniEdoardo: done

tests/test_convolution_1d.py
BalzaniEdoardo: this is a leftover after the line was split in two. it is needed for the next line but it can be removed

tests/test_convolution_1d.py
BalzaniEdoardo: I removed looping on filter_types on the test_check_dim. The difference between the tests was that this one raises an exception

tests/test_convolution_1d.py
BalzaniEdoardo: good catch

docs/gen_ref_pages.py
BalzaniEdoardo: I'll add a comment but it basically solves the problem of having an empty "Neurostatslib" page in the docs. Instead of that you'll have an index page with the list of modules

src/neurostatslib/basis.py
```diff
@@ -321,7 +322,7 @@ def _evaluate(self, *xi: NDArray) -> NDArray:
         Returns
         -------
         :
-            The basis function evaluated at the samples (number of samples x number of basis)
+            The basis function evaluated at the samples (number of samples x n_basis_funcs)
```
billbrod, line 325:
```suggestion
            The basis function evaluated at the samples, shape (n_samples, n_basis_funcs)
```
To  be consistent with other dosctrings.

src/neurostatslib/basis.py
```diff
@@ -374,7 +375,7 @@ def _evaluate(self, *xi: NDArray) -> NDArray:
...
```
billbrod, line 378:
```suggestion
            The basis function evaluated at the samples, shape (n_samples, n_basis_funcs)
```
To  be consistent with other dosctrings.

docs/gen_ref_pages.py
billbrod: oh cool, I was wondering the way to do this

src/neurostatslib/utils.py
```diff
@@ -4,44 +4,271 @@
...
+        # trials have different length
+        conv_trials = [
+            _CORR_VARIABLE_TRIAL_DUR(jnp.atleast_2d(trial), basis_matrix)
+            for trial in time_series
+        ]
+
+    return conv_trials
+
+
+def pad_dimension(
```
billbrod, line 135:
```suggestion
def _pad_dimension(
```
Make it private, because we don't expect users to use it.

src/neurostatslib/utils.py
```diff
@@ -4,44 +4,271 @@
...
+    array: jnp.ndarray,
+    axis: int,
+    window_size: int,
+    filter_type: Literal["causal", "acausal", "anti-causal"] = "causal",
+    constant_values: float = jnp.nan,
+) -> jnp.ndarray:
+    """
+    Add padding to the last dimension of an array based on the convolution type.
+
```
billbrod, line 144:
```suggestion
    """
    Add padding to the last dimension of an array based on the convolution type.

    This is a helper function used by `nan_pad_conv`, which is the function we expect the user will interact with.
```

billbrod (approved):
Looks good to me. Just a couple suggestions that should be included and then good to go.

PR #36, opened 2023-08-03
Transpose basis

BalzaniEdoardo:
This PR simply transposes the results of basis evaluate and evaluate_on_grid so that the time axis is the first. I should have edited the documentation accordingly. please give it a check to see if everything looks fine!

### DIFF
--- a/docs/examples/plot_ND_basis_function.py
+++ b/docs/examples/plot_ND_basis_function.py
@@ -137,7 +137,7 @@ axs[1].set_title(f"$A_{{{basis_elem_idx[1]}}}(x,y) = "
                  f"b_{{{basis_b_element}}}(x)$", color="b")
 
 for cc in range(len(basis_elem_idx)):
-    axs[cc].contourf(X, Y, Z[:, :, basis_elem_idx[cc]], cmap="Blues")
+    axs[cc].contourf(X, Y, Z[..., basis_elem_idx[cc]], cmap="Blues")
     axs[cc].set_xlabel("x-coord")
     axs[cc].set_ylabel("y-coord")
 plt.tight_layout()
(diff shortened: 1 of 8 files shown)


docs/examples/plot_ND_basis_function.py
```diff
@@ -137,7 +137,7 @@
                  f"b_{{{basis_b_element}}}(x)$", color="b")
 
 for cc in range(len(basis_elem_idx)):
-    axs[cc].contourf(X, Y, Z[basis_elem_idx[cc]], cmap="Blues")
+    axs[cc].contourf(X, Y, Z[:, :, basis_elem_idx[cc]], cmap="Blues")
```
billbrod, line 140:
```suggestion
    axs[cc].contourf(X, Y, Z[..., basis_elem_idx[cc]], cmap="Blues")
```
these are equivalent, but the `...` syntax is more robust to adding dimensions. just pointing it out, in case you aren't familiar with it

billbrod (approved):
This looks good to me! 

Nitpicks / things to consider:

- Also saw that, in several places, you use `number of basis` to describe the shape of an array (e.g., `evaluate_on_grid`). Wouldn't it be better to use `n_basis_funcs`? That's the name of both the argument at initialization and the attribute.
- Similarly, you use `number of samples`. I don't think there's an analogous arg/attribute, but it feels weird to use a phrase like that. Not sure what to replace it with though.
- I just noticed that, while you changed the name of the attributes and methods to say "input dimensionality", the error message still refers to input number. may want to change that.

docs/examples/plot_ND_basis_function.py
billbrod: There are other locations where this syntax could be helpful too, but I just noted it here.

Issue #30, opened 2023-07-28
Docstring/signature and inheritance


billbrod:
With `evaluate` and `evaluate_on_grid` defined in the base class, the docstring is consistent for all concrete basis classes. This makes it generic (no details about the specific basis and e.g., its allowed domain) and confusing -- in particular, `evaluate_on_grid` says it accepts multiple ints for n_samples, but if you try:

```python
bs = nsl.basis.MSplineBasis()
bs.evaluate_on_grid(5,5)
```

which the docstring suggests should be okay, you get an exception.

Could restructure the code to require subclasses to implement `evaluate`, which would then allow for more specific docstrings and signatures. Could have them call `super.evaluate()` to get the checks, or move that functionality to different methods and note in the docs that they should be used? Not sure there's a way to *force* them to be used if we restructure, but could just add that to the tests run on all basis classes (in particular, check that an exception is raised, rather than that they use that specific method)

PR #39, opened 2023-08-21
Bspline

BalzaniEdoardo:
Added BSplineBasis and CyclicBSplineBasis to the basis module


src/neurostatslib/basis.py
```diff
@@ -465,6 +462,13 @@ def _generate_knots(
         )
         return self.knot_locs
 
+    @staticmethod
+    def _check_samples_non_empty(sample_pts):
+        if sample_pts.shape[0] == 0:
+            raise ValueError(
+                "Empty sample array provided. At least one sample is required for evaluation!"
```
billbrod, line 469: why is this here and only for BSpline?

from the tests, looks like you're expecting it when `evaluate_on_grid(x)` for `x<=0`, but is that really an issue? for the negative case, we'll get an error in `np.linspace` and for `x=0`, you'd get an empty array otherwise, which seems reasonable. Is there a specific reason to check for this?

and in general, don't like that we'll have different behavior for BSpline than for the rest, when really, it's the same concerns, right?

src/neurostatslib/basis.py
```diff
@@ -533,6 +537,218 @@ def _check_n_basis_min(self) -> None:
...
+        The higher this number, the smoother the basis representation will be.
+
+    Attributes
+    ----------
+    order :
+        Spline order.
+
+    Methods
+    -------
+    _evaluate(x_tuple)
```
billbrod, line 560: we had decided not to document hidden methods / attributes, right (because we don't want users calling them)? could just replace this with `evaluate` and the information is still right

src/neurostatslib/basis.py
```diff
@@ -533,6 +537,218 @@ def _check_n_basis_min(self) -> None:
...
+
+        Raises
+        ------
+        AssertionError
+            If the sample points are not within the B-spline knots range unless `outer_ok=True`.
+
+        Notes
+        -----
+        This method evaluates the B-spline basis functions at the given sample points.
+        It requires the knots to be defined through the `_generate_knots` method.
```
billbrod, line 596: again, feels weird to discuss the hidden method here. Do we expect users to inherit `BSplineBasis` themselves? that's the only reason they'd need this info, yeah?

billbrod: wait, but `_generate_knots` is implemented in `SpineBasis` and that's the version that gets used by all the subclasses. So why do users need to know this -- I read this as "make sure to define `_generate_knots`, but that's not the case.

src/neurostatslib/basis.py
```diff
@@ -533,6 +537,218 @@ def _check_n_basis_min(self) -> None:
...
+        sample_pts[ind] = sample_pts[ind] + knots.max() - knots_orig[0]
+        # restore the original knots
+        self.knot_locs = knots_orig
+
+        return basis_eval
+
+    def _check_n_basis_min(self) -> None:
+        """Check that the user required enough basis elements.
+
+        Check that Cuclic-BSplineBasis has at least as many basis as the order of the spline +2
```
billbrod, line 738:
```suggestion
        Check that Cyclic-BSplineBasis has at least as many basis as the order of the spline +2
```

src/neurostatslib/basis.py
```diff
@@ -533,6 +537,218 @@ def _check_n_basis_min(self) -> None:
...
+        and at least 2*order - 2 basis.
```
billbrod, line 739: Might be worth explaining why this is the case (or including a reference that explains it).

src/neurostatslib/basis.py
```diff
@@ -533,6 +537,218 @@ def _check_n_basis_min(self) -> None:
...
+    Attributes
+    ----------
+    n_basis_funcs : int
+        Number of basis functions.
+    order : int
+        Order of the splines used in basis functions.
+
+    Methods
+    -------
+    _evaluate(sample_pts, der)
```
billbrod, line 651: again, don't think we should include hidden methods here.

src/neurostatslib/basis.py
```diff
@@ -533,6 +537,218 @@ def _check_n_basis_min(self) -> None:
...
+class CyclicBSplineBasis(SplineBasis):
+    """
+    B-spline 1-dimensional basis functions for cyclic splines.
+
+    Parameters
+    ----------
+    n_basis_funcs :
+        Number of basis functions.
+    order :
+        Order of the splines used in basis functions. Must lie within [1, n_basis_funcs].
```
billbrod, line 638: this isn't true for the cyclic one, is it?

src/neurostatslib/basis.py
```diff
@@ -533,6 +537,218 @@ def _check_n_basis_min(self) -> None:
...
+
+        Raises
+        ------
+        AssertionError
+            If the sample points are not within the B-spline knots range unless `outer_ok=True`.
+
+        Notes
+        -----
+        This method evaluates the B-spline basis functions at the given sample points.
+        It requires the knots to be defined through the `_generate_knots` method.
```
billbrod, line 685: same point about `_generate_knots`

src/neurostatslib/basis.py
```diff
@@ -533,6 +537,218 @@ def _check_n_basis_min(self) -> None:
...
+
+        Parameters
+        ----------
+        sample_pts :
+            The sample points at which the B-spline is evaluated. Must be a tuple of length 1.
+
+        Returns
+        -------
+        NDArray
+            The basis function evaluated at the samples (Time points x number of basis)
```
billbrod, line 675:
```suggestion
            The basis function evaluated at the samples, shape (n_samples, n_basis_funcs)
```
to be consistent with other basis objects

src/neurostatslib/basis.py
```diff
@@ -533,6 +537,218 @@ def _check_n_basis_min(self) -> None:
...
+
+        Parameters
+        ----------
+        sample_pts :
+            The sample points at which the B-spline is evaluated.
+
+        Returns
+        -------
+        NDArray
+            The basis function evaluated at the samples (Time points x number of basis)
```
billbrod, line 586:
```suggestion
            The basis function evaluated at the samples, shape (n_samples, n_basis_funcs)
```
to be consistent with other basis objects

src/neurostatslib/basis.py
```diff
@@ -533,6 +537,218 @@ def _check_n_basis_min(self) -> None:
...
+                self.knot_locs[0]
+                - self.knot_locs[-1]
+                + self.knot_locs[nk - self.order : nk - 1],
+                self.knot_locs,
+            )
+        )
+        ind = sample_pts > xc
+
+        # temporarily set the extended knots as attribute
+        self.knot_locs = knots
```
billbrod, line 717: just curious, why do this with an attribute (instead of just using the `knots` variable below)

src/neurostatslib/basis.py
```diff
@@ -870,3 +1086,83 @@ def mspline(x: NDArray, k: int, i: int, T: NDArray):
...
+    ----------
+    sample_pts :
+        An array containing sample points for which B-spline basis needs to be evaluated.
+    knots :
+        An array containing knots for the B-spline basis. The knots are sorted in ascending order.
+    order :
+        The order of the B-spline basis. Default is 4.
+    der :
+        The derivative of the B-spline basis to be evaluated. Default is 0.
+    outer_ok :
```
billbrod, line 1114: don't need to document default here, since it's in the function signature (and hard to keep in sync if we change defaults)

billbrod (changes requested):
Several comments, largely on inheritance structure and documentation. Some bigger points:

- in general, as mentioned in #30, having `evaluate` come from `Basis` and the subclasses implement `_evaluate` gets confusing, because the docstrings for the function we want users to call is generic. might be worth revisiting whether the super class has `evaluate` or `_evaluate` (which the subclasses call). The other way to do this would be to remove the `eval_basis = self._evaluate(*xi)` from `Basis.evaluate()` and have the subclasses call `super().evaluate()` to get the checks, but then I don't know whether we can then require users to implement `evaluate` (maybe we can, just by marking it as `@abstract`, but I'm not sure). Out of scope for this PR, but it's making me even more inclined to thinking we should change this. If you'd like, I can try it out while you're working on the GLM class (it shouldn't change the interface at all).
- For all spline bases, does `knot_locs` need to be an attribute (or at least, a public one)? seems like `_generate_knots` is always called when it's needed. And we don't expect the user to use it, right? so probably shouldn't be an attribute
- I don't love the tests (e.g., `test_minimum_number_of_basis_required_is_matched` for either BSpline) that have a long boolean expression as `raise_exception` and then a bunch of ORs in the `match` for `pytest.raises` -- it feels like we're losing the specificity of the test (and the readability), since I don't know which error is expected in which condition. the alternative would be to split them up into separate tests or to have an `if` block defining the match string. both are long, but I think the second is worth doing, since it seems like a decent tradeoff between readability and compactness. Thoughts?

src/neurostatslib/basis.py
BalzaniEdoardo: If you don't check, with x=0 you get an "invalid input data" error message from the scipy splev call. 

I am fine with that but:
- it is a bit nested and less clear for a user what may have caused it.
- MSpline, RaisedCosineLog, RaisedCosineLinear returns an empty array, not an error.
- OrthExp raises another class specific error, which is that you need at least as many samples as basis function to define the basis (otherwise the ortonormalization would be not-defined).

Uniformity in behavior is hard to get given that different basis construction have different requirements.
My rationale here was try to surface the error and give a  basis-type specific explanation of the error which is understandable. 
If we try to get something that looks more uniform there are a few way:
- return an empty array for B-spline too instead of calling splev in case of empty input
- raise the error instead of returning an empty array for all of the basis function we have

Which one do you prefer?

src/neurostatslib/basis.py
BalzaniEdoardo: I figured it is actually true. I was wrapping too many knots in my original algorithm. that did not change the final results because I was adding a bunch of zero where the wrapping was not necessary.

src/neurostatslib/basis.py
BalzaniEdoardo: I checked my algorithm and I was wrapping too many knots. one need to wrap only order-1 knots instead I was wrapping more. this has been fixed now and it turns out that all the splines has the same condition for the number of basis

src/neurostatslib/basis.py
BalzaniEdoardo: none really, it is a left-over of previous implementations in which the evaluation in BSplineBasis used the self.knots_locs to generate the basis, so I needed to change the attribute to get the desired results. I changed that logic so that I use the bspline function now, so i can use knots directly

gviejo (approved):

src/neurostatslib/basis.py
```diff
@@ -94,6 +94,11 @@ def evaluate(self, *xi: NDArray) -> NDArray:
         -------
         :
             The generated basis functions.
+        Raises
```
billbrod, line 97: I feel like there's supposed to be an empty line here

src/neurostatslib/basis.py
billbrod: Gotcha. I'm inclined to work towards uniformity in behavior, and I don't have a strong preference here, but I think raising the error for all of them is the right way? because if someone calls `evaluate_on_grid(0)`, they've *really* misunderstood what it does. right?

src/neurostatslib/basis.py
```diff
@@ -592,39 +594,20 @@ def _evaluate(self, sample_pts: NDArray) -> NDArray:
...
         Notes
         -----
-        This method evaluates the B-spline basis functions at the given sample points.
-        It requires the knots to be defined through the `_generate_knots` method.
-
         The evaluation is performed by looping over each element and using `splev`
         from SciPy to compute the basis values.
         """
-        super()._check_samples_non_empty(sample_pts)
+        # super()._check_samples_non_empty(sample_pts)
```
billbrod, line 600: if not using this, should remove this line

billbrod (approved):
Looks good to me, just remove the one commented-out line, and let's discuss the uniformity of behavior for `evaluate_on_grid(0)`.

Like Guillaume mentioned, I think these tests are really hard to understand, but that seems out of scope here.

PR #48, opened 2023-10-14
Mkdocs config

BalzaniEdoardo:
- Created a `gallery_config.py` file to facilitate sorting tutorials in the documentation index based on the alphabetical order of the tutorial script names.
- Renamed the tutorials so that each file begins with "plot_N_...", where 'N' represents a two-digit number indicating the display order in the documentation index.
- Configured the gallery in `mkdocs.yaml` to execute the configuration script during the build process.


docs/gallery_conf.py
```diff
@@ -0,0 +1,20 @@
...
+min_reported_time = 0
+if 'SOURCE_DATE_EPOCH' in os.environ:
+    min_reported_time = sys.maxint if sys.version_info[0] == 2 else sys.maxsize
+
+# To be used as the "base" config,
+# this script is referenced in the mkdocs.yaml under `conf_script` option: docs/gallery_conf.py
+conf = {
+    'reset_argv': DefaultResetArgv(),
+    'min_reported_time': min_reported_time,
+    'within_subsection_order': FileNameSortKey
```
billbrod, line 19: can we get some explanation for what these do?

docs/gallery_conf.py
```diff
@@ -0,0 +1,20 @@
+import os
+import sys
+
+import plotly.io as pio
+pio.renderers.default = 'sphinx_gallery'
```
billbrod, line 5: what does this do?

pyproject.toml
```diff
@@ -63,7 +63,8 @@ docs = [
     "mkdocs-gen-files",             # Plugin for generating additional files in mkdocs
     "mkdocs-literate-nav>=0.6.1",   # Plugin for literate-style navigation in mkdocs
     "mkdocs-gallery",               # Plugin for adding image galleries to mkdocs
-    "mkdocs-material",
+    "mkdocs-material",              # Material theme for mkdocs
+    "plotly",                       # Plotly library for interactive visualizations (used in config)
```
billbrod, line 67: why is plotly added? none of the examples look interactive.

Also, unsure why the Github interface is highlighting everything in red

billbrod (changes requested):
I think we need some more explanation in the configure script.

- Is it possible to sort the tutorials using something other than alphabet sort? This is what I do for plenoptic as well, but it seems a bit odd to expose it so much to the user. Something like a variable or custom function that determines the order (like `sorted` accepts) which would go in the configuration somewhere.
- Is it possible to modify the name of the scripts that get built for tutorials? We probably don't want them all to start with `plot_`. Is there a reason not to just use every `.py` file in `examples/`?
- Probably not necessary yet, but is it possible to break the tutorials into groups? So we have one set for "Basis objects", one for "Building design matrices", etc.

docs/gallery_conf.py
BalzaniEdoardo: the first reset to defaults, I don't know why it is needed, I copy-pasted the config from 

https://github.com/smarie/mkdocs-gallery/blob/main/docs/gallery_conf.py

clearly it is not commented enough, nor it is in the docs

https://smarie.github.io/mkdocs-gallery/#2-configure-mkdocs

I'll add the comments of each one, the important one is `FileNameSortKey`, I explain in the ccn-template how to change sorting method, we can customize it. 
We should discuss about how to sort things

docs/gallery_conf.py
BalzaniEdoardo: It seems like it sets the default renderer for Plotly... since i am not using it, I removed it. It is included because I used the config script that is in the mkdocs gallery as starting point for mine. this is actually useless for us

docs/gallery_conf.py
BalzaniEdoardo: in particular I describe the parameters here
https://ccn-template--39.org.readthedocs.build/en/39/notes/03-documentation/#set-up-mkdocs-gallery

BalzaniEdoardo: this is the call to sorted. the `__call__` method of self.conf["within_subsection_order"] returns the order you need.


```python
if sort_files:
    listdir = sorted(listdir, key=self.conf["within_subsection_order"]())
```

more details https://github.com/flatironinstitute/ccn-template/issues/37

billbrod: I think we want to link to the [sphinx-gallery config page](https://sphinx-gallery.github.io/stable/configuration.html), which has a good explanation -- you said the mkdocs-gallery dev is trying to duplicate what they're doing, right?

billbrod: From the sphinx-gallery docs, it looks like [DefaultResetArgv](https://sphinx-gallery.github.io/stable/gen_modules/sphinx_gallery.gen_gallery.html#sphinx_gallery.gen_gallery.DefaultResetArgv) does nothing -- it returns an empty list (whereas if it were not set, it [passes nothing](https://sphinx-gallery.github.io/stable/configuration.html#passing-command-line-arguments-to-example-scripts)). Does behavior change when this is set? For me building locally, doesn't look like it.

billbrod (changes requested):
Still want a bit more details in that config, questions about reset argv, and the bigger picture questions I posted in my last review (pasted below). Not all of those need to be addressed in this PR, but I'd like a bit of clarity on them, maybe create relevant issues.

billbrod (approved):
Answering my comments from earlier:
- `conf['within_subsection_order']` gets passed to sorted and called on the filenames (within a subsection, as that name implies; we currently only have every tutorial in one subsection), so anything that works with sorted will work there.
- we now execute every .py file within `docs/examples/`. The default behavior is to *include* (but not execute) every .py file, and only execute those that start with `plot_` (I'm not sure why you would want to include but not execute a file -- I guess if it took too long?). See [sphinx-gallery docs](https://sphinx-gallery.github.io/stable/configuration.html#parsing-and-executing-examples-via-matching-patterns) for more details on how to control that behavior (you can also exclude things entirely).
- #76 will handle the organization of tutorials.

PR #49, opened 2023-10-14
Add fourier

BalzaniEdoardo:
- Added `FourierBasis` to the basis module
- tests and docs modified accordingly


docs/examples/plot_1D_basis_function.py
```diff
@@ -81,3 +83,87 @@
...
+# %%
+# ### The Fourier Basis
+# Another type of basis available is the Fourier Basis. Fourier basis are ideal to capture periodic and
+# quasi-periodic patterns. Such oscillatory, rhythmic behavior is a common signature of many neural signals.
+# Additionally, the Fourier basis has the advantage of being orthogonal, which simplifies the estimation and
+# interpretation of the model parameters, each of which will represent the relative contribution of a specific
+# oscillation frequency to the overall signal.
+
+
+# A Fourier basis can be instantiated with the usual syntax.
```
billbrod, line 95: Currently this renders as a comment on top of the first code line -- did you mean for it to be rendered markdown?

docs/examples/plot_1D_basis_function.py
```diff
@@ -81,3 +83,87 @@
...
+# The user can pass the desired frequencies for the basis or
+# the frequencies will be set to np.arange(n_basis_funcs//2).
```
billbrod, line 97: I don't think description is correct -- there's only one option for initialization, setting `n_freqs`, whereas this makes it sounds like there's two. 

And I think the meaning of `n_freqs` needs to be explained. When `n_freqs=4`, we get cosines and sines from 0 to 3 periods. I think maybe the name of the initialization argument should be changed? To something like `max_freq` (which is the current `n_freq-1`), and then describe the basis set as consisting of cosines and negatives sines from 0 to 3 periods? I would use the term "quadrature pair" here, but my guess is that's not any more transparent to most people...

docs/examples/plot_1D_basis_function.py
```diff
@@ -81,3 +83,87 @@
...
+# %%
+# !!! note "Fourier basis convolution and Fourier transform"
+#     The Fourier transform of a signal $ s(t) $ restricted to a temporal window $ [t_0,\;t_1] $ is
+#     $$ \\hat{x}(\\omega) = \\int_{t_0}^{t_1} s(\\tau) e^{-j\\omega \\tau} d\\tau. $$
+#     where $ e^{-j\\omega \\tau} = \\cos(\\omega \\tau) - j \\sin (\\omega \\tau) $.
+#
+#     When computing the cross-correlation of a signal with the Fourier basis functions,
+#     we essentially measure how well the signal correlates with sinusoids of different frequencies,
+#     within a specified temporal window. This process mirrors the operation performed by the Fourier transform.
+#     Therefore, it becomes clear that computing the cross-correlation of a signal with the Fourier basis defined here
```
billbrod, line 123:
```suggestion
#     Therefore, computing the cross-correlation of a signal with the Fourier basis defined here
```

docs/examples/plot_1D_basis_function.py
```diff
@@ -81,3 +83,87 @@
...
+plt.subplot(121)
+plt.title("Cos")
+plt.plot(samples, eval_basis[:, :4])
+plt.subplot(122)
+plt.title("Sin")
+plt.plot(samples, eval_basis[:, 4:])
+plt.tight_layout()
+
+# %%
+# !!! note "Fourier basis convolution and Fourier transform"
```
billbrod, line 115: I think this probably shouldn't be an admonition: the next section of code is just demonstrating that this is true, so it should be regular text.

docs/examples/plot_1D_basis_function.py
```diff
@@ -81,3 +83,87 @@
...
+
+# define a signal
+signal = np.random.normal(size=n_samples)
+
+# evaluate the basis
+_, eval_basis = nmo.basis.FourierBasis(n_freqs=n_freqs).evaluate_on_grid(n_samples)
+
+# compute the cross-corr with the signal and the basis
+# Note that we are inverting the time axis of the basis because we are aiming
+# for a cross-correlation, while np.convolve compute a convolution which would flip the time axis.
```
billbrod, line 138: Can't we just compute the correlation directly to avoid this confusion? It's true, but provides an extra hurdle for folks (and then we could call out this equivalency in an admonition)

docs/examples/plot_1D_basis_function.py
```diff
@@ -81,3 +83,87 @@
...
+    ]
+)
+
+# compute the power (add back sin(0 * t) = 0)
+fft_complex = np.fft.fft(signal)
+fft_amplitude = np.abs(fft_complex[:n_freqs])
+fft_phase = np.angle(fft_complex[:n_freqs])
+# compute the phase and amplitude from the convolution
+xcorr_phase = np.arctan2(np.hstack([[0], xcorr[n_freqs:]]), xcorr[:n_freqs])
+xcorr_aplitude = np.sqrt(xcorr[:n_freqs] ** 2 + np.hstack([[0], xcorr[n_freqs:]]) ** 2)
```
billbrod, line 152: probably worth explaining why this is the same as computing the phase and amplitude

billbrod: This would probably involve a bit more describing why we look at the phase and amplitude... Something like

"The Fourier transform is often interpreted by looking at its amplitude and phase, which is the polar representation: amplitude is the distance from the origin, while phase is the angle. The amplitude conveys how much "stuff" is present at a given frequency, whereas the phase describes its alignment [or something??]. Thus, to compute the amplitude we ... and to compute the phase we ..."

I'm sure Eero has a good intuitive explanation of this somewhere, maybe in the math tools notes.

src/nemos/basis.py
```diff
@@ -103,7 +104,7 @@ def _check_evaluate_input(self, *xi: ArrayLike) -> Tuple[NDArray]:
             # make sure array is at least 1d (so that we succeed when only
             # passed a scalar)
             xi = tuple(np.atleast_1d(np.asarray(x, dtype=float)) for x in xi)
-        except TypeError:
+        except (TypeError, ValueError):
```
billbrod, line 107: what additionally is being caught here?

src/nemos/basis.py
```diff
@@ -1115,6 +1117,92 @@ def evaluate_on_grid(self, n_samples: int) -> Tuple[NDArray, NDArray]:
...
+    """Set of 1D Fourier basis.
+
+    Parameters
+    ----------
+    n_freqs
+            Number of frequencies. The number of basis function will be 2*n_freqs - 1.
+    """
+
+    def __init__(self, n_freqs: int):
+        super().__init__(n_basis_funcs=2 * n_freqs - 1)
```
billbrod, line 1130: see my comment earlier about changing this parametrization

src/nemos/basis.py
```diff
@@ -1115,6 +1117,92 @@ def evaluate_on_grid(self, n_samples: int) -> Tuple[NDArray, NDArray]:
...
+            Evaluated Fourier basis, shape (n_samples, n_basis_funcs).
+
+        Notes
+        -----
+            If the frequencies provided are np.arange(n_freq), convolving a signal
+            of length n_samples with this basis is equivalent, but slower,
+            then computing the FFT truncated to the first n_freq components.
+
+            Therefore, convolving a signal with this basis is equivalent
+            to compute the FFT over sliding window.
```
billbrod, line 1171: this is over-indented

src/nemos/basis.py
```diff
@@ -1115,6 +1117,92 @@ def evaluate_on_grid(self, n_samples: int) -> Tuple[NDArray, NDArray]:
...
+            >>> import nemos as nmo
+            >>> import numpy as np
+            >>> n_samples, n_freqs = 1000, 10
+            >>> basis = nmo.basis.FourierBasis(n_freqs*2)
+            >>> eval_basis = basis.evaluate(np.linspace(0, 1, n_samples))
+            >>> sinusoid = np.cos(3 * np.arange(0, 1000) * np.pi * 2 / 1000.)
+            >>> conv = [np.convolve(eval_basis[::-1, k], sinusoid, mode='valid')[0] for k in range(2*n_freqs-1)]
+            >>> fft = np.fft.fft(sinusoid)
+            >>> print('FFT power:   ', np.round(np.real(fft[:10]), 4))
+            >>> print('Convolution: ', np.round(conv[:10], 4))
```
billbrod, line 1184: This renders fine, but is also over-indented

src/nemos/basis.py
```diff
@@ -1115,6 +1117,92 @@ def evaluate_on_grid(self, n_samples: int) -> Tuple[NDArray, NDArray]:
...
+        Examples
+        --------
+            >>> import nemos as nmo
+            >>> import numpy as np
+            >>> n_samples, n_freqs = 1000, 10
+            >>> basis = nmo.basis.FourierBasis(n_freqs*2)
+            >>> eval_basis = basis.evaluate(np.linspace(0, 1, n_samples))
+            >>> sinusoid = np.cos(3 * np.arange(0, 1000) * np.pi * 2 / 1000.)
+            >>> conv = [np.convolve(eval_basis[::-1, k], sinusoid, mode='valid')[0] for k in range(2*n_freqs-1)]
+            >>> fft = np.fft.fft(sinusoid)
```
billbrod, line 1182: don't think these last three lines are needed

tests/test_basis.py
```diff
@@ -1395,7 +1569,7 @@ def test_evaluate_returns_expected_number_of_basis(
                 f"The first dimension of the evaluated basis is {eval_basis.shape[1]}",
             )
 
-    @pytest.mark.parametrize("sample_size", [6, 30, 35])
+    @pytest.mark.parametrize("sample_size", [12, 30, 35])
```
billbrod, line 1572: why this change?

billbrod (changes requested):
I think we should change the parametrization of this new basis, as discussed in my comments, and probably work a bit on how it's explained. I don't really have any comments on how it's implemented, looks great.

docs/examples/plot_1D_basis_function.py
BalzaniEdoardo: yes, I'll make sure it works as intended

tests/test_basis.py
BalzaniEdoardo: 6 samples is not enough for the frequencies that i am testing for. It raises the aliasing error in the Fourier basis init

docs/examples/plot_1D_basis_function.py
BalzaniEdoardo: I'll give it a try. Let me know if the description improves!

BalzaniEdoardo: the description was of a earlier version in which I allowed to pass a frequency vector but I changed it to this more restrictive parametrization

src/nemos/basis.py
```diff
@@ -1202,15 +1206,15 @@ def _check_n_basis_min(self) -> None:
         Raises
         ------
         ValueError
-            If an insufficient number of basis element is requested for the basis type
+            If an insufficient number of basis element is requested for the basis type.
         """
         if self.n_basis_funcs < 1:
             raise ValueError(
-                f"Object class {self.__class__.__name__} requires >= 1 basis elements. "
+                f"Object class {self.__class__.__name__} requires >= 0 basis elements. "
```
billbrod, line 1213: I see the error message now matches the check, but shouldn't this actually at least 1 basis element? With `max_freq=0` (just the DC term), you get a `n_basis_funcs=1`. And it doesn't make sense to have `max_freq<0`

docs/examples/plot_1D_basis_function.py
```diff
@@ -81,3 +83,87 @@
 plt.plot(samples, eval_basis)
 plt.show()
 
+# %%
+# ### The Fourier Basis
+# Another type of basis available is the Fourier Basis. Fourier basis are ideal to capture periodic and
+# quasi-periodic patterns. Such oscillatory, rhythmic behavior is a common signature of many neural signals.
+# Additionally, the Fourier basis has the advantage of being orthogonal, which simplifies the estimation and
```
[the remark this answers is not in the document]
billbrod: still think this. at least a foot note or link

docs/examples/plot_1D_basis_function.py
billbrod: yup, this looks good now!

docs/examples/plot_1D_basis_function.py
billbrod: still think this

docs/examples/plot_1D_basis_function.py
billbrod: still think this

docs/examples/plot_1D_basis_function.py
```diff
@@ -112,44 +113,45 @@
 plt.tight_layout()
 
 # %%
-# !!! note "Fourier basis convolution and Fourier transform"
-#     The Fourier transform of a signal $ s(t) $ restricted to a temporal window $ [t_0,\;t_1] $ is
-#     $$ \\hat{x}(\\omega) = \\int_{t_0}^{t_1} s(\\tau) e^{-j\\omega \\tau} d\\tau. $$
-#     where $ e^{-j\\omega \\tau} = \\cos(\\omega \\tau) - j \\sin (\\omega \\tau) $.
+# ## Fourier Basis Convolution and Fourier Transform
```
billbrod, line 116:
```suggestion
# #### Fourier Basis Convolution and Fourier Transform
```

should be header level underneath the Fourier basis one

docs/examples/plot_1D_basis_function.py
```diff
@@ -112,44 +113,45 @@
...
-#     within a specified temporal window. This process mirrors the operation performed by the Fourier transform.
-#     Therefore, it becomes clear that computing the cross-correlation of a signal with the Fourier basis defined here
-#     is equivalent to computing the discrete Fourier transform on a sliding window of the same size
-#     as that of the basis.
+# When computing the cross-correlation of a signal with the Fourier basis functions,
+# we essentially measure how well the signal correlates with sinusoids of different frequencies,
+# within a specified temporal window. This process mirrors the operation performed by the Fourier transform.
+# Therefore, it becomes clear that computing the cross-correlation of a signal with the Fourier basis defined here
+# is equivalent to computing the discrete Fourier transform on a sliding window of the same size
+# as that of the basis.
```
billbrod, line 126:
```suggestion
# we are measuring how well the signal correlates with sinusoids of different frequencies,
# within a specified temporal window. This process mirrors the operation performed by the Fourier transform.
# Therefore, computing the cross-correlation of a signal with the Fourier basis defined here
# is equivalent to computing the discrete Fourier transform on a sliding window of the same size
# as that of the basis.
```

billbrod (changes requested):
I like this parametrization much more. Still have some comments about explanation from the last round that need addressing, and then some small wording suggestions here.

src/nemos/basis.py
```diff
@@ -1177,6 +1179,96 @@ def evaluate_on_grid(self, n_samples: int) -> Tuple[NDArray, NDArray]:
...
+        >>> import nemos as nmo
+        >>> import numpy as np
+        >>> n_samples, max_freq = 1000, 10
+        >>> basis = nmo.basis.FourierBasis(max_freq)
+        >>> eval_basis = basis.evaluate(np.linspace(0, 1, n_samples))
+        >>> sinusoid = np.cos(3 * np.arange(0, 1000) * np.pi * 2 / 1000.)
+        >>> conv = [np.convolve(eval_basis[::-1, k], sinusoid, mode='valid')[0] for k in range(2*max_freq+1)]
+        >>> fft = np.fft.fft(sinusoid)
+        >>> print('FFT power:   ', np.round(np.real(fft[:max_freq]), 4))
+        >>> print('Convolution: ', np.round(conv[:max_freq], 4))
```
kushalkolar, line 1250: Is there a way in mkdocs to set this as a python codeblock? Can you remove the `>>>`? Because right now in the docs the `>>>` shows up and when you copy it copies with the `>>>` so a user can't just copy and paste and run from the docs :smile: 

for example at the bottom here the arrows render and get copied: https://nemos.readthedocs.io/en/latest/reference/nemos/utils/#nemos.utils.pytree_map_and_reduce

src/nemos/basis.py
```diff
@@ -1177,6 +1179,96 @@ def evaluate_on_grid(self, n_samples: int) -> Tuple[NDArray, NDArray]:
...
+        """
+        (sample_pts,) = self._check_evaluate_input(sample_pts)
+        # assumes equi-spaced samples.
+        if sample_pts.shape[0] / np.max(self._frequencies) < 2:
+            raise ValueError("Not enough samples, aliasing likely to occur!")
```
kushalkolar, line 1255: Maybe report `sample_pts.shape[0]` and `max(self._frequencies)` to the user?

kushalkolar (commented):
minor comments :)

billbrod:
Now being handled by #300 instead

PR #50, opened 2023-10-14
Reparametrize raisedcosine

BalzaniEdoardo:
This edits introduce a one extra hyper-parameters for initializing the linear raised cosine bumps, and an two one for the log-raised cosines:

- **width**: this hyper-parameter is set for both linear- and log-spaced raised cosines. It controls the width of the cosine bumps, default is 2. The original implmentation was assuming this to be 1, but that produces aliasing
- **time_scaling**: this hyper-parameter is available only for log-spaced raised cosines and controls the logarithmic stretch magnitude. as time_scaling approaches to 0, the transformation becomes linear. large time_scaling result in basis that are tightly concentrated around t=0
- **enforce_decay_to_zero**: boolean, if true, it enforce the last basis to decay to zero.


.github/workflows/ci.yml
```diff
@@ -16,7 +16,7 @@ jobs:
   tox:
     strategy:
       matrix:
-        os: [ubuntu-latest] #[ubuntu-latest, macos-latest, windows-latest]
+        os: [ubuntu-latest, macos-latest] #[ubuntu-latest, macos-latest, windows-latest]
```
billbrod, line 19: it's important to have cross-platform tests, but is there a particular reason this is added in this PR? we could open a separate PR adding all OSs and versions (we should probably do this before the workshop in February)

src/nemos/basis.py
```diff
@@ -505,7 +505,7 @@ def _check_n_basis_min(self) -> None:
 
 
 class MSplineBasis(SplineBasis):
-    """M-spline 1-dimensional basis functions.
+    """M-spline[$^1$](references) 1-dimensional basis functions.
```
billbrod, line 508:
```suggestion
    """M-spline[$^1$](#references) 1-dimensional basis functions.
```

src/nemos/basis.py
```diff
@@ -577,7 +576,7 @@ def evaluate_on_grid(self, n_samples: int) -> Tuple[NDArray, NDArray]:
 
 class BSplineBasis(SplineBasis):
     """
-    B-spline 1-dimensional basis functions.
+    B-spline[$^1$](references) 1-dimensional basis functions.
```
billbrod, line 579:
```suggestion
    B-spline[$^1$](#references) 1-dimensional basis functions.
```

src/nemos/basis.py
```diff
@@ -771,34 +769,73 @@ def evaluate_on_grid(self, n_samples: int) -> Tuple[NDArray, NDArray]:
         return super().evaluate_on_grid(n_samples)
 
 
-class RaisedCosineBasis(Basis, abc.ABC):
-    def __init__(self, n_basis_funcs: int) -> None:
+class RaisedCosineBasisLinear(Basis):
+    """Represent linearly-spaced raised cosine basis functions.
+
+    This implementation is based on the cosine bumps used by Pillow et al.[$^1$](references)
```
billbrod, line 775:
```suggestion
    This implementation is based on the cosine bumps used by Pillow et al.[$^1$](#references)
```

src/nemos/basis.py
```diff
@@ -771,34 +769,73 @@ def evaluate_on_grid(self, n_samples: int) -> Tuple[NDArray, NDArray]:
...
+    to uniformly tile the internal points of the domain.
```
billbrod, line 776: This phrase is confusing, should be explained somewhere.

src/nemos/basis.py
```diff
@@ -771,34 +769,73 @@ def evaluate_on_grid(self, n_samples: int) -> Tuple[NDArray, NDArray]:
...
-    def _transform_samples(self, sample_pts: NDArray) -> NDArray:
-        """
-        Abstract method for transforming sample points.
+    @property
+    def width(self):
+        """Return width of the raised cosine."""
+        return self._width
+
+    @width.setter
+    def width(self, width: float):
```
billbrod, line 804: we don't want users to set width after the fact right? do we want to make the public attribute set-only? (and then just use `_check_width` during initialization)

src/nemos/basis.py
```diff
@@ -815,14 +852,18 @@ def evaluate(self, sample_pts: NDArray) -> NDArray:
...
-        transform_sample_pts = self._transform_samples(sample_pts)
-
-        shifted_sample_pts = (
-            transform_sample_pts[:, None]
-            - (np.pi * np.arange(self.n_basis_funcs))[None, :]
+        peaks = np.linspace(0, 1, self.n_basis_funcs)
+        delta = peaks[1] - peaks[0]
+        basis_funcs = 0.5 * (
+            np.cos(
+                np.clip(
```
billbrod, line 859: I think I'd like some comments explaining this kind of confusing set of operations. What we're doing here is constructing a different set of shifted sampled points for each of the basis functions and constraining them to a single period of the cosine function. So saying something along that.

src/nemos/basis.py
```diff
@@ -845,127 +886,136 @@ def evaluate_on_grid(self, n_samples: int) -> Tuple[NDArray, NDArray]:
...
-            A transformed version of the sample points that matches the Raised Cosine basis domain,
-            shape (number of samples, ).
-        """
-        return sample_pts * np.pi * (self.n_basis_funcs - 1)
-
     def _check_n_basis_min(self) -> None:
         """Check that the user required enough basis elements.
 
-        Check that the number of basis is at least 1.
+        Check that the number of basis is at least 2.
```
billbrod, line 892: why did this change?

src/nemos/basis.py
```diff
@@ -845,127 +886,136 @@ def evaluate_on_grid(self, n_samples: int) -> Tuple[NDArray, NDArray]:
...
 
-class RaisedCosineBasisLog(RaisedCosineBasis):
-    """Log-spaced raised cosine basis functions used by Pillow et al. [2]_.
+class RaisedCosineBasisLog(RaisedCosineBasisLinear):
+    """Represent log-spaced raised cosine basis functions.
 
-    These are "cosine bumps" that uniformly tile the space.
+    Similar to `RaisedCosineBasisLinear` but the basis functions are log-spaced.
+    This implementation is based on the cosine bumps used by Pillow et al.[$^1$](#references)
+    to uniformly tile the internal points of the domain.
```
billbrod, line 911: same point about how this phrase is confusing

src/nemos/basis.py
billbrod: Also, does the Pillow paper ever actually use this basis function?

src/nemos/basis.py
```diff
@@ -845,127 +886,136 @@ def evaluate_on_grid(self, n_samples: int) -> Tuple[NDArray, NDArray]:
...
+            self.time_scaling = 50.0
+        else:
+            self.time_scaling = time_scaling
+
+    @property
+    def time_scaling(self):
+        return self._time_scaling
+
+    @time_scaling.setter
+    def time_scaling(self, time_scaling):
```
billbrod, line 951: Same point as `width` about whether we want it to be settable after initialization

src/nemos/basis.py
```diff
@@ -845,127 +886,136 @@ def evaluate_on_grid(self, n_samples: int) -> Tuple[NDArray, NDArray]:
...
+        sample_pts : NDArray
+            Sample points used for evaluating the splines,
+            shape (n_samples, ).
 
         Returns
         -------
-        :
-            A transformed version of the sample points that matches the Raised Cosine basis domain,
-            shape (n_sample_points, ).
+        NDArray
```
billbrod, line 970: don't think the type needs to be here?

src/nemos/basis.py
```diff
@@ -845,127 +886,136 @@ def evaluate_on_grid(self, n_samples: int) -> Tuple[NDArray, NDArray]:
...
-        return (
-            np.power(
-                10,
-                -(np.log10((self.n_basis_funcs - 1) * np.pi) + 1) * sample_pts
-                + np.log10((self.n_basis_funcs - 1) * np.pi),
-            )
-            - 0.1
+        # This log-stretching of the sample axis has the following effect:
+        # - as the time_scaling tends to 0, the points will be linearly spaced.
+        # - as the time_scaling tends to inf, basis will be dense around 0.
```
billbrod, line 976:
```suggestion
        # This log-stretching of the sample axis has the following effect:
        # - as the time_scaling tends to 0, the points will be linearly spaced across the whole domain.
        # - as the time_scaling tends to inf, basis will be small and dense around 0 and progressively larger and less dense towards 1.
```

src/nemos/basis.py
```diff
@@ -845,127 +886,136 @@ def evaluate_on_grid(self, n_samples: int) -> Tuple[NDArray, NDArray]:
...
+        (sample_pts,) = self._check_evaluate_input(sample_pts)
+        if not self.enforce_decay_to_zero:
+            # keep the last basis, i.e. do not enforce decay to zero
+            # for the filter.
+            eval_basis = super().evaluate(self._transform_samples(sample_pts))
+        else:
+            # temporarily add n_trim basis element
+            # guaranteeing that the last basis element decays to 0 for any width.
+            n_trim = int(np.ceil(self.width))
+            self.n_basis_funcs += n_trim
```
billbrod, line 1009: this kind of modification feels hacky. An alternative would be to have `evaluate` accept `n_basis_funcs` as an optional arg, defaulting to `self.n_basis_funcs`. But we don't want the public method to accept that argument, so not sure...

tests/test_basis.py
```diff
@@ -204,14 +209,94 @@ def test_evaluate_on_grid_input_number(self, n_input):
...
+                pytest.raises(
+                    ValueError, match="Only strictly positive time_scaling are allowed"
+                ),
+            ),
+            (
+                0,
+                pytest.raises(
+                    ValueError, match="Only strictly positive time_scaling are allowed"
+                ),
+            ),
```
billbrod, line 269:
```suggestion
            (-1, pytest.raises(ValueError, match="Only strictly positive time_scaling are allowed")),
            (0, pytest.raises(ValueError, match="Only strictly positive time_scaling are allowed")),
```

billbrod (changes requested):
This looks pretty much good to me! Some small changes scattered throughout, and then the following things I think we should discuss:
- Basis docstrings. I think they should have a line or two before parameters that describe what the basis looks like and when you would use them. E.g., do they tile uniformly, are they each about the same size, what types of variables make sense (visual input, head direction, etc.). And then the footnote to references belongs in those lines (it looks weird in the docstring in jupyter or whatever to have the link right at the top). This is probably out-of-scope for this PR, in which case we should open an issue for it.
- We need an explanation of the particularities of the different parameters for the Raised Cosine parameters. Much of that will end up in our separate repo, but we need some here. Do we want to hold off on that until we've worked through it in the separate repo and then add something here? (in which case, we should open an issue)

.github/workflows/ci.yml
BalzaniEdoardo: yes, makes sense. I'll remove it from this pr, but let's keep it in mind for the future

src/nemos/basis.py
BalzaniEdoardo: yep, it's eqn (5) in the paper

src/nemos/basis.py
BalzaniEdoardo: I'll add an example in the plot 1D basis function.  I think we should have a pr in which we work out a better description of the basis function and their properties. I started one in which i tried to make a table of the various properties.

BalzaniEdoardo: I think we should have  a minimal example here

src/nemos/basis.py
BalzaniEdoardo: sounds right

src/nemos/basis.py
BalzaniEdoardo: the log-stretching needed at least 2 samples, in this way the two raised cosines behave coherently.

src/nemos/basis.py
BalzaniEdoardo: I agree, the other way would be to be smarter in the way we compute the stretching. the equivalent operation would be to modify the scaling on the fly, but leaving the normalization constant (log(self.time_scaling + 1)) as it is.
basically the transform samples should look something like:
```python
    def _transform_samples(self, sample_pts: NDArray, scaling: float) -> NDArray:
        """
        Map the sample domain to log-space.

        Parameters
        ----------
        sample_pts :
            Sample points used for evaluating the splines,
            shape (n_samples, ).

        Returns
        -------
            Transformed version of the sample points that matches the Raised Cosine basis domain,
            shape (n_samples, ).
        """
        # This log-stretching of the sample axis has the following effect:
        # - as the time_scaling tends to 0, the points will be linearly spaced.
        # - as the time_scaling tends to inf, basis will be dense around 0.
        log_spaced_pts = np.log(scaling * sample_pts + 1) / np.log(
            self.time_scaling + 1
        )
        return log_spaced_pts
```

And the scaling will be equal to self.time_scaling, if no decay to zero is needed, some other constant computed appropriately otherwise.
what do you think of this solution?

billbrod: I think this solution is fine but it's not clear to me why this is equivalent. And I worry about being able to explain it clearly and succinctly.

The other alternative would be to have a private helper method `_shift_sample_pts`, which accepts `n_basis_funcs` and runs:
```python
        peaks = np.linspace(0, 1, self.n_basis_funcs)
        delta = peaks[1] - peaks[0]
        # generate a set of shifted sample points between 0 and 1
        return sample_pts[:, None] - peaks[None]) / (delta * self.width)
```

Or could potentially include the multiplication by pi and the clipping.

Or clip between -1 and 1 (and then `evaluate` just has the multiplication by pi, no clipping).

billbrod: And evaluate checks `self.enforce_decay_to_zero` to determine `n_basis_funcs`?

BalzaniEdoardo: in order to achieve the desired effect with the shifting, we should have that for the last basis that the peaks do not terminate in 1:

```
def scale_pts(time_scaling, width, n_basis_funcs, sample_pts):
    log_spaced_pts = np.log(time_scaling * sample_pts + 1) / np.log(
        time_scaling + 1
    )
    peaks = np.linspace(0, 1 - np.ceil(width) / (n_basis_funcs + np.ceil(width) - 1), n_basis_funcs)
    delta = 1 / (n_basis_funcs+ np.ceil(width) -1)
    basis_funcs = 0.5 * (
            np.cos(
                np.clip(
                    np.pi * (log_spaced_pts[:, None] - peaks[None]) / (delta * width),
                    -np.pi,
                    np.pi,
                )
            )
            + 1
    )
    return basis_funcs
hacky_version = nmo.basis.RaisedCosineBasisLog(10,width=2, time_scaling=20).evaluate_on_grid(1000)[1]
adjusting_peaks = scale_pts(time_scaling=20., width=2, n_basis_funcs=10, sample_pts=np.linspace(0,1,1000))
print(np.linalg.norm(hacky_version-adjusting_peaks))
```

BalzaniEdoardo: the function above compute the peaks appropriately, the delta is adjusted to maintaining the original spacing. The same result could be achieved by adjusting the log stretching, as I mentioned previously.

BalzaniEdoardo: My point is that, in order to have the math work out exactly, we need some computation that is not exactly transparent; and i think we need to preserve the transformation that we have now (the hacky solution of trimming basis). 

in the sense that the hacky solution is the right thing to do mathematically because we want the same temporal log-stretching for a given time_scaling parameter. if we simply adjust the peak locations and not the delta at the denominator of the raised cosines, this would not be true anymore.

BalzaniEdoardo:
I changed the code so that the peak location is computed so that the decay to zero of the last basis element happens as late as possible. this is not equivalent to what we had before, (it would be if i use `np.ceil(width)` instead of `width` to compute the last peak location.

src/nemos/basis.py
BalzaniEdoardo: I computed the peaks conditionally on the enforce_decay_to_zero flag, not it is easy to understand what's happening. the last peak location ensures that the last basis element decays to zero

src/nemos/basis.py
BalzaniEdoardo: I added the issue #72 that would be aimed to improve those docstrings

src/nemos/basis.py
billbrod: I think the current solution makes sense.

billbrod (approved):
I think this looks good to me!

PR #56, opened 2023-11-07
Reserve package name NEmoS

BalzaniEdoardo:
This is an alpha version of the code. We are releasing this version with the aim of reserving the package name on PyPI


BalzaniEdoardo:
changed only the deploy workflow

PR #80, opened 2024-01-12
updates authors/copyright info

billbrod:
change copyright holder from me to the authors, and adds citation file specifying authors. once we have something for people to cite (zenodo doi, paper, etc), we'll add that info to the citation file.

@BalzaniEdoardo : any other keywords we should add?


BalzaniEdoardo (changes requested):
I added a suggestion for the keyword, if you like it we can merge in

BalzaniEdoardo (approved):
looks good

PR #82, opened 2024-01-13
changed correlation to pytree

BalzaniEdoardo:
This small PR replaces the loop over trials in the convolution_1d with a tree map.
The function returns gets a tree with the time series (assuming the appropriate leaf structure) and returns a tree with the same structure with the convolved trials.

Tests have been improved and updated, checks on the tree structure have been also updated and scorporated from the main convolve function. They can potentially be used as validation and eventually being moved to a validation.py script


src/nemos/utils.py
```diff
@@ -123,21 +123,21 @@ def convolve_1d_trials(
...
+        conv_trials = _CORR_SAME_TRIAL_DUR(time_series, basis_matrix)
     else:
         # trials have different length
-        conv_trials = [
-            _CORR_VARIABLE_TRIAL_DUR(jnp.atleast_2d(trial), basis_matrix)
-            for trial in time_series
-        ]
+        conv_trials = jax.tree_map(
+            lambda x: _CORR_VARIABLE_TRIAL_DUR(jnp.atleast_2d(x), basis_matrix),
+            time_series, basis_matrix
```
billbrod, line 139: I think this is a typo right? Putting `basis_matrix` as an arg to tree_map means that it should be a pytree with the same structure as `time_series`, but including in the call the `_CORR_VARIABLE_TRIAL_DUR` means that it's being correlated with each leaf. shouldn't it be something like:
```suggestion
            lambda x, y: _CORR_VARIABLE_TRIAL_DUR(jnp.atleast_2d(x), y),
            time_series, basis_matrix
```

src/nemos/utils.py
```diff
@@ -63,7 +63,7 @@ def check_dimensionality(
 def convolve_1d_trials(
     basis_matrix: ArrayLike,
     time_series: Union[Iterable[NDArray], NDArray, Iterable[jnp.ndarray], jnp.ndarray],
-) -> List[jnp.ndarray]:
+) -> Union[Iterable[NDArray], NDArray, Iterable[jnp.ndarray], jnp.ndarray]:
```
billbrod, line 66: I'm confused by the `Iterable` in the type hinting here (both for return and `time_series`). Why do that instead of List?

src/nemos/utils.py
billbrod: Or it should be:
```suggestion
            lambda x: _CORR_VARIABLE_TRIAL_DUR(jnp.atleast_2d(x), basis_matrix),
            time_series
```

billbrod (changes requested):
I think there's a typo in the change, see my comment for details.

Also, should add tests.

BalzaniEdoardo (commented):
PR is ready on my side

src/nemos/utils.py
billbrod: Is this the current best practice for pytree annotation? That feels ... not great. Maybe we can be more specific about this? As in, we look for how we use it in our docs and library, and set it to that. Or make it clear somewhere else?

billbrod: I guess the docstring can be more specific

src/nemos/utils.py
```diff
@@ -39,106 +38,181 @@
...
     try:
-        # this should fail for variable trial length
-        time_series = jnp.asarray(time_series)
         if time_series.ndim != 3:
-            raise ValueError
-
-    except ValueError:
-        # convert each trial to array
-        time_series = [jnp.asarray(trial) for trial in time_series]
+            raise AttributeError
```
billbrod, line 89: This should be `ValueError` to match docstring ,right? And have the same error message as below.

src/nemos/utils.py
```diff
@@ -39,106 +38,181 @@
...
+        If any array in the pytree is empty (i.e., has a zero dimension).
+    """
+    if pytree_map_and_reduce(lambda x: 0 in x.shape, any, pytree):
+        raise ValueError(
+            f"Empty array provided. At least one of dimension in {pytree_name} is empty."
+        )
+
+
+def check_trials_longer_then_window_size(
+    time_series: Any, window_size: int, sample_axis: int
```
billbrod, line 121:
```suggestion
    time_series: Any, window_size: int, sample_axis: int = 0
```

src/nemos/utils.py
```diff
@@ -39,106 +38,181 @@
...
+    """
+    # convert to jax arrays
+    basis_matrix = jnp.asarray(basis_matrix)
+    time_series = jax.tree_map(jnp.asarray, time_series)
+
+    # check dimensions
+    check_convolve_input_dims(basis_matrix, time_series)
+
+    # check for empty inputs
+    check_non_empty(basis_matrix, "basis_matrix")
```
billbrod, line 194: in practice, when could this happen? if someone messes up indexing?

src/nemos/utils.py
```diff
@@ -252,37 +331,30 @@ def nan_pad_conv(
...
     start, end = adjust_indices[filter_type]
-
+    # convert to jax ndarray
+    conv_trials = jax.tree_map(jnp.asarray, conv_trials)
     try:
-        conv_trials = jnp.asarray(conv_trials)
         if conv_trials.ndim != 4:
-            raise ValueError(
-                "conv_trials must be an iterable of 3D arrays or a 4D array!"
+            raise AttributeError(
```
billbrod, line 338:
```suggestion
            raise ValueError(
```
Shouldn't this be a ValueError, to be consistent with the docstring, and the error with identical error message raised below?

billbrod (changes requested):
Generally looks good, just a couple consistency fixes.

- `utils.py` line 319, above the definition of `adjust_indices`, says "add a note about this in docstrings". I agree, there should be a note!

BalzaniEdoardo:
> Generally looks good, just a couple consistency fixes.
> 
>     * `utils.py` line 319, above the definition of `adjust_indices`, says "add a note about this in docstrings". I agree, there should be a note!

I added a note, admittedly not the clearest, but the idea of slicing the last index in causal filter, and first in anti-causal, is the same as the reason why we were getting rid of the the last sample in the head direction tutorial:
When using the convolution output as predictor, the very last entry of the convolution output, will be used to predict the next sample, however the for the next time step we don't have matching prediction (we run out of data, quoting nemos workshop head direction tutorial, the red rectangle is out of range).

It was hard for me to explain and it is a thing we need to do and want to hide from users, because it is annoying. can you try to rephrase and explain the concept better?

BalzaniEdoardo (commented):
ready for a second pass

billbrod (approved):

Issue #86, opened 2024-01-24
overwriting proximal gradient 


BalzaniEdoardo:
If one re-initialize the kwargs params for the solver without re-initializing the model, the prox operator is not re-set. this won't make jaxopt fail but will return weird results.

```
import nemos as nmo
model = nmo.glm.GLM(regularizer=nmo.regularizer.Lasso())
# this prints {'prox': <function Lasso._get_proximal_operator.<locals>.prox_op at 0x2bc6abee0>}
print(model.regularizer.solver_kwargs)

# if you set the solver kwargs after init you'll overwrite prox
model.regularizer.set_params(solver_kwargs=dict(tol=1))
```

Solution: 
You define a setter for solccer_kwargs at the level of the concrete class Lasso, GroupLasso, that updates the dictionary including the proximal operator

PR #87, opened 2024-01-25
issue #86: first attempt to fix the proximal operator overwriting

BalzaniEdoardo:
Started a PR to fix the edge case of user overwriting the proximal operator [issue #86](https://github.com/flatironinstitute/nemos/issues/86).

The solution adopted here should be discussed and it is a first attempt. I think overwriting user defined proximal operators is a good idea because each regularization type has its own specific operator.

Another option would be add/overwriting the prox in the setter for `solver_kwargs`; this has the benefit of raising the warning when the solver argument is passed and when fit is called. However, in this case it feels weird to add items to the user defined `solver_kwargs` (this is what happened so far and created the unexpected behavior we describe in the issue bug) .

### DIFF
--- a/src/nemos/regularizer.py
+++ b/src/nemos/regularizer.py
@@ -7,6 +7,7 @@ with various optimization methods, and they can be applied depending on the mode
 """
 import abc
 import inspect
+import warnings
 from typing import Any, Callable, List, Optional, Tuple, Union
 
 import jax.numpy as jnp
@@ -74,13 +75,10 @@ class Regularizer(Base, abc.ABC):
         **kwargs,
     ):
         super().__init__(**kwargs)
-        self._check_solver(solver_name)
-        self._solver_name = solver_name
+        self.solver_name = solver_name
         if solver_kwargs is None:
-            self._solver_kwargs = dict()
-        else:
-            self._solver_kwargs = solver_kwargs
-        self._check_solver_kwargs(self.solver_name, self.solver_kwargs)
+            solver_kwargs = dict()
+        self.solver_kwargs = solver_kwargs
 
     @property
     def solver_name(self):
@@ -146,7 +144,7 @@ class Regularizer(Base, abc.ABC):
             )
 
     def instantiate_solver(
-        self, loss: Callable, *args: Any, **kwargs: Any
+        self, loss: Callable, *args: Any, prox: Optional[Callable] = None, **kwargs: Any
     ) -> SolverRunner:
         """
         Instantiate the solver with the provided loss function.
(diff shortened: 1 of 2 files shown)


billbrod (commented):
In general, this solution looks good to me. I think the main question is whether to raise a warning and overwrite a user-defined prox, like here, or to raise an error. You know my preference for explicit errors. In this case, it's likely they've mis-understood something if they're setting prox directly, and it's easy to overlook the output of warnings. Also seems weird that we raise an error in one set of cases where the user has accidentally set prox (Unregularized, Ridge), but not in the other (Lasso, GroupLasso).

If you do decide to stick with warning, it's confusing that the message says it's overwriting prox and then if the user checks `model.solver_kwargs`, it's unchanged. I think I see why that's the case -- so that you can have the same behavior everytime instantiate_solver gets called, but it's still confusing (maybe add a comment explaining that bit). Another argument in favor of the exception, I think!

I also find the logic of the if statements confusing. If I understand correctly: the prox argument is never user-set, that comes from the sub-classes and so "prox is not None" is a proxy (haha) for "this class needs prox, and we've specified it", whereas "prox in self.solver_kwargs" tells you that the user tried to specify it, right? I think that's what the comments are getting at, but could be a bit clearer. I feel like I'd rather have the top if statement be `if prox in self.solver_kwargs`, and then change the error message based on whether `prox is None` or not. So it could look like:

```python
# then the user specified prox, which they shouldn't have done.
if "prox" in self.solver_kwargs:
    if prox is None:
        err_msg = "..."
    else:
        err_msg = "..."
    raise ValueError(err_msg)
# Copy the kwarg dictionary so that we get the same behavior every time this is called
solver_kwargs = self.solver_kwargs.copy()
solver_kwargs['prox'] = prox
```

That last line might not work, depending on whether Unregularized can take `prox=None`, in which case you just need another if statement

Also, the only case that comes to mind where someone might try to set the prox, and knows what they're doing, is because actually they're trying to change the regularizer. In that case, might be helpful to have the message additionally say something like "If you would like to use a different proximal operator, please create a new Regularizer class (link to docs)"

Also: probably worth testing that this all works when using `set_params` as well, since that's what go us into this situation.

BalzaniEdoardo:
> ```python
> if "prox" in self.solver_kwargs:
>     if prox is None:
>         err_msg = "..."
>     else:
>         err_msg = "..."
> ```

I kept the logic as is because proximal methods may be applied in context where the loss is differentiable. (i.e. when a user can opt to specify a proximal operator, while the our regularizer class does not in principle.

PR #88, opened 2024-01-25
updates PRNGKey to remove deprecation warning

billbrod:
In the readthedocs-built documentation for our [upcoming workshop](https://github.com/flatironinstitute/nemos-workshop-feb-2024), we were getting a lot of jax deprecation warnings on initial `nemos` import. That's due to [this change](https://jax.readthedocs.io/en/latest/jep/9263-typed-keys.html). I've updated nemos so we should no longer those deprecation warnings with most recent version of jax.

Specifically:
- calls to `jax.random.PRNGKey(x)` where `x` is an `int` have been updated to `jax.random.key(x)`
- type annotations have been updated from `jax.random.PRNGKeyArray` or `KeyArray` to `jax.Array`, which is apparently the correct type annotation.


BalzaniEdoardo (approved):
This looks good to me!

PR #96, opened 2024-02-07
Install instruction

BalzaniEdoardo:
Added install instruction. I suggested using `venv` since we are on pip. @billbrod  do you think this make sense? I need to check that the instruction work on windows. 
How should I do it, should change the `connect` workflow to run on windows and check on the runner?

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,21 +9,51 @@ NEural MOdelS, a statistical modeling framework for neuroscience.
 ## Disclaimer
 This is an alpha version, the code is in active development and the API is subject to change.
 
-## Setup
+## Installation
 
-To install, clone this repo and install using `pip`:
+To install `nemos` we recommend to creating and activating a virtual environment. You can create do so through
+`venv`.
 
-``` sh
-git clone git@github.com:flatironinstitute/nemos.git
-cd nemos/
-pip install -e .
+For  Mac and Linux,
+```shell
+python -m venv <DIR>
+source <DIR>/bin/activate
 ```
 
-If you have a GPU, you may need to install jax separately to get the proper
-build. The following has worked for me on a Flatiron Linux workstation: `conda
-install jax cuda-nvcc -c conda-forge -c nvidia`. Note this should be done
-without `jax` and `jaxlib` already installed, so either run this before the
-earlier `pip install` command or uninstall them first (`pip uninstall jax
-jaxlib`). See [jax docs](https://github.com/google/jax#conda-installation) for
-details (the `pip` instructions did not work for me).
+For Windows,
+```shell
+python -m venv <DIR>
+<DIR>\Scripts\activate
+```
+
+### CPU Install
+
+To install `nemos` in your environment run,
+
+``` shell
+pip install --upgrade pip
+pip install nemos@git+https://github.com/flatironinstitute/nemos.git
+```
+
+### GPU install
+If you have a GPU, you must need to install jax separately first to get the proper
+build. 
+In order to install `jax` and `jaxlib` for GPU, follow the instruction provided in the 
+[jax docs](https://jax.readthedocs.io/en/latest/installation.html).
+
+To check that `jax` for GPU is installed correctly, run python and type,
+
+```python
+import jax
+print(jax.devices())
+```
+This should print a list of devices that `jax` can access. If your `GPU` is listed, your installation 
+was successful.
+
+Once `jax` is installed correctly, continue with the usual steps for installing `nemos`,
+
+``` shell
+pip install --upgrade pip
+pip install nemos@git+https://github.com/flatironinstitute/nemos.git
+```
 


billbrod (changes requested):
I think `venv` is the right way to do this, but I think:

- A shorter version of this belongs on the readme and `docs/index.md`, with a link to a separate `docs/install.md` page that lists these longer instructions. Shorter version should just be "install using this pip command into your virtual environment", then the longer one says "we recommend venv", and walks people through it.
- The longer instructions should also include "developer instructions", which tells people to clone, use an editable install, and discusses the optional dependencies.
- For `venv`, the tricky thing for people is that you're placing it in a specific directory. I'd move that point to the top, and say something like "For [venv](link to venv docs), we need to install our virtual environment in a specific directory. For this example, we'll use `~/nemos` (or something), but use whatever makes sense to you." And then use that example path throughout
- I don't think we should tell people to upgrade `pip` in our instructions. Is it necessary?
- I like syntax highlighting the code blocks
- @ericthomson can you double-check instructions on windows for us?

You can see [plenoptic docs](https://plenoptic.readthedocs.io/en/latest/) for an example of the short/long versions that I'm talking about.

EricThomson:
> * I like syntax highlighting the code blocks
> * @EricThomson can you double-check instructions on windows for us?


Sure. At gate about to get on plane, but happy to do probably Monday if that’s ok (and I need to learn this non-conda virtual env tooling anyway good opportunity!!)

README.md
```diff
@@ -9,21 +9,74 @@ NEural MOdelS, a statistical modeling framework for neuroscience.
...
+
+2. Install `nemos` directly from the GitHub repository:
+
+   ```
+   pip install git+https://github.com/flatironinstitute/nemos.git
+   ```
+
+#### GPU Installation
+
+For systems equipped with a GPU, a specific installation process is required to utilize the GPU with `nemos`.
```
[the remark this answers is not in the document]
BalzaniEdoardo: if you don't specify the cuda or other with pip install jax[cuda-or-similar] it defaults install [CPU](https://jax.readthedocs.io/en/latest/installation.html#nvidia-gpu)

docs/installation.md
```diff
@@ -0,0 +1,77 @@
...
+    ```
+
+**For Windows:**
+
+1. Open a command prompt.
+
+2. Execute the commands below to create and activate the virtual environment:
+    ```bash
+    python -m venv C:\Users\username\python_venvs\nemos
+    C:\Users\username\python_venvs\nemos\Scripts\activate
```
EricThomson, line 29: I had to cd into the virtual env and then activate.
```
cd C:\Users\username\python_venvs\nemos\
.\Scripts\activate
```

docs/index.md
```diff
@@ -22,6 +22,15 @@ lesson](https://compneuro.neuromatch.io/tutorials/W1D3_GeneralizedLinearModels/s
...
 tutorial](https://www.youtube.com/watch?v=NFeGW5ljUoI&t=424s).
 
+## Installation
+Run the following `pip` command in your virtual environment.
+
+```bash
+pip install git+https://github.com/flatironinstitute/nemos.git
+```
+
+For more comprehensive instructions, including specifics for GPU users and developers, refer to [this section](installation.md).
```
billbrod, line 32:
```suggestion
For more comprehensive instructions, including specifics for GPU users and developers, refer to [the installation page](installation.md).
```

docs/installation.md
```diff
@@ -0,0 +1,77 @@
+## Prerequisites
+
+Ensure you have Python version 3.9 or above installed on your system.
```
billbrod, line 3: I think we're only testing version 3.10 right now -- do we want to add 3.9 to the tests? have you checked with 3.9?

docs/installation.md
```diff
@@ -0,0 +1,77 @@
...
+
+We recommend creating and activating a Python virtual environment using [`venv`](https://docs.python.org/3/library/venv.html) before installing `nemos`. This practice helps manage dependencies and avoid conflicts with other Python packages.
+
+For `venv`, create your virtual environment in a specific directory. This example uses `~/python_venvs/nemos` for Linux/Mac and `C:\Users\username\python_venvs\nemos` for Windows; in general, free to use whatever directory works best for you.
```
billbrod, line 7: People will think that `Users\username` is the path they should use, I'd replace it with `%HOMEPATH%`, which I think is the windows equivalent of `$HOME` ([stackoverflow](https://superuser.com/questions/1319038/how-to-set-home-variable-so-that-cmd-find-the-path))

@EricThomson  can you check that the commands in this section work with `%HOMEPATH%\python_venvs\nemos\`, just like that?

docs/installation.md
```diff
@@ -0,0 +1,77 @@
...
+    
+**Note: Replace `username` with your actual Windows username.*
+
+## Installation Steps
+
+### CPU Installation
+
+To install `nemos` on a system without a GPU, run this command from within your activated environment, 
+
+ ```shell
```
billbrod, line 40:
```suggestion
 ```bash
```

docs/installation.md
```diff
@@ -0,0 +1,77 @@
...
+    python -m venv ~/python_venvs/nemos
+    source ~/python_venvs/nemos/bin/activate
+    ```
+
+**For Windows:**
+
+1. Open a command prompt.
+
+2. Execute the commands below to create and activate the virtual environment:
+    ```bash
```
billbrod, line 27:
```suggestion
    ```
```
I'm not sure what highlighting should be used for the windows prompt but it shouldn't be bash

docs/installation.md
```diff
@@ -0,0 +1,77 @@
...
+    python -m venv C:\Users\username\python_venvs\nemos
+    C:\Users\username\python_venvs\nemos\Scripts\activate
+    ```
+    
+**Note: Replace `username` with your actual Windows username.*
```
billbrod, line 32: see comment above -- if we get the HOMEPATH variable working, I think that's the way to do it

docs/installation.md
```diff
@@ -0,0 +1,77 @@
...
+
+## Installation Steps
+
```
billbrod, line 35:
```suggestion
After creating you virtual environment, follow one of the following sections below, depending on whether you need GPU support or not:
```
I think some transition is necessary

billbrod (changes requested):
Couple small changes still. Main thing is checking whether we can use the `%HOMEPATH%` variable on windows command line.

docs/installation.md
```diff
@@ -0,0 +1,77 @@
...
+### CPU Installation
+
+To install `nemos` on a system without a GPU, run this command from within your activated environment, 
+
+ ```shell
+ pip install git@github.com:flatironinstitute/nemos.git
```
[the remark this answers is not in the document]
BalzaniEdoardo: wow!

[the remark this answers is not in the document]
BalzaniEdoardo: thanks Eric!

docs/installation.md
BalzaniEdoardo: I'll add 3.9 to the CI

docs/installation.md
BalzaniEdoardo: I checked and it works

PR #97, opened 2024-02-08
Drop nan

BalzaniEdoardo:
In this PR I changed the behavior of the GLM class in the presence of NaNs or Infs. 

This will allow convolve + nan padding which keep the time sample axis intact. Added benefit: easier to convert in and out from pynapple.

New behavior for preprocessing fit:
- nan will be dropped
- UserWarning raised

The behavior for preprocessing simulate will be unchanged (exception if provided predictor contains nans). Here we simulate each time point one at the time recursively, and so dropping the nan would violate this assumption of equi-spaced samples.


src/nemos/base_class.py
```diff
@@ -384,8 +384,23 @@ def _preprocess_fit(
...
+        valid_y, err_y = check_invalid_entry(y, "y")
+
+        # warn the user of the dropped samples.
+        if err_x:
+            message = err_x.join(" Dropping corresponding samples.")
+            warnings.warn(message=message, category=UserWarning)
+        if err_y:
+            message = err_y.join(" Dropping corresponding samples.")
+            warnings.warn(message=message, category=UserWarning)
+
```
billbrod, line 397: This structure feels weird: why have the check and drop happen in the utility function but the warning here? Why not do them all in the utility function? (would clean this code up and avoid the potential problem of developers forgetting to do anything with the message from `check_invalid_entry` ).

billbrod: Also, message should make clear that we're dropping the same samples from *both* arrays

billbrod: (that message might need to be in this function)

src/nemos/utils.py
```diff
@@ -400,9 +401,7 @@ def row_wise_kron(a, c):
     return K
 
 
-def check_invalid_entry(
-    pytree: Union[FeaturePytree, jnp.ndarray], pytree_name: str
-) -> None:
+def check_invalid_entry(pytree: Any, pytree_name: str) -> Tuple[Any, Union[None, str]]:
```
billbrod, line 404: like I mentioned in other comment, I don't think it's great to have this just return the error message, without doing anything with it -- too easy to forget  to raise it.

Alternative proposal: we have a `_check_invalid_entry` that is basically this and then a `warn_invalid_entry` and `error_invalid_entry`, which do the two versions of this: checks/warns/drops/returns valid entries, and checks/raise exception.

Thoughts?

src/nemos/utils.py
```diff
@@ -412,17 +411,45 @@ def check_invalid_entry(
...
         The array name.
 
-    Raises
-    ------
-        - ValueError: If any entry of `pytree` is either NaN or inf.
+    Returns
+    -------
+    is_valid:
+        A pytree with leaves flat boolean arrays of shape (n_sample_points, ) with
+        the invalid indices (samples with Nans or Infs).
```
billbrod, line 418: This has the same structure as `pytree` right? so it should say that: "Boolean pytree with same structure as `pytree`, with values specifying whether corresponding values in `pytree` are valid or not".

src/nemos/base_class.py
billbrod: just realized that the dropping does happen here -- see below for proposal on some restructure of this

src/nemos/utils.py
billbrod: Or, if `warn_invalid_entry` shouldn't do dropping, then a separate function for the dropping? probably should be a helper function, since it will be useful.

billbrod (changes requested):
Looks good, some thoughts:

- I think some reworking of how the check/warn/drop is necessary. Not sure the proposal I put in is right, happy to talk through it.
- tests should include the multidimensional leaf case, since that's what we actually use.
- should test that we can fit a GLM that includes some nans
- what happens with predict? don't we want it to behave similarly to fit? (dropping and warning about nans, and then giving an output), because we regularly fit and then predict on same model matrix for evaluation.
    - once we do that, should test that predict gives something with different shape than input if input has some nans.
- what do we do if entire column/row is full of invalid values? I feel like we should still raise an error in that case, right? in particular, that would mean corresponding coef/intercept would be completely unconstrained?

src/nemos/utils.py
BalzaniEdoardo: The description is wrong. It returns a flat vector of boolean of length that of the sample axis. What the code does is the following:
```python
# this creates the pytree you are describing
is_valid = jax.tree_map(lambda x: check_not_inf(x) & check_not_nan(x), pytree)
# this gets the leaves of that tree and apply a logical and because we want to check all the variables
is_valid = reduce(jnp.logical_and, jax.tree_util.tree_leaves(is_valid))
```
That's the vector returned, and it should be used to filter the time axis

BalzaniEdoardo:
> Looks good, some thoughts:
> 
>     * I think some reworking of how the check/warn/drop is necessary. Not sure the proposal I put in is right, happy to talk through it.
That sounds good

>     * tests should include the multidimensional leaf case, since that's what we actually use.
I'll add that
 
>     * should test that we can fit a GLM that includes some nans
this happens in test_fit_param_values

>     * what happens with predict? don't we want it to behave similarly to fit? (dropping and warning about nans, and then giving an output), because we regularly fit and then predict on same model matrix for evaluation.

I think we should keep the predict without the dropping, we saw how bad it gets when we mess with the number of samples, imagine to put the output back in pynapple after we remove samples... The fit will not return anything with a time axis, therefore for the user, dropping nan has no consequences. With the predict, dropping nans would change the number of samples, and I don't want that. What we can do is raise a warning, Nan in predict will result in nan in the rate.
       
>       * once we do that, should test that predict gives something with different shape than input if input has some nans.
I think we should check the expected beahavior, and add the warning

>     * what do we do if entire column/row is full of invalid values? I feel like we should still raise an error in that case, right? in particular, that would mean corresponding coef/intercept would be completely unconstrained?
This is a good point, it should raise an error

src/nemos/utils.py
BalzaniEdoardo: we need to merge the invalid entries of X and y, should we pass both to the checker? I think we should and I think it should implement something like _check_invalid_entry(*args), _warn_invalid_entry(*args), even though I would like the error message to be variable specific

BalzaniEdoardo: New design:
1) func check raises error
2) func drop invalid (warn and drop) receives all *args, pass (arg, argname).
3) fit needs drop and predict none of the two. Test that predict with nan has expected behavior

src/nemos/utils.py
```diff
@@ -248,120 +254,268 @@ def _pad_dimension(
...
-    The `adjust_indices` dictionary specifies the start and end indices for
-    slicing the input data based on the filter type.
+    The location of the NaN-padding depends on the value of `predictor_causality`.
+
+    - `'causal'`: `pad_size-1` NaNs are placed at the beginning of the
+      temporal dimension.
+    - `'acausal'`: `floor(pad_size/2)` NaNs are placed at the beginning of
+      the temporal dimensions, `ceil(pad_size/2)` placed at the end.
+    - `'anti-causal'`: `pad_size-1` NaNs are placed at the end of the
+      temporal dimension.
```
billbrod, line 331:
```suggestion
    - `'causal'`: `pad_size` NaNs are placed at the beginning of the
      temporal dimension.
    - `'acausal'`: `floor(pad_size/2)` NaNs are placed at the beginning of
      the temporal dimensions, `ceil(pad_size/2)` placed at the end (for `pad_size`
      total NaNs).
    - `'anti-causal'`: `pad_size` NaNs are placed at the end of the
      temporal dimension.
```

src/nemos/utils.py
```diff
@@ -248,120 +254,268 @@ def _pad_dimension(
...
+        raise ValueError(
+            f"pad_size must be a positive integer! Pad size of {pad_size} provided instead!"
+        )
+
+    causality_choices = ["causal", "acausal", "anti-causal"]
+    if predictor_causality not in causality_choices:
+        raise ValueError(
+            f"predictor_causality must be one of {causality_choices}. {predictor_causality} provided instead!"
+        )
+    if predictor_causality == "acausal" and (pad_size % 2 == 0):
```
billbrod, line 344:
```suggestion
    if predictor_causality == "acausal" and (pad_size % 2 == 1):
```

src/nemos/utils.py
```diff
@@ -248,120 +254,268 @@ def _pad_dimension(
...
+        warnings.warn(
+            "With acausal filter, pad_size should probably be even,"
```
billbrod, line 346: Should catch this warning in `create_convolutional_predictor` and reword it so it matches the arguments there (`basis_matrix.shape[0] should be odd` or similar)

billbrod: note that it's both the argument name and even -> odd that need to change

src/nemos/utils.py
```diff
@@ -248,120 +254,268 @@ def _pad_dimension(
...
+           n_features)`
+    predictor_causality:
+        Causality of this predictor, which determines how to shift time_series.
+
+    Returns
+    -------
+    shifted_time_series :
+        `time_series` that has been shifted. shift is determined by value of
+        `predictor_causality` (see Notes), and `shifted_time_series` structure is
+        determined by that of `time_series`:
```
billbrod, line 405:
```suggestion
        `predictor_causality` (see Notes), and `shifted_time_series` structure
        matches that of `time_series`:
```

src/nemos/utils.py
```diff
@@ -248,120 +254,268 @@ def _pad_dimension(
...
+        Causality of this predictor, which determines where padded values are added.
 
     Returns
     -------
-    :
-        A pytree with the padded trials that has the same structure as conv_trials.
+    padded_conv_time_series :
+        `conv_time_series` with NaN padding. NaN location is determined by
+        value of `predictor_causality` (see Notes), and `padded_conv_time_series`
+        structure is determined by that of `conv_time_series`:
```
billbrod, line 302:
```suggestion
        value of `predictor_causality` (see Notes), and `padded_conv_time_series`
        structure matches that of `conv_time_series`:
```

src/nemos/utils.py
```diff
@@ -248,120 +254,268 @@ def _pad_dimension(
...
+def create_convolutional_predictor(
+    basis_matrix: ArrayLike,
+    time_series: Any,
+    predictor_causality: Literal["causal", "acausal", "anti-causal"] = "causal",
+    shift: bool = True,
+):
+    """Create predictor by convolving basis_matrix with time_series.
+
+    To create the convolutional predictor, three steps are taken, which calls
+    three functions in sequence. See their docstrings for more details about
```
billbrod, line 474:
```suggestion
    To create the convolutional predictor, three steps are taken, each of 
    calls a single function. See their docstrings for more details about
```

src/nemos/utils.py
```diff
@@ -248,120 +254,268 @@ def _pad_dimension(
...
+    each step.
+
+    - Convolve `basis_matrix` with `time_series` (function: `convolve_1d_trials`)
+    - Pad output with `basis_matrix.shape[0]`-1 NaNs, with location based on
```
billbrod, line 478:
```suggestion
    - Pad output with `basis_matrix.shape[0]-1` NaNs, with location based on
```

src/nemos/utils.py
```diff
@@ -248,120 +254,268 @@ def _pad_dimension(
...
+        1. A single array of 3-dimensions, `(n_trials, n_time_bins, n_neurons)`.
+        2. Any pytree with 2-dimensional arrays, `(n_time_bins, n_neurons)`, as
+           leaves. Note that neither `n_time_bins` nor `n_neurons` need to be
+           identical across leaves.
+    predictor_causality:
+        Causality of this predictor, which determines where padded values are
+        added and how the predictor is shifted.
+    shift :
+        Whether to shift predictor based on causality (only valid if
+        predictor_causality is not acausal)
```
billbrod, line 501:
```suggestion
        `predictor_causality != 'acausal'`)
```

billbrod (approved):
- The tutorial will need more work, but I think that can be a separate PR. (In particular, we need to explain why acasual has `shift=False`. Partly that requires deciding who this tutorial is for: right now, it's largely for us, to show what decisions were made, in which case it doesn't belong at this location. if it's for users, needs a bit more exposition.

For this PR:
- I think it should be made explicit that `convolve_1d_trials` will remove time points, `nan_pad` will add `pad_size` time points, and `shift_time_series` and `create_convolutional_predictor` will not change the shape at all. This info is present in the docstring, but I worry it gets lost. Worth showing to somebody else (Alex? Sam? Isabel? Erik?) and getting their opinion
- `test_validation` is mispelled (it's only a single `l`)
- Still need to raise error if all entries in row/column are invalid
- Still need to raise warning in predict if there's a nan

PR #119, opened 2024-03-22
Hpc data

gviejo:
Adding `plot_05_place_cells.py`.

Data is a recording of  hippocampal neurons when a rat is traversing a linear track.
It showcases 

- pynapple data analysis
- Basis instantiation and evaluation
- Model fitting
- Model scoring and comparison


docs/neural_modeling/plot_05_place_cells.py
```diff
@@ -0,0 +1,410 @@
...
+glm_pf = nap.compute_1d_tuning_curves_continuous(predicted_rate, position, 50)
+glm_pos_theta, xybins = nap.compute_2d_tuning_curves_continuous(
+    predicted_rate, data, 30
+)
+glm_speed = nap.compute_1d_tuning_curves_continuous(predicted_rate, speed, 30)
+
+# %%
+# Let's display both tuning curves together.
+
+plt.figure()
```
BalzaniEdoardo, line 256: maybe we should move this in the utils functions since we have other plotting code there too

docs/neural_modeling/plot_05_place_cells.py
```diff
@@ -0,0 +1,410 @@
...
+    "position": (position,),
+    "position + speed": (position, speed),
+    "position + phase": (position, theta),
+    "position + phase + speed": (position, theta, speed),
+}
+
+# %%
+# In a loop, we can (1) evaluate the basis, (2), fit the model, (3) compute the score and (4) predict the firing rate. For evaluating the score, we can define a train set of intervals and a test set of intervals.
+
+train_iset = position.time_support[
```
BalzaniEdoardo, line 310: Since we have a bunch of disjoint epochs, we could do: position.time_support[::2] for train and position.time_support[1::2] as a better splitting strategy, so that we sample more "uniformily" in time.

docs/neural_modeling/plot_05_place_cells.py
```diff
@@ -0,0 +1,410 @@
...
+        "position": nap.compute_1d_tuning_curves_continuous(
+            predicted_rates[m], position, 50
+        ),
+        "speed": nap.compute_1d_tuning_curves_continuous(
+            predicted_rates[m], speed, 30, minmax=(0, 100)
+        ),
+    }
+
+
+def plot_tuning_curves(axis, tc, m):
```
BalzaniEdoardo, line 377: This probably goes into the utils too. you'll need to pass all the arguments, it is using some globals which would be  confusing if a person tries to use it outside the script.

docs/neural_modeling/plot_05_place_cells.py
```diff
@@ -0,0 +1,410 @@
...
+    plt.subplot(gs[0, 1])
+    plt.plot(tc_speed[neuron], "--")
+    plt.plot(tc["speed"][0])
+    plt.xlabel("Speed (cm/s)")
+
+
+fig = plt.figure(figsize=(8, 4))
+outer_grid = fig.add_gridspec(2, 2)
+for i, m in enumerate(models):
+    plot_tuning_curves(outer_grid[i // 2, i % 2], tuning_curves[m], m)
```
BalzaniEdoardo, line 396: Fits looks poor, I think it is because you did not restrict to the right interval set before computing the tuning curves from the raw spikes. I am pretty sure it would look much better if you did. We should restrict before fitting and recompute tuning before fitting. 

Another way could be computing the firing rate over the whole recording, and then compute the tuning curve on the desired iset for both spikes and rates at the same time.

Like in your loop
```python
for m in models:
    print("1. Evaluating basis : ", m)
    X = models[m].evaluate(*features[m])
    
    # ... rest of the loop
    # ...

    print("4. Predicting rate")
    # DO NOT RESTRICT HERE
    predicted_rates[m] = glm.predict(X[:, np.newaxis, :]) / bin_size
```

After the loop you can compute the tuning passing the epoch of the test set

```python

 tuning_curves_raw = {
        "position": nap.compute_1d_tuning_curves(
            spikes, position, 50, ep=test_iset
        ),
        "speed": nap.compute_1d_tuning_curves_continuous(
            predicted_rates[m], speed, 30, minmax=(0, 100), ep=test_iset
        ),
    }

tuning_curves_model = {}

for m in models:
    tuning_curves_model[m] = {
        "position": nap.compute_1d_tuning_curves_continuous(
            predicted_rates[m], position, 50, ep=test_iset
        ),
        "speed": nap.compute_1d_tuning_curves_continuous(
            predicted_rates[m], speed, 30, minmax=(0, 100), ep=test_iset
        ),
    }
  
```

docs/neural_modeling/plot_05_place_cells.py
```diff
@@ -0,0 +1,410 @@
...
+# There are a lot of neurons but for this analysis, we will focus on one neuron only.
+
+neuron = 175
+
+# %%
+# To show the theta phase precession, we can also display the spike as a function of both position and theta. In this case, we use the function `value_from` from pynapple.
+
+theta_pos_spikes = spikes[neuron].value_from(data)
+
+plt.figure()
```
BalzaniEdoardo, line 118: consider moving this in the utils

BalzaniEdoardo (changes requested):
This in general is very good! I merged main into this to keep it updated.

I had a couple of extra points which are mostly:

1. if we want to be consistent we should move lengthy/complex plotting code into the utils
2. I would avoid using globals in defs that can be confusing
3. This is the most important, the tuning curve looks of but I suspect they are not, the problem is that you did not restrict the tuning computed from raw spikes to the test set. We should do that to show that the fits are actually good

docs/neural_modeling/plot_05_place_cells.py
```diff
@@ -0,0 +1,410 @@
...
+plt.ylabel("Theta Phase (rad)")
+
+plt.tight_layout()
+
+
+# %%
+# ## Speed modulation
+# The speed at which the animal traverse the field is not homogeneous. Does it influence the firing rate of hippocampal neurons? We can compute tuning curves for speed as well as average speed across the maze.
+
+speed = [
```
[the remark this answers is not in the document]
gviejo: Done

docs/neural_modeling/plot_05_place_cells.py
```diff
@@ -0,0 +1,410 @@
...
+    np.pad(np.abs(np.diff(data["position"].get(s, e))), [0, 1], mode="edge") * data.rate
+    for s, e in data.time_support.values
+]
+speed = nap.Tsd(t=data.t, d=np.hstack(speed), time_support=data.time_support)
+
+tc_speed = nap.compute_1d_tuning_curves(spikes, speed, 20)
```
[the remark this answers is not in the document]
gviejo: Done

docs/neural_modeling/plot_05_place_cells.py
gviejo: Done

docs/neural_modeling/plot_05_place_cells.py
gviejo: Done

docs/neural_modeling/plot_05_place_cells.py
gviejo: Done

docs/neural_modeling/plot_05_place_cells.py
gviejo: Done

docs/neural_modeling/plot_05_place_cells.py
gviejo: Done

BalzaniEdoardo (approved):
looks good!

PR #143, opened 2024-05-08
127 batching

BalzaniEdoardo:
In this PR we will add `update` method to GLM that enables batching.

## GLM class

- Store initialize, update and run callable for the solver as read-only properties
- Create an `init_update` that initializes the solver states and, if no init param are provided, the params
- Create an `update` method that will run an update of the algorithm

## Regularizer class

- Updated the instantiate solver so that it returns 3 callable (the initializer, the update and the runner), this will take care of setting the appropriate regularization hyperparameters, so that the call to the initializer/runner/solver is always func(params, X, y), for any regularization type and algorithm


src/nemos/regularizer.py
```diff
@@ -32,6 +32,29 @@
     jaxopt.OptStep,
 ]
 
+SolverInit = Callable[
+    [
+        Tuple[
```
gviejo, line 37: Any

src/nemos/regularizer.py
```diff
@@ -32,6 +32,29 @@
...
+        ],  # Model parameters (for now tuple, eventually pytree)
+        jnp.ndarray,  # Predictors (i.e. model design for GLM)
+        jnp.ndarray,
+    ],  # Output (neural activity)
+    NamedTuple,
+]
+
+SolverUpdate = Callable[
+    [
+        Tuple[
```
gviejo, line 48: Any

gviejo (approved):

PR #145, opened 2024-05-21
Identifiability

BalzaniEdoardo:
## Take care of model identifiability

Added a flag to basis that, if True, checks that the design matrix including the intercept, is full rank; if not, it drops a column and mean center the data.

This will take care of BSplines, and CyclicBslplines, which sum to 1, and may result in a family of equivalent parameters all maximizing the likelihood.


gviejo (approved):
Good

PR #155, opened 2024-06-03
Fix transformer

BalzaniEdoardo:
Inherit Base in "TransformerBasis".

### DIFF
--- a/src/nemos/basis.py
+++ b/src/nemos/basis.py
@@ -86,6 +86,16 @@ class TransformerBasis(Base):
     """
 
     def __init__(self, basis: Basis):
+        self.basis = basis
+
+    @property
+    def basis(self):
+        return self._basis
+
+    @basis.setter
+    def basis(self, basis: Basis):
+        if not isinstance(basis, Basis):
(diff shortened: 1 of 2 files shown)


billbrod (commented):
Why not always have the Basis be a transformer? Does it add overhead or something else?

BalzaniEdoardo:
> Why not always have the Basis be a transformer? Does it add overhead or something else?

This needs more work, @bagibence is working on it. 

The issue is that we are breaking the API for `BaseTransformer` of sklearn, which does not allow variable positional or keyword arguments. 

The variable args are inputs to the convolution called in `compute_features` of `Basis`. The solution is to accept a conv_kwarg argument in the`Basis.__init__` defaulted to None with dict as type.

Additionally, he is rewriting the `__getitem__` to resurface the attributes of the basis, this will allow you to perform stuff like

```python
grid = dict(n_basis_funcs=[5, 6])  # grid search the num of basis
pipe = pipeline.Pipeline([("eval", bas), ("fit", model)])
cls = GridSerachCV(pipe, grid).fit(x, Y)
```

While with the current`TransformerBasis` the only parameter that can be get and set is the `TransformerBasis.basis`.

BalzaniEdoardo:
Resolved by #169

Issue #149, opened 2024-05-30
Notes on Fit injected current demo (dev branch)


EricThomson:
Overall this is great, not a lot of comments that aren't just cosmetic. The way I'm approaching this one is pretending it's my first attempt to start really digging in with nemos after going over the general intro demo (this is not hard to pretend, it's pretty much my case). And I'm coming with mindset of an experimentalist who wants to use nemos who knows their way around Python, but doesn't have much familiarity with pynapple yet. 

A few minor notes.

1. At one place it says "throughough this workshop". It also says "the dataset contains several different pynapple objects, which we discussed earlier today." This  maybe just amend to "which we will show how to use soon."

2. >  For now, the only possible observation model is the Poisson, though this may change in future releases. They can be found within nemos.observation_models.

    Now there are Gamma distributions right? So this can be amended.

3. There are some very minor heading glitches (e.g., the heading for `generate and examine model predictions` is wonky for me). So I'd amend to fix those. 

But overall this is great and I don't really have much to say. :100: 

If my suggestions above sound good I can make a PR soon.

billbrod:
Thanks Eric! Yes, we should definitely change the things you mentioned in points 1 and 2. For point 3: those aren't actually headings, but they're the remains of notes that I used during the workshop, which we forgot to remove. If you look at the raw `.py` file (`docs/neural_modeling/plot_01_current_injection.py`), you'll see that those points are wrapped in some html, `<div class="notes">`. They should just be deleted.

Do you want to open a PR making these changes? I can also do so.

EricThomson:
Billy thanks for the clarification on some of those headers I can check out the raw `py` files to see the intent :smile: I can work on a PR and then you can maybe help in review?

billbrod:
That sounds great! Just ping me when you open the PR

PR #157, opened 2024-06-04
Current injection minor updates

EricThomson:
These are mostly cosmetic updates and this resolves issue #149 

It's already very good, nothing substantive mostly formatting and word choice. 

I had to remove the kwarg `lazy_loading` in two files as it was stopping the docs from building for me. 

One thing @billbrod  near the end, it says:

> In particular, if your GLM includes auto-regressive inputs (e.g., neurons are connected to themselves or each other), simulate can behave poorly.

I'm not sure what the intended meaning is here.

### DIFF
--- a/docs/neural_modeling/plot_01_current_injection.py
+++ b/docs/neural_modeling/plot_01_current_injection.py
@@ -33,9 +33,9 @@ of this notebook, we'll use only the input current and the recorded spikes
 displayed in the first and third rows.
 
 First, let us see how to load in the data and reproduce the above figure, which
-we'll do using [pynapple](https://pynapple-org.github.io/pynapple/), which
-we'll use throughout this workshop, as it simplifies handling this type of
-data. After we've explored the data some, we'll introduce the Generalized
+we'll do using [Pynapple](https://pynapple-org.github.io/pynapple/). We will rely on 
+pynapple throughout this notebook, as it simplifies handling this type of
+data (we will explain the essentials of pynapple as they are used). After we've explored the data some, we'll introduce the Generalized
 Linear Model and how to fit it with NeMoS.
 
 ## Learning objectives {.keep-text}
@@ -104,8 +104,8 @@ print(data)
 
 # %%
 #
-# The dataset contains several different pynapple objects, which we discussed
-# earlier today. Let's see how these relate to the data we visualized above:
+# The dataset contains several different pynapple objects, which we will explore soon. 
+# Let's see how these relate to the data we visualized above:
 #
 # ![Annotated view of the data we will analyze.](../../assets/allen_data_annotated.gif)
 # <!-- this gif created with the following imagemagick command: convert -layers OptimizePlus -delay 100 allen_data_annotated-units.svg allen_data_annotated-epochs.svg allen_data_annotated-stimulus.svg allen_data_annotated-response.svg -loop 0 allen_data_annotated.gif -->
@@ -284,10 +284,11 @@ count
 # Now, let's convert the binned spikes into the firing rate, by smoothing them
 # with a gaussian kernel. Pynapple again provides a convenience function for
 # this:
-
+#
 # the inputs to this function are the standard deviation of the gaussian and
 # the full width of the window, given in bins. So std=50 corresponds to a
 # standard deviation of .05 seconds, total size of the filter, 0.05 sec *20 = 1 sec.
+
 firing_rate = count.smooth(std=0.05, size_factor=20)
 # convert from spikes per bin to spikes per second (Hz)
 firing_rate = firing_rate / bin_size
@@ -302,12 +303,9 @@ print(type(firing_rate))
 # %%
 #
 # Now that we've done all this preparation, let's make a plot to more easily
-# visualize the data.
+# visualize the data. We're hiding the details of the plotting function for the purposes of this
+# tutorial, but you can find it in [the source code](https://github.com/flatironinstitute/nemos/blob/development/docs/neural_modeling/examples_utils/plotting.py) if you are interested.
 
-# we're hiding the details of the plotting function for the purposes of this
-# tutorial, but you can find it in the associated github repo if you're
-# interested:
-# https://github.com/flatironinstitute/nemos-workshop-feb-2024/blob/binder/src/workshop_utils/plotting.py
 plotting.current_injection_plot(current, spikes, firing_rate)
 
 # %%
@@ -423,7 +421,7 @@ print(f"count sampling rate: {count.rate/1000:.02f} KHz")
 #
 # Because we only have a single predictor feature, we'll use
 # [`np.expand_dims`](https://numpy.org/doc/stable/reference/generated/numpy.expand_dims.html)
-# to handle .
+# to ensure it is a 2d array.
 
 
 # add two dimensions for axis 1.
@@ -443,7 +441,7 @@ print(f"count shape: {count.shape}")
 #
 #     In NeMoS, we always fit Generalized Linear Models to a single neuron at a
 #     time. We'll discuss this more in the [following
-#     tutorial](../02_head_direction/), but briefly: you get the same answer
+#     tutorial](../plot_02_head_direction/), but briefly: you get the same answer
 #     whether you fit the neurons separately or simultaneously, and fitting
 #     them separately can make your life easier.
 #
@@ -488,10 +486,10 @@ print(f"count shape: {count.shape}")
 #     different results.)
 #
 # - Observation model: this object links the firing rate and the observed
-#   spikes, describing the distribution of neural activity (and thus changing
-#   the log-likelihood). For now, the only possible observation model is the
-#   Poisson, though this may change in future releases. They can be found
-#   within `nemos.observation_models`.
+#   data (in this case spikes), describing the distribution of neural activity (and thus changing
+#   the log-likelihood). For spiking data, we use the Poisson observation model, but
+#   we discuss other options for continuous data 
+#   in [the calcium imaging analysis demo](../plot_06_calcium_imaging/).
 #
 # For this example, we'll use an un-regularized LBFGS solver. We'll discuss
 # regularization in a later tutorial.
@@ -505,10 +503,8 @@ print(f"count shape: {count.shape}")
 #     often less sensitive to step-size. Try other solvers to see how they
 #     behave!
 #
-# <div class="notes">
-#   - GLM objects need regularizers and observation models
-# </div>
 
+# Initialize the model w/regularizer and solver
 model = nmo.glm.GLM(regularizer=nmo.regularizer.UnRegularized(solver_name="LBFGS"))
 
 # %%
@@ -545,10 +541,6 @@ print(f"intercept_ shape: {model.intercept_.shape}")
 # the smoothed spike train. By calling `predict()` we can get the model's
 # predicted firing rate for this data. Note that this is just the output of the
 # model's linear-nonlinear step, as described earlier!
-#
-# <div class="notes">
-#   - generate and examine model predictions.
-# </div>
 
 # mkdocs_gallery_thumbnail_number = 4
 
@@ -592,10 +584,6 @@ plotting.current_injection_plot(current, spikes, firing_rate,
 #
 # To get a better sense, let's look at the mean firing rate over the whole
 # period:
-#
-# <div class="notes">
-#   - what do we see?
-# </div>
 
 # compare observed mean firing rate with the model predicted one
 print(f"Observed mean firing rate: {np.mean(count) / bin_size} Hz")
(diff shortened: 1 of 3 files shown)


docs/neural_modeling/plot_01_current_injection.py
```diff
@@ -104,8 +104,8 @@
 
 # %%
 #
-# The dataset contains several different pynapple objects, which we discussed
-# earlier today. Let's see how these relate to the data we visualized above:
+# The dataset contains several different pynapple objects, which we will explore soon. 
+# Let's see how these relate to the data we visualized above:
```
billbrod, line 108: Do you feel like there's enough of an explanation about the pynapple objects in this notebook? In the workshops, we have some sessions before this just focusing on pynapple. I think maybe we should at least drop a link to the pynapple docs, like "(see [pynapple docs]() for more details, if interested)". or something

docs/neural_modeling/plot_01_current_injection.py
```diff
@@ -302,12 +303,9 @@
 # %%
 #
 # Now that we've done all this preparation, let's make a plot to more easily
-# visualize the data.
+# visualize the data. We're hiding the details of the plotting function for the purposes of this
+# tutorial, but you can find it in [the source code](https://github.com/flatironinstitute/nemos/blob/development/docs/neural_modeling/examples_utils/plotting.py) if you are interested.
```
billbrod, line 307: I want to do something else with this `examples_utils` code, so I've opened an issue about it #158

docs/neural_modeling/plot_01_current_injection.py
```diff
@@ -284,10 +284,11 @@
 # Now, let's convert the binned spikes into the firing rate, by smoothing them
 # with a gaussian kernel. Pynapple again provides a convenience function for
 # this:
-
+#
```
billbrod, line 287: This changes how the following lines are rendered, from comment to markdown. I think they belong as comments, because they're just explaining the specific details of the arguments (so they're not conceptually important). What do you think?

docs/neural_modeling/plot_01_current_injection.py
```diff
@@ -284,10 +284,11 @@
...
 # the inputs to this function are the standard deviation of the gaussian and
 # the full width of the window, given in bins. So std=50 corresponds to a
 # standard deviation of .05 seconds, total size of the filter, 0.05 sec *20 = 1 sec.
```
billbrod, line 290:
```suggestion
# the inputs to this function are the standard deviation of the gaussian in seconds and
# the full width of the window, in standard deviations. So std=.05 and size_factor=20 
# gives a total filter size of 0.05 sec * 20 = 1 sec.
```

The meaning of these args has changed

billbrod (commented):
Thanks for this Eric!

The `lazy_loading` is a pynapple version issue -- can you update your local pynapple and try again? The docs are failing to build because they actually need the `lazy_loading=False` kwarg there.

Other thoughts in comments. I also have some minor changes I was thinking about as I prepare for FENS, which I'll propose later today.

EricThomson:
Thanks a lot for all the helpful suggestions Billy, indeed I just needed to update Pynapple (sorry @gviejo :pineapple: ). I just finished the calcium imaging one, will get to this first thing tomorrow!

BalzaniEdoardo:
> Thanks a lot for all the helpful suggestions Billy, indeed I just needed to update Pynapple (sorry @gviejo 🍍 ). I just finished the calcium imaging one, will get to this first thing tomorrow!

Thanks for this, I updated the `pyproject.toml` in main to require `pynapple>=0.6.6`

billbrod:
For your comment about: 

> Note that this is not actually that informative and, in general, it is
> recommended that you focus on firing rates when interpreting your model. In
> particular, if your GLM includes auto-regressive inputs (e.g., neurons are
> connected to themselves or each other), simulate can behave poorly [^1].

What we mean is that simulating the spikes can behave poorly when the model includes auto-regressive inputs, because it can explode. Thus, you should interpret the firing rates instead of running them through the poisson process and comparing the spikes. I'm realizing that I basically just restated that paragraph -- was it any clearer? what in particular was unclear?

billbrod:
I also just pushed a commit with some other small changes (since they were in places not touched by your changes, I couldn't figure out how to propose them in the PR) -- check them over and see what you think! They are:

- Removed duplicate mention of Dayan and Abbott
- Replace `--` with `&mdash;`
- Moved mention of "hidden plotting function" to an admonition
- Removed out-of-date info about converting to jax arrays
- Added emojis to the discussion of model success/failure (inspired by Eric!)

EricThomson:
> For your comment about:
> 
> > Note that this is not actually that informative and, in general, it is
> > recommended that you focus on firing rates when interpreting your model. In
> > particular, if your GLM includes auto-regressive inputs (e.g., neurons are
> > connected to themselves or each other), simulate can behave poorly [^1].
> 
> What we mean is that simulating the spikes can behave poorly when the model includes auto-regressive inputs, because it can explode. Thus, you should interpret the firing rates instead of running them through the poisson process and comparing the spikes. I'm realizing that I basically just restated that paragraph -- was it any clearer? what in particular was unclear?

I was not understanding the final sentence of the original quote: I'm still a little unclear.  Is the point that spike estimates are so unreliable that feeding them back into the model (in the form of spike history filters and coupling filters) can be disadvantageous? I.e., even the firing rate estimates become unreliable when you start using the spiking outputs as inputs?  (Aside: if this is true can people use the firing rate instead of spikes in the coupling filters?)

Also, I'm not sure what the reference is for this (it goes to fn 1 but that is a bunch of stuff at the bottom of the page, I think).

billbrod:
My understanding is that the system is unstable: when the neuron is connected to itself, spiking leads to more spiking that can eventually get out of control. I think this is true if you use firing rates instead of spikes -- it's about the self-connectivity, rather than anything with spikes. @BalzaniEdoardo is that all correct? We should go through the paper, and maybe ask Memming, to get an accurate brief description of the problem.

For the citation, that's our bad, we're mixing two different ways of citing things. The footnote is supposed to lead to the footnote way at the bottom (in the red rectangle below), but the Allen Institute has a pretty extensive citation they want for people using their data, so that's the whole section above that. I agree we shouldn't mix them, but not sure how to get the best of both worlds: I like footnotes so it's clear exactly what citation goes with what, but I think the Allen Institute citation is probably too long for a footnote... what do you think?

EricThomson:
> My understanding is that the system is unstable: when the neuron is connected to itself, spiking leads to more spiking that can eventually get out of control. I think this is true if you use firing rates instead of spikes -- it's about the self-connectivity, rather than anything with spikes. @BalzaniEdoardo is that all correct? We should go through the paper, and maybe ask Memming, to get an accurate brief description of the problem.


Yikes ok this is pretty big then. I didn't realize this was a problem all the papers I've read have been like "yeah refractory period and connectivity!". On one hand I'm surprised at this because I would think it would learn negative weights or very small weights and with regularization things would be more stable. Plus the problem is convex so it is learning a good set of weights on the training data, so it is a problem of generalization?  But I also am _not_ surprised because...feedback systems. I have so many questions now, but I think for purposes of this demo, finding a clean and concise problem description would be great.

Cutting to the chase, there seem to be two issues. First, for interpreting your model, it is easier to focus on the rate than the Poisson outputs. This is pretty clear and obvious. Second, including feedback in your model may seem great, but it can produce problems. This seems a more meaty problem that probably a deserves a little bit more clarification (see next section on ref though -- my bad!).  (I'm not sure how much this extends to coupling between neurons versus being solely a self-connectivity issue). 

> For the citation, that's our bad, we're mixing two different ways of citing things. The footnote is supposed to lead to the footnote way at the bottom (in the red rectangle below), but the Allen Institute has a pretty extensive citation they want for people using their data, so that's the whole section above that. I agree we shouldn't mix them, but not sure how to get the best of both worlds: I like footnotes so it's clear exactly what citation goes with what, but I think the Allen Institute citation is probably too long for a footnote... what do you think?

Oops I totally missed that little thing at the end 😄  My initial gut reaction says to keep the Allen stuff as an acknowledgment section. Then, for paper references, either just use text links when they are discussed (I have a slight preference for this), or footnotes which will be separate from Acknowledgments (if there is already a convention to use footnotes instead of text links). Footnotes are tricky they can be easy to overlook and force you to jump around a page. Links tend to just work: they open a new page, and don't mess with your flow as much (IMHO).

ahwillia:
We should cite papers that tackle the runaway excitation problem, e.g:
https://ieeexplore.ieee.org/abstract/document/8008426?casa_token=-Aag9djSnEUAAAAA:X62c8_urj81Aiqa75J9D8HST7LTSV68Tkiz9fqIhBWJHxfdKW5JMUrw26dqrlcls4h5FJgS5KjSm

I think it is an issue with using maximum likelihood for parameter
estimation, among other things.

EricThomson:
> We should cite papers that tackle the runaway excitation problem, e.g: https://ieeexplore.ieee.org/abstract/document/8008426?casa_token=-Aag9djSnEUAAAAA:X62c8_urj81Aiqa75J9D8HST7LTSV68Tkiz9fqIhBWJHxfdKW5JMUrw26dqrlcls4h5FJgS5KjSm I think it is an issue with using maximum likelihood for parameter estimation, among other things.

This is helpful. Now that I go through stuff, e.g, [one of my favorite papers](https://pubmed.ncbi.nlm.nih.gov/28957020/), I see they do actually mention this and I never registered what they were talking about.

BalzaniEdoardo:
> > We should cite papers that tackle the runaway excitation problem, e.g: https://ieeexplore.ieee.org/abstract/document/8008426?casa_token=-Aag9djSnEUAAAAA:X62c8_urj81Aiqa75J9D8HST7LTSV68Tkiz9fqIhBWJHxfdKW5JMUrw26dqrlcls4h5FJgS5KjSm I think it is an issue with using maximum likelihood for parameter estimation, among other things.
> 
> This is helpful. Now that I go through stuff, e.g, [one of my favorite papers](https://pubmed.ncbi.nlm.nih.gov/28957020/), I see they do actually mention this and I never registered what they were talking about.

It is a problem of any auto regressive model fit with maximum likelihood or with weak regularization. The weights are good for predicting the next step, but there are no guarantees that they will generate a random process with similar statistics to the observed one. 
A simple  solution that make the glm stable is increase the regulation, if you shrink enough the parameters you avoid the excess self excitation. How much is enough regularization and how strong the bias will be is unclear a priori.

docs/neural_modeling/plot_01_current_injection.py
EricThomson: Yes good idea I'll add that.

docs/neural_modeling/plot_01_current_injection.py
BalzaniEdoardo: I agree with @billbrod, it is annoying that people have to download extra code to run the notebooks

BalzaniEdoardo:
I merged development here too

docs/neural_modeling/plot_01_current_injection.py
```diff
@@ -634,29 +614,34 @@
...
 #
-# Secondly, you may want a number with which to evaluate your model's
+# Finally, you may want a number with which to evaluate your model's
 # performance. As discussed earlier, the model optimizes log-likelihood to find
 # the best-fitting weights, and we can calculate this number using its `score`
 # method:
-# [^1]: Arribas, Diego, Yuan Zhao, and Il Memming Park. "Rescuing neural spike train models from bad MLE." Advances in Neural Information Processing Systems 33 (2020): 2293-2303.
+#
+# [^1]: Hocker, David, and Memming Park. "Multistep inference for generalized linear 
+# spiking models curbs runaway excitation." International IEEE/EMBS Conference on Neural Engineering,
```
BalzaniEdoardo, line 641:
```suggestion
#
# [^1]: Arribas, Diego, Yuan Zhao, and Il Memming Park. "Rescuing neural spike train models from bad MLE." Advances in Neural Information Processing Systems 33 (2020): 2293-2303.
# [^2]: Hocker, David, and Memming Park. "Multistep inference for generalized linear 
# spiking models curbs runaway excitation." International IEEE/EMBS Conference on Neural Engineering,
```

They are both good!

BalzaniEdoardo (changes requested):
Both papers talk about the same problem, which is still not fully solved.

docs/neural_modeling/plot_01_current_injection.py
```diff
@@ -634,29 +614,34 @@
...
-# recommended that you focus on firing rates when interpreting your model. In
-# particular, if your GLM includes auto-regressive inputs (e.g., neurons are
-# connected to themselves or each other), simulate can behave poorly [^1].
+# recommended that you focus on firing rates when interpreting your model. 
+#
+# Also, while 
+# including spike history is often helpful, it can sometimes make simulations unstable:
+# if your GLM includes auto-regressive inputs (e.g., neurons are
+# connected to themselves or each other), simulations can sometimes can behave 
+# poorly because of runaway excitation [^1].
```
BalzaniEdoardo, line 633:
```suggestion
# poorly because of runaway excitation [^1,2].
```

BalzaniEdoardo (commented):
added citation of second paper

EricThomson:
> added citation of second paper

Thanks for cleaning up my mess! 😊

I think I'm done futzing with this one it was pretty clean from the start. :ballot_box_with_check:

BalzaniEdoardo (approved):
All good!

BalzaniEdoardo:
First contribution from @EricThomson! 🚀🚀🚀

Issue #151, opened 2024-05-31
Notes on fit calcium imaging demo


EricThomson:
This is a great addition, and impressive that you added it so quickly!

I would suggest adding details about the fluorophore (e.g., gcamp6f?) and its kinetics, so we have a better sense of what we are recording here. When we load the calcium traces, I would probably add a few plots of the raw traces just to give people a sense of what the raw data look like (about half of users will have strong intuitions for ephys, but not optical phys, so I think just giving them a view of the slow time course etc will be useful).

Generally, this notebook marks a pretty major shift in the GLM theory from the LNP backbone we have been following until now. I guess it's now an LNG model :man_shrugging:  Previous notebooks take quite a bit of time to explain conceptual transitions additions. I think this deserves a bit of space. E.g. currently there is:

> Since the transients are non-negative, we will use a Gamma distribution from nemos with a soft-plus non linearity.

I think _way_ more is going to be needed to massage intuitions about what exactly is happening in the shift from Poisson to Gamma.  For example, something like:

- Everything up to this is still the same, but we are adding a new noise/observation model. 
- Need continuous distribution because we are no longer modeling a count process.
- Previously we interpreted the output of the linking function as mean firing rate now we interpret it as mean calcium signal and we use it to generate a noisy signal on top of that. We can't use Gaussian because Gaussian has negative values. 

*Something* like that -- that's just me spitballing not finished or good. :laughing: 

I think it would help to plot, like in the intro to GLM with LNP case, the output of the rate estimate, and then the output of the Gamma observation model, so people can visualize how the operations work with these continuous signals. 

We should probably add R^2 for the linear regression vs GLM, it's hard to compare them by just looking. 

In general, while this is promising, it would help to have a discussion of why the results are so far off sometimes (there are clear false positive and false negative calcium signals). I'm not sure what needs to be added to it to improve accuracy? Gamma params need some further constraints? Coupling params needed? Maybe it's better for high SNR units (it's not too hard to calculate SNR from Ca  traces -- I wonder if it would be better for 2p data which has better SNR). 

Anyway those are my initial thoughts this seems very cool and an impressive addition to nemos! It is concerning that the results are so similar to linear regression. A lot to chew on here.

Sorry I'm getting tired but I can tweak it with a PR in some of the ways above, and put in little notes where I would need help or am unsure of things.

BalzaniEdoardo:
[...]

PR #159, opened 2024-06-04
update to calcium imaging demo

EricThomson:
This pertains to Issue #151, mostly minor update to the calcium imaging demo. I did my best to address many of the things I mentioned in that issue, but not all. For instance:

1. I said it was `GCaMP6f` I'm not sure if this is right can we find out? :smile: 
2. The Pynapple object has lots of data we don't discuss it might be helpful to very briefly mention. E.g., are `x,y,z` animal location? are `rx` and `rz` head direction like `ry`, but on a different plane?
3. Is there an advantage of the soft-plus nonlinearity when using calcium imaging data?
4. In general at the end it sort of just sputters out -- we show how to fit the GLM to the data, and compare to linear regression, but don't really discuss or compare and contrast. Also, what might be added to improve the fit, reduce the bias (the bias looks fairly high)? Tweaking the regularization? Am I right in assuming we only have activity history filter, but not inter-neuron filter? As it stands the results are pretty far off, so I think this needs to be addressed. I would probably low-pass filter the signals with a Gaussian (on top of the subsampling already done) and then re-run it to see how it does, it just seems to not be handling the high frequency stuff very well (but this is conjecture on my part).

### DIFF
--- a/docs/neural_modeling/plot_06_calcium_imaging.py
+++ b/docs/neural_modeling/plot_06_calcium_imaging.py
@@ -4,7 +4,7 @@ Fit Calcium Imaging
 ============
 
 
-For the example dataset, we will be working with a recording of a freely-moving mouse imaged with a Miniscope (1-photon imaging using the genetically encoded calcium indicator GCaMP6f). The area recorded for this experiment is the postsubiculum - a region that is known to contain head-direction cells, or cells that fire when the animal's head is pointing in a specific direction.
+For the example dataset, we will be working with a recording of a freely-moving mouse imaged with a Miniscope (1-photon imaging at 30Hz using the genetically encoded calcium indicator GCaMP6f). The area recorded for this experiment is the postsubiculum - a region that is known to contain head-direction cells, or cells that fire when the animal's head is pointing in a specific direction.
 
 The data were collected by Sofia Skromne Carrasco from the Peyrache Lab.
 
@@ -119,8 +119,9 @@ plt.show()
 # %%
 # ## Basis instantiation
 # 
-# We can define a cyclic-BSpline for capturing the encoding of the heading angle, and a
-# log-spaced raised cosine basis for the coupling filters. We can combine the two basis.
+# We can define a cyclic-BSpline for capturing the encoding of the heading angle, and a log-spaced raised cosine basis for the coupling filters between neurons. Note that we are not including a self-coupling (spike history) filter, because in practice we have found it results in overfitting.
+#
+# We can combine the two bases.
 
 heading_basis = nmo.basis.CyclicBSplineBasis(n_basis_funcs=12)
 coupling_basis = nmo.basis.RaisedCosineBasisLog(3, mode="conv", window_size=10)
@@ -135,18 +136,21 @@ basis = heading_basis + coupling_basis
 # %%
 # ## Gamma GLM
 #
-# Until now, we have been modeling spike trains, and have used a Poisson distribution for the observability model. With calcium traces, things are quite different: we no longer have counts but continuous signals, so the Poisson is no longer appropriate. We cannot use a Gaussian because the calcium traces are non-negative. To satisfy these constraints, we will use a Gamma distribution from `nemos` with a soft-plus non linearity.
+# Until now, we have been modeling spike trains, and have used a Poisson distribution for the observability model. With calcium traces, things are quite different: we no longer have counts but continuous signals, so the Poisson is no longer appropriate. We cannot use a Gaussian because the calcium traces are non-negative. To satisfy these constraints, we will use the built-in Gamma distribution provided with `nemos`. For the linking function, instead of the default exponential function, we will use a soft-plus non linearity.
 
 
 model = nmo.glm.GLM(
-    regularizer=nmo.regularizer.UnRegularized(solver_name="LBFGS", solver_kwargs=dict(tol=10**-13)),
+    regularizer=nmo.regularizer.Ridge(solver_name="LBFGS", 
+                                      regularizer_strength=0.02, 
+                                      solver_kwargs=dict(tol=10**-13)),
     observation_model=nmo.observation_models.GammaObservations(inverse_link_function=jax.nn.softplus)
 )
+
 
 # %%
 # We select one neuron to fit later, so remove it from the list of predictors
 
-neu = 4
+neu = 12
 selected_neurons = jnp.hstack(
     (jnp.arange(0, neu), jnp.arange(neu+1, Y.shape[1]))
 )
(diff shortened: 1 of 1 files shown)


BalzaniEdoardo:
> This pertains to Issue #151, mostly minor update to the calcium imaging demo. I did my best to address many of the things I mentioned in that issue, but not all. For instance:
> 
>     1. I said it was `GCaMP6f` I'm not sure if this is right can we find out? 😄
> 
>     2. The Pynapple object has lots of data we don't discuss it might be helpful to very briefly mention. E.g., are `x,y,z` animal location? are `rx` and `rz` head direction like `ry`, but on a different plane?
> 
>     3. Is there an advantage of the soft-plus nonlinearity when using calcium imaging data?
> 
>     4. In general at the end it sort of just sputters out -- we show how to fit the GLM to the data, and compare to linear regression, but don't really discuss or compare and contrast. Also, what might be added to improve the fit, reduce the bias (the bias looks fairly high)? Tweaking the regularization? Am I right in assuming we only have activity history filter, but not inter-neuron filter? As it stands the results are pretty far off, so I think this needs to be addressed. I would probably low-pass filter the signals with a Gaussian (on top of the subsampling already done) and then re-run it to see how it does, it just seems to not be handling the high frequency stuff very well (but this is conjecture on my part).

Hi Eric, for the bias stand point, we can cross-validate the regularizer. 

```python
from sklearn.model_selection import GridSearchCV
# assume you have the glm model
grid = dict(regularizer__regularizer_strength=[0.001, 0.01, 0.1])
clf = GridSearchCV(model, grid)
clf.fit(X, Y)
```
This should work, modulo "regularizer__regularizer_strength" being the right parameter name. You can check the key to use,
```
print(model.get_params())
``` 

For what it concern what we are fitting, we are fitting only the coupling filters from other neuron, not the autocorr filter. If you add an autocorr filter, there is a long autocorr, that will result in overfitting

docs/neural_modeling/plot_06_calcium_imaging.py
```diff
@@ -103,26 +126,26 @@
...
 basis = heading_basis + coupling_basis
 
+
 # %%
 # ## Gamma GLM
 #
-# Since the transients are non-negative, we will use a Gamma distribution from `nemos` with a soft-plus non linearity.
-#
+# Until now, we have been modeling spike trains, and have used a Poisson distribution for the observability model. With calcium traces, things are quite different: we no longer have counts but continuous signals, so the Poisson is no longer appropriate. We cannot use a Gaussian because the calcium traces are non-negative. To satisfy these constraints, we will use a Gamma distribution from `nemos` with a soft-plus non linearity.
+
```
BalzaniEdoardo, line 139: Following up on your comment, we can add a note about the non-linearity. something of this kind
```suggestion
!!! note "Non-linearity"
    Different option are possible. With a soft-plus we are assuming an "additive" effect the predictors, while an exponential non-linearity assumes multiplicative effects. Deciding which firing rate model works best is an empirical question. You can fit different configurations to see which one capture best the neural activity.
```

docs/neural_modeling/plot_06_calcium_imaging.py
```diff
@@ -103,26 +126,26 @@
...
 
 model = nmo.glm.GLM(
     regularizer=nmo.regularizer.UnRegularized(solver_name="LBFGS", solver_kwargs=dict(tol=10**-13)),
     observation_model=nmo.observation_models.GammaObservations(inverse_link_function=jax.nn.softplus)
 )
 
 # %%
-# We select a neuron to fit and remove it from the list of predictors
+# We select one neuron to fit later, so remove it from the list of predictors
```
BalzaniEdoardo, line 147: In this case, we are removing the self-coupling from the predictors. If this wasn't clear to you. @EricThomson we should rephrase it. The neuron was left out because of the long auto-corr time; the autocorr filter resulted in over fitting.

docs/neural_modeling/plot_06_calcium_imaging.py
```diff
@@ -215,6 +239,8 @@
 plt.show()
 
 # %%
+# While there is some variability in the fit for both models, one advantage of the gamma distribution is clear: the nonnegativity constraint is followed with the data. 
+#
```
BalzaniEdoardo, line 243: Maybe we can add, 
```suggestion
# This is needed if you want to use GLMs for predict the firing rate, which needs to be positive, in response a simulated input, as shown in [this paper](https://www.nature.com/articles/s41467-017-01908-3) for a Poisson GLM.

**Review by BalzaniEdoardo — APPROVED** [review]
Thank you Eric, this is a very nice editing!

I added a couple of comments to this but looks good to me!

File: docs/neural_modeling/plot_06_calcium_imaging.py (hunk 1 above)
Comment (EricThomson) [reply docs/neural_modeling/plot_06_calcium_imaging.py:139]: These are all helpful feedback, I will incorporate asap!

**Comment by gviejo** [comment]
Yes it's gcamp6f

**Comment by EricThomson** [comment]
Thanks a lot Edoardo for this  reply. I'm adjusting things accordingly. I tried running on some Gaussian smoothed data -- it did not go well, things looked ... not better :smile: I'm adding a regularizer so the tuning curves seem to better match the actual, and also this might give students a chance to tweak it (I did a manual grid search :upside_down_face: ).

**Comment by EricThomson** [comment]
I added a little thing at the end about how this is a more experimental tool than the stuff for spiking models. Feel free to alter/cut I often err on the side of verbose or overly flowery language. :rose: 

I could do a lot more with this, but I'm about out of time I have to get ready for a trip soon so have at it; what you did looks really good, I just made some minor adjustments! :fire:

File: docs/neural_modeling/plot_06_calcium_imaging.py [hunk 4]
```diff
@@ -91,38 +114,43 @@
...
+heading_basis.identifiability_constraints = True
 basis = heading_basis + coupling_basis
 
+
 # %%
 # ## Gamma GLM
 #
-# Since the transients are non-negative, we will use a Gamma distribution from `nemos` with a soft-plus non linearity.
-#
+# Until now, we have been modeling spike trains, and have used a Poisson distribution for the observability model. With calcium traces, things are quite different: we no longer have counts but continuous signals, so the Poisson is no longer appropriate. We cannot use a Gaussian because the calcium traces are non-negative. To satisfy these constraints, we will use a Gamma distribution from `nemos` with a soft-plus non linearity.
```
BalzaniEdoardo:
```suggestion
# Until now, we have been modeling spike trains, and have used a Poisson distribution for the observability model. With calcium traces, things are quite different: we no longer have counts but continuous signals, so the Poisson is no longer appropriate. A a Gaussian model is not ideal since the calcium traces are non-negative. To satisfy these constraints, we will use a Gamma distribution from NeMoS with a soft-plus non linearity.
```

BalzaniEdoardo (approved):
I'll add a couple of commits, but this is great. the fit looks better too

Issue #147, opened 2024-05-24
Notes on GLM an Introduction doc


EricThomson:
Hi all I'm just going through all the demo stuff and taking notes to pester you. 😄 Starting with the Intro to GLM.

**GLM an intro notes** 
https://nemos.readthedocs.io/en/latest/generated/background/plot_00_conceptual_intro/

Many people will be going to this page as their *very first exposure* to GLMs, including lots of experimentalists. So I'm reading it with that mindset.  Overall, _this is great_. It's intuitive what is happening, and I really like how you explain the three operations. It's very well done. The explanation at the end of how you compare your model to the data is really nice too!

I have a few minor comments/questions.

1.  LNP vs GLM
    One worry I have is that from reading this, beginners might come away thinking that the GLM is just the LNP model. I might add something discussing what you'd need to add to the picture here to make a true GLM. Is it the spike history filter? If so, then maybe something like Figure 11.5B from Neuronal Dynamics: https://neuronaldynamics.epfl.ch/online/Ch11.S2.html

![lnp_vs_glm](https://github.com/flatironinstitute/nemos/assets/6643322/969d54ac-2464-436b-9268-12ca4c05126e)

2. "to make this plot work well, keep this to three values"

    Does three 'values' mean three different neurons? Three trials? Also, what are the spike count targets? What is this supposed to be? Is there an experiment, or is this simulated data to make a conceptual point. The sausage is made in the `lnp_schematic` function, so I'd explain a bit more. Experimentalists reading this will appreciate details about what this example is supposed to be showing. In particular, is there a target spike count here -- at the end you nicely discuss measuring goodness of fit using log likelihood, but we don't see this in action.

3. The aside on affine transformation   

    I know it will drive you crazy to cut it because it's accurate and the math details are important. But beginners won't care, and computationalists will cut you some slack because they know that people talk like this. I'd consider calling it a "technical aside" or something so people know they can skip it. Or...just cut it. 😅

ahwillia:
Thanks Eric, this is very helpful. A quick comment on the "LNP vs GLM" distinction... A GLM is a broad class of models, and an LNP is one example of a GLM. The Neural Dynamics chapter / link you sent is unfortunately quite confusing on this point.

Wikipedia summarizes a GLM as any model with the following three components. By this measure *both* the LNP and GLM models in the diagram are GLMs.

[...]

PR #163, opened 2024-06-06
Intro demo update

EricThomson:
Addresses issue #147 .

This is mostly minor clarification stuff that I thought would help make it easier for a newcomer to understand.  I didn't mess with the figure situation after talking to Edoardo, for now I think this is fine -- instead I just discussed the more general GLM stuff at the bottom in a new section. I can cut the reference to https://www.ncbi.nlm.nih.gov/pmc/articles/PMC6226578/ if that's too much, but I thought it was pretty cool. :smile:


docs/background/plot_00_conceptual_intro.py
```diff
@@ -188,3 +194,7 @@
 #     `score` method, passing the predictors and the counts. The method first
 #     computes the rate $\lambda(t)$ using (2) and then the likelihood using
 #     (4). This method is used under the hood during optimization.
+
+# %%
+### More general GLMs
+# So far, we have focused on the relatively simple LNP model of spike generation, which is a special case of a GLM. The LNP model has some known shortcomings (Pillow et al 2008). For instance, LNP ignores things like refactory periods and other history-dependent features of spiking in a neuron. As we will show in other demos, such _spike history filters_ can be built into GLMs to give more accurate results. We will also show how, if you have recordings from a large _population_ of neurons simultaneously, you can build connections between the neurons into the GLM in the form of _coupling filters_. This can help answer the degree to which activity is driven primarily by the input X, or by network influences in the population. Some researcheers have used cascades of GLMs to model dendritic computation in individual neurons (Ujfalussy et al 2018).
```
billbrod, line 200: I like this previewing what else they can find in our docs. For the papers, can you add citations using footnotes? (you can see an example in the current injection example).

billbrod (commented):
Thanks for the suggestions Eric! Following up on the issue, I see you mentioned using the images we have to build up a schematic of the model. Are you still thinking about doing that? Just want to check!

Also, I'm not sure why the docs build is suddenly failing, since none of your changes should have affected the code -- @BalzaniEdoardo any ideas?

EricThomson:
> Thanks for the suggestions Eric! Following up on the issue, I see you mentioned using the images we have to build up a schematic of the model. Are you still thinking about doing that? Just want to check!

I decided to just leave the figures as is for now -- I think the goal is ultimately to replace that image with something more like the fancy images in the repo readme. I looked over those and I wasn't confident in my ability to quickly integrate those into this demod, so I'm leaving that for a future editor. Instead I just added the text at the end to discuss all the possibilities available with GLMs.

docs/background/plot_00_conceptual_intro.py
EricThomson: Yes I can do that. Incidentally I'm planning to adapt to all these super helpful review comments at _some_ point today/tonight. I'm getting ready for a big family trip and doing a bunch of other stuff simultaneously, but I will adjust PRs soon! 😅

BalzaniEdoardo:
> Thanks for the suggestions Eric! Following up on the issue, I see you mentioned using the images we have to build up a schematic of the model. Are you still thinking about doing that? Just want to check!
> 
> Also, I'm not sure why the docs build is suddenly failing, since none of your changes should have affected the code -- @BalzaniEdoardo any ideas? It looks like it can't pick an initial value and so wants us to do that, but I'm not sure why that's just becoming a problem now

I fixed the error. It was caused by a change I made in the parameter initialization of the GLM. Now I am relying on scipy root finding to numerically invert the link function to set the initial intercept. 

This method finds good solutions (as expected since our link functions are smooth, univariate, monotonic functions) but I I did not realize that sometimes the root finding could result in an unsuccessful optimization even if the error is of the order of 10**-10. 

Instead of checking the root finding "success" flag, now I am simply checking that the solution found is close enough to zero (`np.allclose`) and this seems to work well.

docs/background/plot_00_conceptual_intro.py
BalzaniEdoardo: Your edits really improved the flow and info content of the notebook, I really like it.

BalzaniEdoardo (commented):
Once the citation are added to the footnotes, this is good to go for me!

docs/background/plot_00_conceptual_intro.py
BalzaniEdoardo, line 200:
```suggestion
# So far, we have focused on the relatively simple LNP model of spike generation, which is a special case of a GLM. The LNP model has some known shortcomings (Pillow et al 2008). For instance, LNP ignores things like refactory periods and other history-dependent features of spiking in a neuron. As we will show in other demos, such _spike history filters_ can be built into GLMs to give more accurate results. We will also show how, if you have recordings from a large _population_ of neurons simultaneously, you can build connections between the neurons into the GLM in the form of _coupling filters_. This can help answer the degree to which activity is driven primarily by the input X, or by network influences in the population.
```

I checked the ref, the model is more similar to a standard RNN than a GLM. Citing: "The error is a non-convex function of the parameters with multiple local minima. ". I think this would be misleading for an inexperienced audiece

BalzaniEdoardo (changes requested):
I checked the ref, and, while the paper cited is very interesting, but the model is a big departure from GLMs, I would remove it not to confuse the audience!

docs/background/plot_00_conceptual_intro.py
BalzaniEdoardo: It's basically an NN with a hierarchical structure to its connection

EricThomson:
> Once the citation are added to the footnotes, this is good to go for me!

I am running off to some meetings, but should be able to finish this and the calcium changes later today. This one is just about ready (it was already very good, so changes were all pretty minor).

billbrod:
> > Thanks for the suggestions Eric! Following up on the issue, I see you mentioned using the images we have to build up a schematic of the model. Are you still thinking about doing that? Just want to check!
> 
> I decided to just leave the figures as is for now -- I think the goal is ultimately to replace that image with something more like the fancy images in the repo readme. I looked over those and I wasn't confident in my ability to quickly integrate those into this demod, so I'm leaving that for a future editor. Instead I just added the text at the end to discuss all the possibilities available with GLMs.

Sounds good! Since this PR will close the issue where we were discussing that, can you create a new one summarizing the plan? then someone else can pick it up later.

EricThomson:
Open to more feedback, but I _think_ this is good to go. I'll add issue discussing future plans with the diagrams.

billbrod (approved):

docs/background/plot_00_conceptual_intro.py
```diff
@@ -188,3 +194,9 @@
 #     `score` method, passing the predictors and the counts. The method first
 #     computes the rate $\lambda(t)$ using (2) and then the likelihood using
 #     (4). This method is used under the hood during optimization.
+
+# %%
+### More general GLMs
+# So far, we have focused on the relatively simple LNP model of spike generation, which is a special case of a GLM. The LNP model has some known shortcomings [^1]. For instance, LNP ignores things like refactory periods and other history-dependent features of spiking in a neuron. As we will show in other demos, such _spike history filters_ can be built into GLMs to give more accurate results. We will also show how, if you have recordings from a large _population_ of neurons simultaneously, you can build connections between the neurons into the GLM in the form of _coupling filters_. This can help answer the degree to which activity is driven primarily by the input X, or by network influences in the population.
```
BalzaniEdoardo, line 200:
```suggestion
# So far, we have focused on the relatively simple LNP model of spike generation, which is a special case of a GLM. The LNP model has some known shortcomings[^1]. For instance, LNP ignores things like refactory periods and other history-dependent features of spiking in a neuron. As we will show in other demos, such _spike history filters_ can be built into GLMs to give more accurate results. We will also show how, if you have recordings from a large _population_ of neurons simultaneously, you can build connections between the neurons into the GLM in the form of _coupling filters_. This can help answer the degree to which activity is driven primarily by the input X, or by network influences in the population.
```

BalzaniEdoardo (approved):
just fixed spacing of foot note

PR #164, opened 2024-06-07
Update basis docstrings

billbrod:
Small PR to update basis docstrings so that they explain what mode does. also adds docstring for `RaisedCosineBasisLog`'s `time_scaling` argument.

As part of this, we now raise a ValueError if someone tries to set args/kwargs for a basis in eval mode.

### DIFF
--- a/docs/api_guide/plot_03_glm_pytree.py
+++ b/docs/api_guide/plot_03_glm_pytree.py
@@ -209,7 +209,7 @@ axes[-1,-1].remove()
 unit_no = 7
 spikes = nwb['units'][unit_no]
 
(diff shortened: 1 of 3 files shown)


docs/api_guide/plot_03_glm_pytree.py
```diff
@@ -209,7 +209,7 @@
 unit_no = 7
 spikes = nwb['units'][unit_no]
 
-basis = nmo.basis.CyclicBSplineBasis(10, 5)
+basis = nmo.basis.CyclicBSplineBasis(10, order=5)
```
BalzaniEdoardo, line 212: good catch

BalzaniEdoardo (approved):
looks good, thanks for this!

Issue #158, opened 2024-06-04
Do something else with `examples_utils` code?


billbrod:
Currently, we have a folder of `examples_utils`, which contains some code leftover from the workshop(s) we ran, including downloading data and plotting. Should think more carefully about what to do with them (downloading data, in particular, should probably be changed to using pooch, see [this plenoptic PR](https://github.com/LabForComputationalVision/plenoptic/pull/252) and the [FENS workshop](https://github.com/flatironinstitute/ccn-software-fens-2024)). And the plotting code could potentially be made more general.

BalzaniEdoardo:
`importlib` is pretty cool, we should use it! For what it concerns the plotting utils, is there a way to include the script when downloading the notebook/script generated by `mkdocs-gallery`? It is a bit annoying that people cannot run the notebook without downloading extra utils.

billbrod:
Agreed that it's annoying and ideally people won't have to download anything else. I think the mplstyle file and data downloading should be moved into the main code. For the plotting code, I think we'll need to think: for some of them, it may be worth generalizing them a little and including in the main code. For others that are too example-specific, ideally they'd be in the notebook but hidden, but I can't find a way to do that in `mkdocs-gallery` (or `sphinx-gallery`, for that matter). It's probably worth poking around other packages, to see what they do, and asking around.

billbrod:
Edoardo figured out how to get collapsible blocks using the `???` syntax, but those cells don't get run, so it doesn't solve this problem. what we want is something like the `hide-input` tag from [myst-nb](https://myst-nb.readthedocs.io/en/latest/render/hiding.html)

BalzaniEdoardo:
close with PR #216

PR #169, opened 2024-06-17
Update `TransformerBasis` and `Basis` to work with scikit-learn pipelining and CV

### LINKED ISSUE #156: Improve TransformerBasis API
**Is your feature request related to a problem? Please describe.**
Basis can be converted into a "sklearn" transformer for pipelining and cross-validation. 
The main difference between our basis and a transformer is that inputs are only X and optionally y. Instead, our basis api assumes any number of 1d inputs depending on the basis type. Currently, "Basis" and "TransformerBasis" behaves differently, specifically, you cannot get the attributes of "Basis" from its transformer version.

**Describe the solution you'd like**
It would be nice to implement a custom "__getitem__" that returns `getattr(self._basis, attrname)`, so that the "transformer basis" behaves as much as possible much as the regular one. This change will enable

```python
from nemos import basis

bas = basis.RaisedCosineBasisLinear(5)
bas = basis.TransformerBasis(bas)
print(bas.n_basis_funcs)
```

This will likely play nice with "get_params" and "set_params" form "Base", helpful in order to cross-validate directly parameters like "n_basis_funcs".

**Describe alternatives you've considered**
I do not see an alternative solution in which we will make our basis a transformer directly. This would require the user to stack all the inputs in a 2D array, which doesn't play nicely with how the time-series are stored in neuroscience generally.

**Additional features**
Implement a "to_transformer" mehtod in "nemos.basis.Basis" that returns `TransformerBasis(self)`

bagibence:
The goal of this PR is to allow `TransformerBasis` to act as a transformer step in a scikit-learn pipeline and to be able to cross-validate its parameters (mainly the number of basis functions).

It makes the following changes:
* Make `Basis` inherit from `Base` and update the `__init__` signature to not have variable number of positional arguments, making it compatible with `Base.get_params` and `Base.set_params` (relying on `Base._get_param_names`) which are used during cross-validation.
* Overwrite `TransformerBasis.__getattr__` for easy access to the attributes of the underlying `Basis` object.
* Overwrite `set_params` and `__sklearn_clone__` in `TransformerBasis` to make it work with CV.
* Add some tests for the points above and remove the one testing if the positional convolution arguments are passed (as we don't have these anymore).
* Add `Basis.to_transformer()` which returns a `TransformerBasis`.
* Add a short tutorial to the docs, showing how to use `TransformerBasis` and `GLM` together in a scikit-learn `Pipeline` and jointly cross-validate parameters of the pipeline's steps on some toy data.

Fixes #156


bagibence:
Here I will document a few decisions we made to make `TransformerBasis` compatible with pickling and scikit-learn's model selection.

### Cloning
When running cross-validation, scikit-learn clones the estimator objects, and the way cloning is done then by default gives an error in our case:

```
File ~/.../sklearn/base.py:123, in _clone_parametrized(estimator, safe)
    121 new_object_params = estimator.get_params(deep=False)
    122 for name, param in new_object_params.items():
--> 123     new_object_params[name] = clone(param, safe=False)
    125 new_object = klass(**new_object_params)
    126 try:
...
--> 125 new_object = klass(**new_object_params)
    126 try:
    127     new_object._metadata_request = copy.deepcopy(estimator._metadata_request)

TypeError: TransformerBasis.__init__() got an unexpected keyword argument '_basis'
```

That is because `TransformerBasis.get_params` returns the dictionary `{"_basis": self._basis, **self._basis.get_params()}` (or `self._basis.get_params()` if not explicitly overwritten).
In the error above `klass` is `TransformerBasis`, and `TransformerBasis.__init__` takes a single argument called `basis`, so we get an error when trying to pass it the dictionary's items as keyword arguments.
A potential solution could be to redefine `get_params`, but then we still get an error `RuntimeError: Cannot clone object <nemos.basis.TransformerBasis object at 0x3089f3e00>, as the constructor either does not set or modifies parameter basis`.
This is because of the copying done in `TransformerBasis.__init__`, but we want that copying because without it the following would produce unexpected behavior:
```
bas = nmo.basis.RaisedCosineBasisLinear(5)
trans_bas = nmo.basis.TransformerBasis(bas)
# setting this is allowed by __setattr__
trans_bas.n_basis_funcs = 10
print(bas.n_basis_funcs)
```
Without copying we would modify the underlying `Basis` and this would give `10` instead of the expected `5`.

The way out is to define `TransformerBasis.__sklearn_clone__` in which we return a new `TransformerBasis` with a copy of the underlying `Basis. For more info see https://scikit-learn.org/stable/developers/develop.html#cloning

### Controlling grid-search parameters with `set_params`
When using `TransformerBasis` as a step in a `Pipeline` and model selection, we want to allow setting the parameters in the parameter grid like this:
```
param_grid = dict(
    transformerbasis__n_basis_funcs=(3, 5, 10, 20, 100)
)
```
or alternatively setting the basis directly like this:
```
param_drid = dict(
    transformerbasis___basis=(
        nmo.basis.MSplineBasis(5),
        nmo.basis.MSplineBasis(10),
        nmo.basis.RaisedCosineBasisLinear(10),
    )
)
```
but we don't want to allow mixing the two like this:
```
param_grid = dict(
    transformerbasis__n_basis_funcs=(3, 5, 10, 20, 100),
    transformerbasis___basis=(
        nmo.basis.RaisedCosineBasisLinear(10),
        nmo.basis.RaisedCosineBasisLog(10),
        nmo.basis.MSplineBasis(10),
    )
)
```
This can be achieved and enforced by defining `set_params`.
Also see https://scikit-learn.org/stable/developers/develop.html#get-params-and-set-params

### Infinite recursion when unpickling
The next problem we can run into with `sklearn.model_selection` has to do with pickling.
When scikit-learn does cross-validation with `n_jobs` set to something larger than 1, it uses joblib to distribute things across processes. To do this, it clones, pickles, and unpickles the estimator object passed to it.
We ran into the errors:
`BrokenProcessPool: A task has failed to un-serialize. Please ensure that the arguments of the function are all picklable.` and `RecursionError: maximum recursion depth exceeded` in `__getattr__`.

The [pickle docs mention](https://docs.python.org/3/library/pickle.html#object.__setstate__) that `__getattr__` may be called when unpickling, and the solution we settled on is to explicitly define `__getstate__` and `__setstate__`, [also mentioned in the pickle docs](https://docs.python.org/3/library/pickle.html#handling-stateful-objects).
Our implementation explicitly writes that the state only contains `self._basis`, which is the only allowed direct attribute (which is enforced by our custom `__setattr__`):
```
def __getstate__(self):
    return {"_basis": self._basis}

def __setstate__(self, state):
    self._basis = state["_basis"]
```

An alternative, less explicit definition of `__getstate__` and `__setstate__` could be:
```
def __getstate__(self):
    return self.__dict__.copy()

def __setstate__(self, state):
    self.__dict__.update(state)
```

Another alternative, but less intuitive/explicit, solution would be to modify `__getattr__` to avoid the infinite recursion:
```
def __getattr__(self, name: str):
    if name in ("__getstate__", "__setstate__", "__reduce__", "__reduce_ex__"):
        raise AttributeError(name)
    return getattr(self._basis, name)

def __getattr__(self, name: str):
    if name == "_basis":
        raise AttributeError(name)
    return getattr(self._basis, name)
```

bagibence:
### Metadata routing and inheriting from `BaseEstimator`

While `GLM.score` supports different ways of evaluating goodness-of-fit through the `score_type` argument, `pipeline.score(X, y, score_type="...")` does not propagate this. It also doesn't warn about it or throw an error, just silently uses the default value of `log-likelihood`

A potential solution to this would be to add scikit-learn as a dependency, inherit from `BaseEstimator`, and use the [metadata routing feature](https://scikit-learn.org/stable/metadata_routing.html), but we decided against this while the API is experimental.

In the meantime, in the docs we suggest [creating a custom scorer](https://scikit-learn.org/stable/modules/generated/sklearn.metrics.make_scorer.html). This scorer can be used to evaluate a pipeline by calling `my_custom_scorer(pipeline, X, y)`.

BalzaniEdoardo (commented):
@bence I added some comments, for myself mostly, I am finishing this up. Thanks for it! It was a very nice job

src/nemos/basis.py
```diff
@@ -692,6 +714,12 @@ def __pow__(self, exponent: int) -> MultiplicativeBasis:
             result = result * self
         return result
 
+    def to_transformer(self) -> TransformerBasis:
+        """
+        Turn the Basis into a TransformerBasis for use with scikit-learn.
```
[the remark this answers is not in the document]
bagibence: I added one. Let me know how you like it.

docs/api_guide/plot_06_sklearn_pipeline_cv_demo.py
```diff
@@ -0,0 +1,473 @@
...
+
+# %%
+# ## What is a scikit-learn pipeline
+#
+# <figure markdown>
+# <img src="../../../assets/pipeline.svg" style="width: 100%"/>
+# <figcaption>Schematic of a scikit-learn pipeline.</figcaption>
+# </figure>
+#
+# A pipeline is a sequence of data transformations. Each step in the pipeline transforms the input data into a different representation. The final step is either another transformation or a model step that fits, predicts, or scores based on the previous step's output and some observations. Setting up such machinery can be simplified using the `Pipeline` class from scikit-learn.
```
billbrod, line 23:
```suggestion
# A pipeline is a sequence of data transformations leading up to a model. Each step before the final one transforms the input data into a different representation, and then the final model step fits, predicts, or scores based on the previous step's output and some observations. Setting up such machinery can be simplified using the `Pipeline` class from scikit-learn.
```

While this is more limited than the pipelines actually are, I think it's clearer (and true for our use cases)

docs/api_guide/plot_06_sklearn_pipeline_cv_demo.py
```diff
@@ -0,0 +1,473 @@
...
+#
+# To set up a scikit-learn `Pipeline`, ensure that:
+#
+# 1. Each intermediate step is a [scikit-learn transformer object](https://scikit-learn.org/stable/data_transforms.html) with a `transform` and/or `fit_transform` method.
+# 2. The final step is either another transformer or an [estimator object](https://scikit-learn.org/stable/developers/develop.html#estimators) with a `fit` method, or a model with `fit`, `predict`, and `score` methods.
```
billbrod, line 28:
```suggestion
# 2. The final step is an [estimator object](https://scikit-learn.org/stable/developers/develop.html#estimators) with a `fit` method, or a model with `fit`, `predict`, and `score` methods.
```

Is there a use case in which we expect users to have a pipeline ending in a transformer? else, we should not describe it for clarity (maybe add an admonition that says it's possible and links to something in sklearn's docs)

docs/api_guide/plot_06_sklearn_pipeline_cv_demo.py
```diff
@@ -0,0 +1,473 @@
...
+# from sklearn.pipeline import Pipeline
+#
+# # Assume transformer_i/predictor is a transformer/model object
+# pipe = Pipeline(
+#     [
+#         ("label_1", transformer_1),
+#         ("label_2", transformer_2),
+#         ...,
+#         ("label_n", transformer_n),
+#         ("label_model", model)
```
billbrod, line 43: should say that these labels are arbitrary and where/how they're used (only for display, right?)

docs/api_guide/plot_06_sklearn_pipeline_cv_demo.py
```diff
@@ -0,0 +1,473 @@
...
+# Xn = transformer_n.fit_transform(Xn_1)
+#
+# # Fit step
+# model.fit(Xn, y)
+# ```
+# And the same holds for `pipe.score` and `pipe.predict`.
+#
+# ## Why pipelines are useful
+#
+# Pipelines not only streamline and simplify your code but also offer several other advantages. The real power of pipelines becomes evident when combined with the scikit-learn `model_selection` module. This combination allows you to tune hyperparameters at each step of the pipeline in a straightforward manner.
```
billbrod, line 63:
```suggestion
# Pipelines not only streamline and simplify your code but also offer several other advantages. The real power of pipelines becomes evident when combined with the scikit-learn `model_selection` module, which includes cross-validation and similar methods. This combination allows you to tune hyperparameters at each step of the pipeline in a straightforward manner.
```

I think we should explicitly say "cross-validation" here

docs/api_guide/plot_06_sklearn_pipeline_cv_demo.py
```diff
@@ -0,0 +1,473 @@
...
+)
+y = np.random.poisson(rate).astype(float).flatten()
+
+# %%
+fig, ax = plt.subplots()
+ax.scatter(X.flatten(), y, alpha=0.2)
+ax.set_xlabel("input")
+ax.set_ylabel("spike count")
+sns.despine(ax=ax)
+
```
billbrod, line 95: should describe this data: this is showing a simulated neuron's tuning curve, which here is bimodal, normally distributed around the peaks of 0.25 and 0.75.

docs/api_guide/plot_06_sklearn_pipeline_cv_demo.py
```diff
@@ -0,0 +1,473 @@
...
+# %%
+# ### Converting NeMoS `Basis` to a transformer
+# In order to use NeMoS `Basis` in a pipeline, we need to convert it into a scikit-learn transformer. This can be achieved through the `TransformerBasis` wrapper class.
+#
+# Instantiating a `TransformerBasis` can be done using the constructor directly or with `Basis.to_transformer()`:
```
billbrod, line 100:
```suggestion
# Instantiating a `TransformerBasis` can be done either using the constructor directly or with `Basis.to_transformer()`:
```

docs/api_guide/plot_06_sklearn_pipeline_cv_demo.py
```diff
@@ -0,0 +1,473 @@
...
+
+# %%
+bas = nmo.basis.RaisedCosineBasisLinear(5, mode="conv", window_size=5)
+trans_bas_a = nmo.basis.TransformerBasis(bas)
```
billbrod, line 104:
```suggestion
# these two ways of creating the TransformerBasis are equivalent
trans_bas_a = nmo.basis.TransformerBasis(bas)
```

docs/api_guide/plot_06_sklearn_pipeline_cv_demo.py
```diff
@@ -0,0 +1,473 @@
...
+#
+# Visualize the fit:
+
+# %%
+
+# Predict the rate.
+# Note that you need a 2D input even if x is a flat array.
+# We are using expand dim to add the extra-dimension
+x = np.expand_dims(np.sort(X.flatten()), axis=1)
+predicted_rate = pipeline.predict(x)
```
billbrod, line 155: why are we using both expand_dims and flatten here?

docs/api_guide/plot_06_sklearn_pipeline_cv_demo.py
```diff
@@ -0,0 +1,473 @@
...
+    predicted_rate,
+    label="predicted rate",
+    color="tab:orange",
+)
+
+ax.legend()
+sns.despine(ax=ax)
+
+# %%
+# There is some room for improvement, so in the next section we'll use cross-validation to tune the parameters of our pipeline.
```
billbrod, line 176: describe this a bit more: the current model captures the bimodal distribution of responses, appropriately picking out the peaks. but it doesn't do a good job capturing the actual spike counts: the peaks are too low and the valleys are too low. this might be because of our choice of basis function, so let's see if tuning those parameters results in a better fit! we could do this manually, but doing this with the sklearn pipeline will make everything much easier!

docs/api_guide/plot_06_sklearn_pipeline_cv_demo.py
```diff
@@ -0,0 +1,473 @@
...
+
+# %%
+# ### Select the number of basis by cross-validation
```
billbrod, line 179: I think, for the first example, we should show what the "manual way" of doing this would be (maybe in an admonition). That will make it both more explicit what this is doing, as well as clarifying how much boilerplate this saves you

docs/api_guide/plot_06_sklearn_pipeline_cv_demo.py
```diff
@@ -0,0 +1,473 @@
...
+# Let's define candidate values for the parameters of each step of the pipeline we want to cross-validate. In this case the number of basis functions in the transformation step and the ridge regularization's strength in the GLM fit:
+
+# %%
+param_grid = dict(
+    glm__regularizer_strength=(0.1, 0.01, 0.001, 1e-6),
+    transformerbasis__n_basis_funcs=(3, 5, 10, 20, 100),
+)
+
+# %%
+# !!! note "Grid definition"
```
billbrod, line 198: this is super useful info and may need to exist elsewhere in the docs as well -- this is the kind of thing that people will be referring to pretty much every time they do cross-validation

docs/api_guide/plot_06_sklearn_pipeline_cv_demo.py
```diff
@@ -0,0 +1,473 @@
...
+#     - Add `"__"` followed by the hyperparameter name (for example, `"n_basis_funcs"`).
+#     - If the hyperparameter is itself an object with attributes, add another `"__"` followed by the attribute name. For instance, `"glm__observation_model__inverse_link_function"`
+#       would be a valid key for cross-validating over the link function of the GLM's `observation_model` attribute `inverse_link_function`.
+#     The values in the dictionary are the parameters to be tested.
+
+
+
+# %%
+# #### Run the grid search
+# Let's run a 5-fold cross-validation of the hyperparameters with the scikit-learn `model_selection.GridsearchCV` class.
```
billbrod, line 211: should describe or link to a description of what "5-fold cross-validation" means

docs/api_guide/plot_06_sklearn_pipeline_cv_demo.py
```diff
@@ -0,0 +1,473 @@
...
+gridsearch.fit(X, y)
+
+# %%
+# #### Visualize the scores
+#
+# Let's extract the scores from `gridsearch` and take a look at how the different parameter values of our pipeline influence the test score:
+
+# %%
+from matplotlib.patches import Rectangle
+
```
billbrod, line 230: should we hide this plotting code somewhere?

docs/api_guide/plot_06_sklearn_pipeline_cv_demo.py
```diff
@@ -0,0 +1,473 @@
...
+# Labeling the colorbar
+colorbar = ax.collections[0].colorbar
+colorbar.set_label('log-likelihood')
+
+ax.set_xlabel("ridge regularization strength")
+ax.set_ylabel("number of basis functions")
+
+highlight_max_cell(cvdf_wide, ax)
+
```
billbrod, line 268: should explain what's being shown in the plot, how to interpret it.

src/nemos/basis.py
```diff
@@ -255,7 +482,6 @@ def __init__(
         self.n_basis_funcs = n_basis_funcs
         self._n_input_dimensionality = 0
         self._check_n_basis_min()
-        self._conv_args = args
```
billbrod, line 258: why is this removed? (seems unrelated to this PR)

tests/test_basis.py
```diff
@@ -4543,6 +4660,87 @@ def test_sklearn_transformer_pipeline(bas, poissonGLM_model_instantiation):
     pipe.fit(X[:, : bas._basis._n_input_dimensionality] ** 2, y)
 
 
+@pytest.mark.parametrize(
```
billbrod, line 4663: I think the pipelining tests should be in a new file

billbrod (changes requested):
Code here all looks good to me, my feedback is pretty much all in the example added to the documentation.

- For top-level headers in the documentation, I think we want to be more consistent: either they should all be questions, or all declarative sentences. So "What is a scikit-learn pipeline?", "Why are pipelines useful?", "How do I build a pipeline with NeMoS?", etc. Or "Scikit-learn pipelines", "Pipelines are useful!", "Creating a pipeline with NeMoS", "Creating a custom scorer" etc.
- Also, we currently have 4 top-level headers and 5 numbered points at the top. should promote "Select the number of basis...".
- Looks like `conv_args` is removed in this PR as well. That seems like it should be a separate PR, no?
- We should add link to the new example in the quickstart, "for more detailed examples". and probably in some of our neural modeling examples as well.
- Should add a test that `TransformerBasis(bas)` and `bas.to_transformer()` end up with identical objects.

billbrod:
Oh and thanks for writing up the documentation of the decisions around cloning, pickling, etc. That's really helpful!

src/nemos/basis.py
BalzaniEdoardo: this is removed because variable positional arguments are not allowed for BaseEstimators.

src/nemos/basis.py
```diff
@@ -172,8 +174,37 @@ def fit_transform(self, X: FeatureMatrix, y=None) -> FeatureMatrix:
...
+        Enable easy access to attributes of the underlying Basis object.
+
+        Example
+        -------
+        from nemos import basis
+
+        bas = basis.RaisedCosineBasisLinear(5)
+        trans_bas = basis.TransformerBasis(bas)
+        print(bas.n_basis_funcs)
+        print(trans_bas.n_basis_funcs)
```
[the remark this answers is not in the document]
bagibence: Thanks! I changed this.

src/nemos/basis.py
```diff
@@ -12,7 +12,9 @@
 from numpy.typing import ArrayLike, NDArray
 from pynapple import Tsd, TsdFrame
 from scipy.interpolate import splev
+from sklearn.base import clone as sk_clone
```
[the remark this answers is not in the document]
bagibence: It's not used anymore.

tests/test_basis.py
```diff
@@ -4523,6 +4342,306 @@ def test_power_of_basis(exponent, basis_class):
...
+        basis.CyclicBSplineBasis,
+        basis.RaisedCosineBasisLinear,
+        basis.RaisedCosineBasisLog,
+    ],
+)
+def test_basis_to_transformer(basis_cls):
+    n_basis_funcs = 5
+    bas = basis_cls(n_basis_funcs)
+    assert isinstance(bas.to_transformer(), basis.TransformerBasis)
+    assert bas.to_transformer().n_basis_funcs == bas.n_basis_funcs
```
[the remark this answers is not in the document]
bagibence: Do you mean something like this?

```python
public_attrs_basis = {attr for attr in dir(bas) if not attr.startswith("_")}
public_attrs_transformerbasis = {
    attr for attr in dir(bas.to_transformer()) if not attr.startswith("_")
}

assert public_attrs_transformerbasis - public_attrs_basis == {
    "fit",
    "fit_transform",
    "transform",
}

assert public_attrs_basis - public_attrs_transformerbasis == set()
```

docs/api_guide/plot_06_sklearn_pipeline_cv_demo.py
```diff
@@ -0,0 +1,473 @@
...
+#     ```
+
+# %%
+# ## Create a custom scorer
+# By default, the GLM score method returns the model log-likelihood. If you want to try a different metric, such as the pseudo-R2, you can create a custom scorer that overwrites the default:
+
+# %%
+from sklearn.metrics import make_scorer
+
+# NOTE: the order of the arguments is reversed
```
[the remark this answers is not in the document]
bagibence: Yes, but I think it might be cleaner in a separate PR. I'll raise an issue about it.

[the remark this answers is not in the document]
bagibence: Turns out I already did: https://github.com/flatironinstitute/nemos/issues/193

tests/test_basis.py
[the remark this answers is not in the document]
bagibence: I used `__dict__` to test that the values of attributes are the same instead of just `n_basis_funcs`, and added another test to test the public names using the snippet above.

docs/api_guide/plot_06_sklearn_pipeline_cv_demo.py
BalzaniEdoardo: pretty impressive reduction in lines of code

docs/api_guide/plot_06_sklearn_pipeline_cv_demo.py
BalzaniEdoardo: fixed

docs/api_guide/plot_06_sklearn_pipeline_cv_demo.py
BalzaniEdoardo: I added a note on the quickstart referring to the tutorial. I don't know if this is optimal, but what else can we do about this?

docs/api_guide/plot_06_sklearn_pipeline_cv_demo.py
BalzaniEdoardo: I could not find a way to hide a code cell block with mkdocs-gallery, so i added a api_guide/utils/plotting.py

BalzaniEdoardo (commented):
I should have addressed all comments

docs/quickstart.md
```diff
@@ -189,4 +192,6 @@ Now we can print the best coefficient.
 {'regularizer__regularizer_strength': 0.001}
 ```
 
+For a more detailed example you can checkout our [tutorial on cross-validation](/generated/api_guide/plot_06_sklearn_pipeline_cv_demo).
```
billbrod, line 195: this path is incorrect, and I don't think we need the link appearing twice back-to-back. My inclination would be to merge the two and place it here:
```suggestion
For more information, including a practical example on how to construct a parameter grid and cross-validate hyperparameters across an entire pipeline, please refer to the [tutorial on pipelining and cross-validation](../generated/api_guide/plot_06_sklearn_pipeline_cv_demo).
```

docs/api_guide/plot_06_sklearn_pipeline_cv_demo.py
billbrod: yeah that's a good point. let's leave it for now

docs/api_guide/plot_06_sklearn_pipeline_cv_demo.py
```diff
@@ -209,6 +219,15 @@
 # %%
 # #### Run the grid search
 # Let's run a 5-fold cross-validation of the hyperparameters with the scikit-learn `model_selection.GridsearchCV` class.
+# ??? info "K-Fold cross-validation (click to expand/collapse)"
+#     <p align="center">
+#     <img src="../../../assets/grid_search_cross_validation.png" alt="Grid Search Cross Validation" style="max-width: 80%; height: auto;">
```
billbrod, line 224: this image is very slightly trimmed (the final "s" in "Finding parameters"). Can probably be fixed by sliding over the text label a bit.

And probably worth including this file as an svg? in order to make it easier to edit

docs/api_guide/plot_06_sklearn_pipeline_cv_demo.py
billbrod: ah damn, really? that's frustrating. #158 is the relevant issue here, maybe we can eventually figure something out.

docs/api_guide/plot_06_sklearn_pipeline_cv_demo.py
```diff
@@ -249,31 +317,18 @@ def highlight_max_cell(cvdf_wide, ax):
...
-colorbar.set_label('log-likelihood')
-
-ax.set_xlabel("ridge regularization strength")
-ax.set_ylabel("number of basis functions")
-
-highlight_max_cell(cvdf_wide, ax)
+plotting.plot_heatmap_cv_results(cvdf_wide)
 
 # %%
+# The plot displays the model's log-likelihood for each parameter combination in the grid. The parameter combination with the highest score, which is the one selected by the procedure, is highlighted with a blue rectangle.
```
billbrod, line 323:
```suggestion
# The plot displays the model's log-likelihood for each parameter combination in the grid. The parameter combination with the highest score, which is the one selected by the procedure, is highlighted with a blue rectangle. We can thus see that we need 10 or more basis functions, and that all of the tested regularization strengths agree with each other. In general, we want the fewest number of basis functions required to get a good fit, so we'll choose 10 here.
```

more details.

docs/api_guide/plot_06_sklearn_pipeline_cv_demo.py
```diff
@@ -295,6 +350,9 @@ def highlight_max_cell(cvdf_wide, ax):
 sns.despine(ax=ax)
 
 # %%
+# !!! success
```
billbrod, line 353: while I like the "success" here, it's not consistent with how we previously interpreted the firing rate plots. I tend to think of these boxes as "extra info", rather than "core info", so I think it  should probably be removed. maybe add emojis instead? :tada: :tada:

docs/api_guide/plot_06_sklearn_pipeline_cv_demo.py
```diff
@@ -346,30 +404,16 @@ def highlight_max_cell(cvdf_wide, ax):
...
-highlight_max_cell(cvdf_wide, ax)
 
 # %%
-# Looks like `RaisedCosineBasisLinear` was probably a decent choice for our toy data:
-
+# As shown in the table, the model with the highest score, highlighted in blue, used a RaisedCosineBasisLinear, which appears to be a suitable choice for our toy data.
+# We can confirm that by plotting the firing rate predictions:
 # %%
 
 # Predict the rate using the optimal configuration
```
billbrod, line 415:
```suggestion

# %%
# Predict the rate using the optimal configuration
```

docs/api_guide/plot_06_sklearn_pipeline_cv_demo.py
```diff
@@ -346,30 +404,16 @@ def highlight_max_cell(cvdf_wide, ax):
...
-ax.set_xlabel("ridge regularization strength")
-ax.set_ylabel("number of basis functions")
+plotting.plot_heatmap_cv_results(cvdf_wide)
 
-highlight_max_cell(cvdf_wide, ax)
 
 # %%
-# Looks like `RaisedCosineBasisLinear` was probably a decent choice for our toy data:
-
+# As shown in the table, the model with the highest score, highlighted in blue, used a RaisedCosineBasisLinear, which appears to be a suitable choice for our toy data.
```
billbrod, line 411:
```suggestion
# As shown in the table, the model with the highest score, highlighted in blue, used a RaisedCosineBasisLinear basis (as used above), which appears to be a suitable choice for our toy data. 
```

docs/api_guide/plot_06_sklearn_pipeline_cv_demo.py
```diff
@@ -389,6 +433,9 @@ def highlight_max_cell(cvdf_wide, ax):
 ax.legend()
 sns.despine(ax=ax)
 
+# %%
+# The plot confirms that the firing rate distribution is accurately captured by our model predictions.
```
billbrod, line 437: given how similar the loglikelihood is for MSplineBasis(10) and RaisedCosineBasisLinear(10), I think it would be interesting to plot their predictions as well.

docs/api_guide/plot_06_sklearn_pipeline_cv_demo.py
```diff
@@ -455,19 +502,8 @@ def highlight_max_cell(cvdf_wide, ax):
...
+plotting.plot_heatmap_cv_results(cvdf_wide, label="pseudo-R2")
 
-# Labeling the colorbar
-colorbar = ax.collections[0].colorbar
-colorbar.set_label('pseudo-R2')
-
-ax.set_xlabel("ridge regularization strength")
-ax.set_ylabel("number of basis functions")
+# %%
+# As you can see, with this new metric scores are normalized between 0 and 1, the highest the better.
```
billbrod, line 508:
```suggestion
# As you can see, the results with pseudo-R2 agree with those of the negative log-likelihood. Note that this new metric is normalized between 0 and 1, with a higher score indicating better performance.
```

billbrod (changes requested):
A couple of small changes here and then this is ready to go!

docs/api_guide/plot_06_sklearn_pipeline_cv_demo.py
BalzaniEdoardo: The original png of sklearn was, i will create my own version as svg

billbrod (approved):
Looks good to me!

Issue #181, opened 2024-07-08
Parsing solver name and regularizer


BalzaniEdoardo:
move instantiate_solver from Regularizer to BaseRegressor class: this grabs the strings and construct the solver run, update and init_solver functions.

add a parsing method to BaseRegressor that checks the config compatibility for solver and regularizer.

If you change thing in the regularizer changes will be reflected in glm fit as well.

BalzaniEdoardo:
Solved in #183

Issue #180, opened 2024-07-08
Adding keyword arg to GLM for solver/regularizer


BalzaniEdoardo:

BalzaniEdoardo:
Solver name should be added as keyword argument to GLM.

Regularizer can be passed as a string, or as a class. Default is UnRegularized, standard GLM.

Default for solver should be None and choose the default appropriately. Provide a nice table with the defaults for each regularizer. Raise an error if solver and regularizer are incompatible. Consider adding a flag that bypass the compatibility check (force_solver or something like that).

BalzaniEdoardo:
Something like:

class BaseRegressor(Base, abc.ABC):

     def __init__(
          self,
          regularizer: str | Regularizer = "unregularized",
          solver: str = None
):
          # parse the regularizer and solver options
          # I think the parsing should just do string matching, so maybe we should add a class attribute to each regularizer 
          # that is the string equivalent 
          if not isinstance(regularizer, str):
               self._parse_options(regularizer=regularizer.name, solver=solver)

               # if no error returned, instantiate the solver because regularizer is already done
               self.solver = solver 
          else:
              self._parse_options(regularizer=regularize, solver=solver)

              # if no error raised, continue instantiating regularizer and solver 
              self.solver = solver # store the solver name so that it can be dynamically changed 

              self.regularizer = Regularizer(*kwargs) # instantiate instance

              
     def _parse_options(self, regularizer: str, solver: str):
          pass

     
     def _create_solver(*kwargs):
           # create solver would also need to parse the options just in case something changed 
           # create solver would get called by glm_fit() every time just in case changes have been made to the regularizer or the solver method
              return solver_run, solver_update, solver_init

BalzaniEdoardo:
Solved in #183

PR #183, opened 2024-07-08
Refactoring regularizer and solver 

### LINKED ISSUE #179: Regularizer refactor
Regularizer should have these methods:

1) augment loss: (if unregularized, return the original loss)
2) proximal operator (to run proxgradient solver) (if unregularized, return the identity operator lambda x: x)

clewis7:
closes #179, #180, #181

Previously, the optimizer was being instantiated within a given regularizer. This PR separates the optimizer and regularizer to make `nemos` more modular. The changes listed below make it much easier to instantiate a GLM model and subsequently change the given optimizer or regularizer being used. Below is a summary of the relevant changes and an example :D

**PR Summary:**

`BaseRegressor`
-----------------------
Now houses the regularizer and solver as separate entities. Also handles instantiating the solver when `model.fit()` is called.

New Attributes:
- `solver_name`
     - `str`, must be an appropriate match for the given `regularizer` argument (**note:** check for match gets made when `.fit()` called
- `regularizer`
     - can be either a `str` or `nemos.Regularizer` object
     - if `str`, will instantiate the appropriate `nemos.Regularizer`
- `solver_kwargs`
     - relevants `kwargs` that get passed when `solver` is instantiated 

New Methods:
- `instantiate_solver()`
      - get called in `model.fit()`
      - completes final check to see if `solver_name` is valid for `regularizer`

`GLM`
---------
Takes in `regularizer` and `solver_name` arguments that gets passed to `BaseRegresser`.

Only other notable change is the parsing that occurs in `initialize_solver()` that parses for using `ProximalGradient` or not. 

`Regularizer`
------------------
No longer contains the `solver`. Every regularizer has a `penalized_loss()` and `get_proximal_operator()` method.

New Attributes:
- `regularizer_strength`
     - default value of `1.0`
     - settable for every `regularizer` except for `UnRegularized` because `proximal_operator()` returns identity 
     
New Methods:
- `penalized_loss()`
     - wraps a model's loss function and returns the augmented loss based on the regularizer
     - if no augmentation, simply returns the loss 
- `get_proximal_operator()`
     - used in `ProximalGradient` solver 
     - returns the proximal operator for a given type of regularization to be used when instantiating the solver 

**EXAMPLE**
--------------------
Data:
```python
import jax
import numpy as np
from matplotlib.patches import Rectangle
from sklearn import model_selection

import nemos as nmo

np.random.seed(111)
# random design tensor. Shape (n_time_points, n_features).
X = 0.5*np.random.normal(size=(100, 5))

# log-rates & weights, shape (1, ) and (n_features, ) respectively.
b_true = np.zeros((1, ))
w_true = np.random.normal(size=(5, ))

# sparsify weights
w_true[1:4] = 0.

# generate counts
rate = jax.numpy.exp(jax.numpy.einsum("k,tk->t", w_true, X) + b_true)
spikes = np.random.poisson(rate)
```

Model with various regularizer/solver can be instantiated multiple ways:
```python

model = nmo.glm.GLM() # gives unregularized GLM with default solver as Gradient Descent
model.fit(X, y)

model2 = nmo.glm.GLM(regularizer="ridge") # ridge regression, default solver is Gradient Descent
model2.fit(X, y)

# can easily change solver method
model2.solver_name = "ProximalGradient"
model2.fit(X, y)

model3 = nmo.glm.GLM(regularizer=nmo.regularizer.Lasso(regularizer_strength=0.01)) # lasso regression with default proximal gradient
model3.fit(X, y)

# can easily change regularizer
model3.regularizer = "ridge" # now using ridge regression with proximal gradient
model3.fit(X, y)
```

**TODO**
---------------
- [x] Add better docstring and code comments 
- [x] fix the linting 
- [x] updates tests to reflect changes without losing code coverage


src/nemos/glm.py
```diff
@@ -106,8 +102,11 @@ def regularizer(self) -> Union[None, reg.Regularizer]:
         return self._regularizer
 
     @regularizer.setter
-    def regularizer(self, regularizer: reg.Regularizer):
+    def regularizer(self, regularizer: str | reg.Regularizer):
         """Setter for the regularizer attribute."""
+        # TODO: allow passing str here and then check solver match
+        super()._parse_regularizer_optimizer_params(regularizer=regularizer, solver=self.solver)
```
BalzaniEdoardo, line 108: why this setter is in the GLM class? should it be in the BaseRegressor instead?

src/nemos/glm.py
```diff
@@ -84,12 +81,11 @@ class GLM(BaseRegressor):
     def __init__(
         self,
         observation_model: obs.Observations = obs.PoissonObservations(),
-        regularizer: reg.Regularizer = reg.UnRegularized("GradientDescent"),
```
BalzaniEdoardo, line 87: this may break the sklearn api for `BaseEstimator`. the `estimator.get_params()` will return only the observation model if the parameter is not in the init. If we want to cv the regularization we have to have it at the init

```python
In [3]: mdl = nmo.glm.GLM()
In [4]: mdl.get_params()
Out[4]: 
{'observation_model__inverse_link_function': <PjitFunction of <function jax.numpy.exp at 0x33068f380>>,
 'observation_model': <nemos.observation_models.PoissonObservations at 0x1120e84d0>}
```

src/nemos/glm.py
```diff
@@ -120,7 +119,15 @@ def regularizer(self, regularizer: reg.Regularizer):
...
-        self._regularizer = regularizer
+
+    @property
+    def solver(self) -> str:
+        return self._solver
+
+    @solver.setter
+    def solver(self, solver: str):
+        # check if solver/regularizer pairing valid
+        self._parse_regularizer_optimizer_params(regularizer=self.regularizer, solver=solver)
```
BalzaniEdoardo, line 130:
```suggestion
        self._solver = self._parse_regularizer_optimizer_params(regularizer=self.regularizer, solver=solver)
```
Also this one may belong to BaseRegressor?

Again, why did we need the _solver here? where is it used?

src/nemos/glm.py
clewis7: so you want to pass a `regularizer` arg here? even if it just going to the `super()` call?

BalzaniEdoardo: we need to be able to cross-validate over "regularizer.regularizer_strength" for sure, so the get_params needs to be aware that we have a regularizer parameter.

BalzaniEdoardo: PS and get_params is only aware of positional or keyword and keyword only arguments of the`__init__`

clewis7: should be fixed now!

BalzaniEdoardo: yes! however i don't see the`regularizer__regularizer_strenght` for Ridge for the same reason. ridge only accepts **kwargs now in the init
```
import nemos as nmo
mdl = nmo.glm.GLM(regularizer=nmo.regularizer.Ridge())
mdl.get_params()
Out[7]: 
{'observation_model__inverse_link_function': <PjitFunction of <function jax.numpy.exp at 0x31d003600>>,
 'observation_model': <nemos.observation_models.PoissonObservations at 0x3195834d0>,
 'regularizer': <nemos.regularizer.Ridge at 0x105a442d0>,
 'solver_kwargs': {},
 'solver_name': 'GradientDescent'}

 ```

Comment (clewis7) [reply src/nemos/glm.py:87]: ahhh, I see...will fix

File: src/nemos/regularizer.py [hunk 4]
```diff
@@ -336,56 +210,22 @@ def _penalization(
...
-        def penalized_loss(params, X, y):
+    def penalized_loss(self, loss: Callable) -> Callable:
+        def _penalized_loss(params, X, y):
             return loss(params, X, y) + self._penalization(params)
 
-        return super().instantiate_solver(penalized_loss, *args, **kwargs)
+        return _penalized_loss
 
 
 class ProxGradientRegularizer(Regularizer, abc.ABC):
```
BalzaniEdoardo, line 231: We don't need the proximal gradient class anymore, because all regularizer can be used with prox gradient and should have the get_proximal_operator method

src/nemos/regularizer.py
```diff
@@ -279,10 +139,17 @@ class are defined in the `allowed_solvers` attribute.
...
     ):
-        super().__init__(solver_name, solver_kwargs=solver_kwargs)
+        super().__init__(regularizer_strength=regularizer_strength)
+
+    def penalized_loss(self, loss: Callable):
+        """Unregularized method does not add any penalty."""
+        return loss
 
 
 class Ridge(Regularizer):
```
BalzaniEdoardo, line 155: Ridge, as any other, should also have a `get_proximal_operator` method. Check the prox operator for ridge in the jaxopt codebase.

src/nemos/regularizer.py
```diff
@@ -74,181 +45,68 @@ class Regularizer(Base, abc.ABC):
...
-        def solver_init_state(params, state, *run_args, **run_kwargs) -> NamedTuple:
-            return solver.init_state(
-                params, state, *args, *run_args, **kwargs, **run_kwargs
-            )
-
-        return solver_init_state, solver_update, solver_run
+        pass
 
 
 class UnRegularized(Regularizer):
```
BalzaniEdoardo, line 112: Unregularized has the identity as a proximal operator, this class as any other regularizer should have the get_proximal_operator method too

src/nemos/regularizer.py
```diff
@@ -279,10 +139,17 @@ class are defined in the `allowed_solvers` attribute.
         "LBFGSB",
     )
 
+    _default_solver = "GradientDescent"
+
     def __init__(
-        self, solver_name: str = "GradientDescent", solver_kwargs: Optional[dict] = None
+            self,
+            regularizer_strength: float = 1.0
```
BalzaniEdoardo, line 146: feels strange to have an unused param regularizer_strength. does it look that bad to have that kwarg only for Lasso, Ridge and GroupLasso?

BalzaniEdoardo (changes requested):
Quick review of the regularizer

src/nemos/regularizer.py
clewis7: That's what I was thinking as well...

I think we just have it for those and then we can parse in the glm model when the solver is instantiated whether that should be an argument passed or not

clewis7:
@BalzaniEdoardo should be good for a review before I work on the tests

clewis7:
**PR Summary:**

`BaseRegressor`
-----------------------
Now houses the regularizer and solver as separate entities. Also handles instantiating the solver when `model.fit()` is called.

New Attributes:
- `solver_name`
     - `str`, must be an appropriate match for the given `regularizer` argument (**note:** check for match gets made when `.fit()` called
- `regularizer`
     - can be either a `str` or `nemos.Regularizer` object
     - if `str`, will instantiate the appropriate `nemos.Regularizer`
- `solver_kwargs`
     - relevants `kwargs` that get passed when `solver` is instantiated 

New Methods:
- `instantiate_solver()`
      - get called in `model.fit()`
      - completes final check to see if `solver_name` is valid for `regularizer`

`GLM`
---------
Takes in `regularizer` and `solver_name` arguments that gets passed to `BaseRegresser`.

Only other notable change is the parsing that occurs in `initialize_solver()` that parses for using `ProximalGradient` or not. 

`Regularizer`
------------------
No longer contains the `solver`. Every regularizer has a `penalized_loss()` and `get_proximal_operator()` method.

New Attributes
----------------------
- `regularizer_strength`
     - only used in `ProximalGradient` solver
     - default value of `1.0`
     - settable for every `regularizer` except for `UnRegularized` because `proximal_operator()` returns identity 
     
New Methods
--------------------
- `penalized_loss()`
     - wraps a model's loss function and returns the augmented loss based on the regularizer
     - if no augmentation, simply returns the loss 
- `get_proximal_operator()`
     - used in `ProximalGradient` solver 
     - returns the proximal operator for a given type of regularization to be used when instantiating the solver 


**TODO**
---------------
- [ ] Add better docstring and code comments 
- [ ] fix the linting 
- [ ] updates tests to reflect changes without losing code coverage

src/nemos/base_class.py
```diff
@@ -172,6 +204,32 @@ class BaseRegressor(Base, abc.ABC):
...
+        Default is `None`. If no solver specified, one will be chosen based on the regularizer.
+        Please see table below forregularizer/optimizer pairings.
+
+    +---------------+------------------+-------------------------------------------------------------+
+    | Regularizer   | Default Solver   | Available Solvers                                           |
+    +===============+==================+=============================================================+
+    | UnRegularized | GradientDescent  | GradientDescent, BFGS, LBFGS, ScipyMinimize, NonlinearCG,   |
+    |               |                  | ScipyBoundedMinimize, LBFGSB                                |
+    +---------------+------------------+-------------------------------------------------------------+
+    | Ridge         | GradientDescent  | GradientDescent, BFGS, LBFGS, ScipyMinimize, NonlinearCG,   |
```
BalzaniEdoardo, line 225: Add proximal gradient to Ridge. The ProximalGradient is equivalent to the GradientDescent for Unregularized, so I don't know if we want to explicitly list it for Unregularized too, but it would also work there

src/nemos/base_class.py
```diff
@@ -180,6 +238,181 @@ class BaseRegressor(Base, abc.ABC):
...
+        utils.assert_is_callable(loss, "loss")
+
+        # final check that solver is valid for chosen regularizer
+        if self.solver_name not in self.regularizer.allowed_solvers:
+            raise ValueError(f"The solver: {self.solver_name} is not allowed for "
+                             f"{self._regularizer.__class__} regularizaration. Allowed solvers are "
+                             f"{self._regularizer.allowed_solvers}.")
+
+        # get the solver with given arguments.
+        # The "fun" argument is not always the first one, but it is always KEYWORD
```
BalzaniEdoardo, line 372: Great, thanks for going through the jaxopt API. I feel like they are not too good with consistency at times. Did you have to look at all their solvers or is there some documentation that explains how the solver arguments are structured? If that document exist, i would add a link to it here in the comments

src/nemos/glm.py
```diff
@@ -886,9 +865,20 @@ def initialize_solver(
...
+                                           self.regularizer.regularizer_strength,
+                                           prox=self.regularizer.get_proximal_operator(),
+                                           )
+            )
+
+        else:
+            self._solver_init_state, self._solver_update, self._solver_run = (
+                super().instantiate_solver(self._predict_and_compute_loss)
+            )
+
```
BalzaniEdoardo, line 881: I think this can be made more generic. 

Any model will have an objective function, that we call "_predict_and_compute_loss", any model will pass a loss function to a solver, so this operation is not GLM specific. It may be worth making "_predict_and_compute_loss" an abstract method of base_regressor, and we can move thus part of the code away. We can have a dedicated function, potentially in a new private module that grabs solver and the appropriate arguments (prox etc.). This function will be called directlty from base_regressor. 

In the future, this module can be used to instantiate more complex solvers (like ProjectedGradient, for example, that requires other arguments). 

Does this make sense? what do you think @clewis7 

PS I am not sure that I like the name "predict_and_compute_loss" fir it however, we will always have a predict method because we are following sklearn conventions, so that's maybe fine

src/nemos/regularizer.py
```diff
@@ -336,156 +230,62 @@ def _penalization(
...
+            Ws, bs = params
+            l2reg /= bs.shape[0]
+            # if Ws is a pytree, l2reg needs to be a pytree with the same
+            # structure
+            if isinstance(Ws, (dict, FeaturePytree)):
+                struct = jax.tree_util.tree_structure(Ws)
+                l2reg = jax.tree_util.tree_unflatten(
+                    struct, [l2reg] * struct.num_leaves
+                )
+            return jaxopt.prox.prox_lasso(Ws, l2reg, scaling=scaling), bs
```
BalzaniEdoardo, line 263:
```suggestion
            return jaxopt.prox.prox_ridge(Ws, l2reg, scaling=scaling), bs
```

src/nemos/regularizer.py
```diff
@@ -511,8 +311,11 @@ def prox_op(params, l1reg, scaling=1.0):
 
         return prox_op
 
+    def penalized_loss(self, loss: Callable) -> Callable:
+        return loss
```
BalzaniEdoardo, line 315: you should add the L1 penalty here:
```python
self.regularizer_strength * jnp.sum(jnp.abs(coeff)) / intercept.shape[0]
```

where `intercept` is `params[1]`.

BalzaniEdoardo: mimic what I did for Ridge in the agumentation

src/nemos/regularizer.py
BalzaniEdoardo: Here there is a thing we need to take care of. The ridge regularized models we are implementing for all solvers but ProximalGradient  minimize the following penalized loss

```
loss(coeff, intercept) - 0.5 * reg_strength * norm(coeff)**2 / len(coeff).
```
There is an 0.5 multiplying the penalization. This is convention, and it's the same as sklearn;
Here we need to make sure that the proximal operator also solves the same problem. It may be that the prox operator is minimizing 

```
loss(coeff, intercept) - reg_strength * norm(coeff)**2 / len(coeff).
```

by default, and we need to either divide by two the regularization strength before passing it to the solver, or there is a "scaling" parameter in jaxopt that we may need to set to 0.5. 

The thing to do to verify that we are solving the same problem with all algorithm is to fit a GLM with Ridge with both ProximalGradient and the LBFGS, and check that the params are "allclose". I would add a test for that.

BalzaniEdoardo (changes requested):
This is looking great already, you are doing an awesome job.

src/nemos/glm.py
clewis7: I think making `_predict_and_compute_loss()` abstract would be good because and model will have it, maybe we change the name to `objective_function()`? @BalzaniEdoardo thoughts?

clewis7: For now, I abstracted away the `instantiate_solver()` code back to the `BaseRegressor` but I think because not every solver needs the same thing it would be helpful to have a separate module for handling getting the arguments, etc. like you mentioned above (would be cleaner anyways)

clewis7:
Two important tests to add:
1. testing `np.allclose()` on coeffs for glm using ridge regularizer with proximal gradient vs another solver 
2. testing that every combination of regularizer/solver can be instantiated and fit for every model type

clewis7:
Need to also make sure to update an examples where the solver is being defined in the regularizer still

src/nemos/glm.py
```diff
@@ -666,6 +675,12 @@ def fit(
         else:
             data = X
 
+        # check if mask has been set is using group lasso
+        # if mask has not been set, use a single group as default
+        if self.regularizer.__class__.__name__ == "GroupLasso":
```
BalzaniEdoardo: I see why you are doing this, damn circular imports!

src/nemos/glm.py
```diff
@@ -13,12 +13,13 @@
 from scipy.optimize import root
 
 from . import observation_models as obs
-from . import regularizer as reg
 from . import tree_utils, validation
-from .base_class import DESIGN_INPUT_TYPE, BaseRegressor
+from .base_regressor import BaseRegressor
 from .exceptions import NotFittedError
 from .pytrees import FeaturePytree
+from .regularizer import GroupLasso, Lasso, Regularizer, Ridge
```
BalzaniEdoardo, line 20: I am importing directly now that the circular imports are solverd

src/nemos/glm.py
```diff
@@ -666,6 +672,12 @@ def fit(
         else:
             data = X
 
+        # check if mask has been set is using group lasso
+        # if mask has not been set, use a single group as default
+        if isinstance(self.regularizer, GroupLasso):
```
BalzaniEdoardo, line 677: we can use isinstance since GroupLasso is imported

src/nemos/regularizer.py
```diff
@@ -273,16 +133,32 @@ class are defined in the `allowed_solvers` attribute.
...
+    def get_proximal_operator(
+        self,
+    ) -> ProximalOperator:
+        """Unregularized method has no proximal operator."""
+
+        def prox_op(params, hyperparams, scaling=0.5):
+            Ws, bs = params
+            return jaxopt.prox.prox_none(Ws, hyperparams, scaling=scaling), bs
+
+        return prox_op
```
BalzaniEdoardo, line 161:
```suggestion
```suggestion
        return jaxopt.prox.prox_none
```

This is probably sufficient since prox_none does nothing to both Ws and bs
```

File: src/nemos/regularizer.py [hunk 17]
```diff
@@ -511,8 +300,44 @@ def prox_op(params, l1reg, scaling=1.0):
...
+
+        Returns
+        -------
+        float
+            The Lasso penalization value.
+        """
+
+        def l1_penalty(coeff: jnp.ndarray, intercept: jnp.ndarray) -> jnp.ndarray:
+            return (
+                0.5
```
BalzaniEdoardo, line 322: 0.5 may may be removed, if so, we should set scaling=1.0 above for consistency.

src/nemos/regularizer.py
```diff
@@ -621,7 +448,35 @@ def _check_mask(mask: jnp.ndarray):
...
+            param_with_extra_axis
+        )  # this masks the param, (group, feature, neuron)
+
+        penalty = jnp.linalg.norm(masked_param, axis=0).sum()
+
+        return penalty
+
+    def penalized_loss(self, loss: Callable) -> Callable:
+        def _penalized_loss(params, X, y):
+            return loss(params, X, y) + self._penalization(params)
```
BalzaniEdoardo, line 475: here we are not multiplying by 0.5, so my bet is that we should set the scaling to 1.0 below.

src/nemos/regularizer.py
```diff
@@ -634,7 +489,7 @@ def _get_proximal_operator(
             intercept term is not regularized.
         """
 
-        def prox_op(params, regularizer_strength, scaling=1.0):
+        def prox_op(params, regularizer_strength, scaling=0.5):
```
BalzaniEdoardo, line 492: maybe the scaling should be 1.0 because the penalization doesn't have a 0.5 multiplied to the L1 loss. We should check on a paper or book that derives the proximal operator for a group lasso to see what's the analytical form of the operator given the L1 penalization.

BalzaniEdoardo (changes requested):
This is amazing! I love the new software architecture; I did a couple of improvements and updated the docs.

I left a couple of changes for the penalization terms. I think we should be careful with the scaling, setting it to 0.5 may be needed only for ridge. 

I did a couple of re-arrangement to avoid circular imports:

- `BaseRegressor` has been moved to a separate module 
- All typing have been moved to a dedicated module. If I missed something let me know. Everything should be moved there.

src/nemos/regularizer.py
```diff
@@ -621,7 +443,35 @@ def _check_mask(mask: jnp.ndarray):
...
+            lambda x: self.mask * x, in_axes=0, out_axes=2
+        )  # this vectorizes the product over the neurons, and adds the neuron axis as the last axis
+
+        masked_param = vec_prod(
+            param_with_extra_axis
+        )  # this masks the param, (group, feature, neuron)
+
+        penalty = jnp.linalg.norm(masked_param, axis=0).sum()
+
+        return penalty
```
clewis7, line 466: @BalzaniEdoardo One thing I noticed here is that we are not using the regularization strength and I think that we should be since the proximal operator does

src/nemos/regularizer.py
```diff
@@ -621,7 +459,42 @@ def _check_mask(mask: jnp.ndarray):
                 f"Data type {mask.dtype} provided instead!"
             )
 
-    def _get_proximal_operator(
+    def _penalization(
+        self, params: Tuple[DESIGN_INPUT_TYPE, jnp.ndarray]
+    ) -> jnp.ndarray:
+        """
+        Calculate the penalization.
```
BalzaniEdoardo, line 466: this is private, but for developers we should point out to an equatio/article. In the proximal_operator I have the reference to the group lasso paper

        Yuan, Ming, and Yi Lin. "Model selection and estimation in regression with grouped variables."
        Journal of the Royal Statistical Society Series B: Statistical Methodology 68.1 (2006): 49-67.

The penalized loss is 
$$\\text{loss}(\beta_1,...,\beta_g) + \alpha \cdot \sum _{j=1...,g} \sqrt{\dim(\beta_j)} || \beta_j||_2$$

where $g$ is the number of groups, $\dim(\cdot)$ is the dimension of the vector, i.e. the number of coefficient in each $\beta_j$, and $||\cdot||_2$ is the euclidean norm.

BalzaniEdoardo (commented):
just a couple of notes about the docstrings, but this looks great.

We should work on the tests

src/nemos/regularizer.py
```diff
@@ -74,181 +35,78 @@ class Regularizer(Base, abc.ABC):
...
+    def regularizer_strength(self) -> float:
+        return self._regularizer_strength
 
-    @solver_kwargs.setter
-    def solver_kwargs(self, solver_kwargs: dict):
-        self._check_solver_kwargs(self.solver_name, solver_kwargs)
-        self._solver_kwargs = solver_kwargs
+    @regularizer_strength.setter
+    def regularizer_strength(self, strength: float):
+        self._regularizer_strength = strength
```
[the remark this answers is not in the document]
clewis7: yeah, can add a check and do the conversion

src/nemos/regularizer.py
clewis7: okay sure, will add

clewis7:
warning for hyperparameter `regularizer_strength`

src/nemos/glm.py
```diff
@@ -100,28 +124,6 @@ def __init__(
         self._solver_update = None
         self._solver_run = None
```
BalzaniEdoardo, line 125: I have a way to add the warning that Alex was talking about. 

- Make the Regularizer stateless (pass the `regularizer_strength` as an input to the two methods)
- Make `regularizer_strength` a property of `BaseRegressor` and move the conversion to float there. Default it to None. Whenever `Ridge`/`Lasso`/`GroupLasso` is set as a regularizer in `BaseRegressor` and `regularizer_strength` is None, raise a warning, set the `regularizer_sregnth` to 1. Whenever regularizer is set to `Unregularized` reset the regularizer strength to None, so that the waning will be triggered again if needed.

What do you think?

BalzaniEdoardo:
Since we touched the computation of the dof of the residual in this PR, I generalized the behavior for population GLM. This allows to compute a scale parameter per neuron correctly (the DOF are computed for each neuron), solving #195 too.

src/nemos/base_regressor.py
```diff
@@ -0,0 +1,418 @@
...
+            - solver_update: Function to perform a single update step in the optimization process,
+            returning new parameters and state.
+            - solver_run: Function to execute the optimization process, applying multiple updates until a
+            stopping criterion is met.
+        """
+        # final check that solver is valid for chosen regularizer
+        if self.solver_name not in self.regularizer.allowed_solvers:
+            raise ValueError(
+                f"The solver: {self.solver_name} is not allowed for "
+                f"{self._regularizer.__class__} regularizaration. Allowed solvers are "
```
billbrod, line 231: same point about regularizer name

src/nemos/base_regressor.py
```diff
@@ -0,0 +1,418 @@
...
+                f"{self._regularizer.allowed_solvers}."
+            )
+
+        # only use penalized loss if not using proximal gradient descent
+        if self.solver_name != "ProximalGradient":
```
billbrod, line 236: if I'm understanding correctly, all regularizers have a `penalized_loss` method, but Lasso and GroupLasso can only use the ProximalGradient solver, so they never get used? is that right?

src/nemos/base_regressor.py
```diff
@@ -0,0 +1,418 @@
...
+        # some parsing to make sure solver gets instantiated properly
+        if self.solver_name == "ProximalGradient":
+            self.solver_kwargs.update(prox=self.regularizer.get_proximal_operator())
+            # add self.regularizer_strength to args
+            args += (self.regularizer_strength,)
+        if self.solver_name == "LBFGSB":
+            if "bounds" not in kwargs.keys():
+                warnings.warn(
+                    "Bounds must be provided for LBFGSB. Reverting back to LBFGS solver."
+                )
```
billbrod, line 255: - [ ] I'd rather raise an exception here -- if someone's trying to use LBFGSB but not setting bounds, they've misunderstood something
- [ ] why is this checking `kwargs` and not `self.solver_kwargs`? if I'm understanding the code correctly, there's no way for the user to specify `kwargs` at this point
- [ ] the following fails: `nmo.glm.GLM(solver_name='LBFGSB', solver_kwargs={'bounds': (0,1)})`. I think that's because bounds must be passed at runtime, not initialization, but then, following my point above, I'm unclear how the user can do so (and I can't find an example in either our docs or tests)

src/nemos/base_regressor.py
```diff
@@ -0,0 +1,418 @@
...
+            return solver.update(
+                params, state, *args, *run_args, **kwargs, **run_kwargs
+            )
+
+        def solver_init_state(params, state, *run_args, **run_kwargs) -> NamedTuple:
+            return solver.init_state(
+                params, state, *args, *run_args, **kwargs, **run_kwargs
+            )
+
+        return solver_init_state, solver_update, solver_run
```
billbrod, line 276: out of curiosity: why return these instead of just setting them as attributes here?

src/nemos/glm.py
billbrod: I think that sounds good

src/nemos/glm.py
```diff
@@ -666,6 +673,12 @@ def fit(
         else:
             data = X
 
+        # check if mask has been set is using group lasso
+        # if mask has not been set, use a single group as default
+        if isinstance(self.regularizer, GroupLasso):
+            if self.regularizer.mask is None:
+                self.regularizer.mask = jnp.ones((1, data.shape[1]))
```
billbrod, line 680: is this the same as just doing regular Lasso? then shouldn't we raise an exception because they've probably misunderstood something?

src/nemos/glm.py
```diff
@@ -1001,14 +1020,33 @@ class PopulationGLM(GLM):
...
         Either a matrix of shape (num_features, num_neurons) or a [FeaturePytree](../pytrees) of 0s and 1s, with
         `feature_mask[feature_name]` of shape (num_neurons, ).
         The mask will be used to select which features are used as predictors for which neuron.
 
+    | Regularizer   | Default Solver   | Available Solvers                                           |
+    | ------------- | ---------------- | ----------------------------------------------------------- |
+    | UnRegularized | GradientDescent  | GradientDescent, BFGS, LBFGS, NonlinearCG, ProximalGradient, LBFGSB |
+    | Ridge         | GradientDescent  | GradientDescent, BFGS, LBFGS, NonlinearCG, ProximalGradient, LBFGSB |
+    | Lasso         | ProximalGradient | ProximalGradient                                            |
+    | GroupLasso    | ProximalGradient | ProximalGradient                                            |
```
billbrod, line 1048: same point as `base_regressor.py` about location of this table.

src/nemos/observation_models.py
```diff
@@ -430,7 +430,7 @@ class PoissonObservations(Observations):
 
     def __init__(self, inverse_link_function=jnp.exp):
         super().__init__(inverse_link_function=inverse_link_function)
-        self.scale = 1
+        self.scale = 1.0
```
billbrod, line 433: why the change? doesn't our setter do this conversion for us?

billbrod (commented):
This looks like a great PR! My comments are largely questions and some small changes. I think this PR is ready to be converted from draft, so we can run everything on it.

One additonal note: I think `estimate_resid_degrees_of_freedom` is new to me (though not to this PR). doesn't look like we use it in any examples, so we'll want to make sure to explain it in documentation somewhere (not part of this PR). Do we have a list of "documentation we need" somewhere?

An additional one, related to this PR, is "how to pick regularizers and solvers".

src/nemos/_regularizer_builder.py
```diff
@@ -0,0 +1,45 @@
...
+    elif name == "ridge":
+        from .regularizer import Ridge
+
+        return Ridge()
+    elif name == "lasso":
+        from .regularizer import Lasso
+
+        return Lasso()
+    elif name == "group_lasso":
+        from .regularizer import GroupLasso
```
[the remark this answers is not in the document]
billbrod: especially given that our solver names are the case names (e.g., `ProximalGradient`, not `proximal_gradient`)

src/nemos/base_regressor.py
clewis7: Yes, for now. The goal was to standardize the API to make it agnostic. I think for Lasso and GroupLasso the hope would be that there would be other solvers available in the future that could use the `penalized_loss`. For now, we are using `scipy.optimize.minimize` with the Nelder Mead method to check that the loss functions we have implemented are reaching the same optimum.

src/nemos/glm.py
clewis7: Maybe we raise a warning similar to `regularizer_strength`? Point them to the docs and tell them we are reverting back to `Lasso`

src/nemos/base_regressor.py
BalzaniEdoardo: I think setting the attribute makes sense. It is a leftover from when we had this function as part of the `Regularizer`

src/nemos/glm.py
BalzaniEdoardo: It is actually the opposite of `Lasso`, it treats all the coeffs as a single group. I would keep it as is, so that the two regularizer defaults have different effects, but in the docstrings should make that clear

BalzaniEdoardo: PS Lasso would be mask=np.eye(data.shape[1])

src/nemos/base_regressor.py
clewis7: We just didn't move those properties from the `GLM` class to the `BaseRegressor`, but we could if we wanted to I think. 

@BalzaniEdoardo thoughts on doing this?

src/nemos/regularizer.py
```diff
@@ -296,29 +154,28 @@ class Ridge(Regularizer):
...
         self,
-        solver_name: str = "GradientDescent",
-        solver_kwargs: Optional[dict] = None,
-        regularizer_strength: float = 1.0,
     ):
-        super().__init__(solver_name, solver_kwargs=solver_kwargs)
-        self.regularizer_strength = regularizer_strength
+        super().__init__()
 
     def _penalization(
```
BalzaniEdoardo, line 177: can you make it a static method?

src/nemos/base_regressor.py
```diff
@@ -0,0 +1,465 @@
...
+        Optional dictionary for keyword arguments that are passed to the solver when instantiated.
+        E.g. stepsize, acceleration, value_and_grad, etc.
+         See the jaxopt documentation for details on each solver's kwargs: https://jaxopt.github.io/stable/
+
+    See Also
+    --------
+    Concrete models:
+
+    - [`GLM`](../glm/#nemos.glm.GLM): A feed-forward GLM implementation.
+    - [`GLMRecurrent`](../glm/#nemos.glm.GLMRecurrent): A recurrent GLM implementation.
```
BalzaniEdoardo, line 63: This should be changed: leftover from past implementatoin
```suggestion
    - [`PopulationGLM`](../glm/#nemos.glm.PopulationGLM): A population GLM implementation.
```

src/nemos/base_regressor.py
```diff
@@ -0,0 +1,465 @@
...
+        self,
+        regularizer: Union[str, Regularizer] = "UnRegularized",
+        regularizer_strength: Optional[float] = None,
+        solver_name: str = None,
+        solver_kwargs: Optional[dict] = None,
+    ):
+        self.regularizer = regularizer
+
+        # check regularizer strength
+        if regularizer_strength is None:
```
BalzaniEdoardo, line 76: should warning should not be raised for unregularized

BalzaniEdoardo: I thought I have added a test for this behavior

src/nemos/base_regressor.py
```diff
@@ -0,0 +1,465 @@
...
+            self._regularizer = create_regularizer(name=regularizer)
+        elif isinstance(regularizer, Regularizer):
+            self._regularizer = regularizer
+        else:
+            raise TypeError(
+                f"The regularizer should be either a string from "
+                f"{AVAILABLE_REGULARIZERS} or an instance of `nemos.regularizer.Regularizer`"
+            )
+
+        if isinstance(self._regularizer, UnRegularized):
```
BalzaniEdoardo, line 116: Moving the warning about the regularizer strength in this setter would make it possible to throw a warning every time a user passes from an UnRegularized solver, associated to None as reg strength, to a regularized one.

src/nemos/base_regressor.py
```diff
@@ -0,0 +1,465 @@
...
+
+    @property
+    def regularizer_strength(self) -> float:
+        return self._regularizer_strength
+
+    @regularizer_strength.setter
+    def regularizer_strength(self, strength: float):
+        try:
+            # force conversion to float to prevent weird GPU issues
+            strength = float(strength)
```
BalzaniEdoardo, line 127: maybe we should tree_map this, for future compatibility (proximal methods allows tree structured regularization, this would enable us to be more flexible with the regularizers.
```suggestion
            strength = jax.tree_util.tree_map(lambda x: jnp.asarray(x, dtype=float), strength) 
```

src/nemos/base_regressor.py
```diff
@@ -0,0 +1,465 @@
...
+    Concrete models:
+
+    - [`GLM`](../glm/#nemos.glm.GLM): A feed-forward GLM implementation.
+    - [`GLMRecurrent`](../glm/#nemos.glm.GLMRecurrent): A recurrent GLM implementation.
+    """
+
+    def __init__(
+        self,
+        regularizer: Union[str, Regularizer] = "UnRegularized",
+        regularizer_strength: Optional[float] = None,
```
BalzaniEdoardo, line 69: If we go with the tree_map (see below the setter) we should use the Optional[Any] signature for regularizer strength, this is annoying but may enable regularizing individual parameters in the future.

src/nemos/base_regressor.py
```diff
@@ -0,0 +1,465 @@
...
+
+        # check regularizer strength
+        if regularizer_strength is None:
+            warnings.warn(
+                UserWarning(
+                    "Caution: regularizer strength has not been set. Defaulting to 1.0. Please see "
+                    "the documentation for best practices in setting regularization strength."
+                )
+            )
+            self._regularizer_strength = 1.0
```
BalzaniEdoardo, line 83:
```suggestion
            self.regularizer_strength = 1.0
```
This becomes mandatory if you accept to tree_map the regularizer strength

tests/test_regularizer.py
```diff
@@ -414,14 +555,17 @@ def test_solver_match_sklearn_gamma(self, gammaGLM_model_instantiation):
...
+        model.regularizer_strength = 0.1
+        runner_bfgs = model.instantiate_solver()[2]
         weights_bfgs, intercepts_bfgs = runner_bfgs(
             (true_params[0] * 0.0, true_params[1]), X, y
         )[0]
         model_skl = GammaRegressor(
-            fit_intercept=True, tol=10**-12, alpha=regularizer.regularizer_strength
+            fit_intercept=True,
+            tol=10**-12,
+            alpha=model.regularizer_strength,
```
BalzaniEdoardo, line 568:
```suggestion
            alpha=float(model.regularizer_strength),
```

This is needed if we go for the tree_map in the setter of reg strength

src/nemos/base_regressor.py
```diff
@@ -0,0 +1,465 @@
...
+        solver_init_state_kwargs = dict()
+        solver_update_kwargs = dict()
+        solver_init_kwargs = dict()
+
+        def dummy_loss():
+            pass
+
+        if self.solver_kwargs:
+            # instantiate a solver to then inspect the params of its various functions
+            solver = getattr(jaxopt, self.solver_name)(fun=dummy_loss)
```
BalzaniEdoardo, line 298:
```suggestion
        if self.solver_kwargs:
            # instantiate a solver to then inspect the params of its various functions
            solver = getattr(jaxopt, self.solver_name)
```
this should work

src/nemos/glm.py
```diff
@@ -43,23 +45,42 @@ def wrapper(*args, **kwargs):
...
 
     This GLM implementation allows users to model neural activity based on a combination of exogenous inputs
     (like convolved currents or light intensities) and a choice of observation model. It is suitable for scenarios where
     the relationship between predictors and the response variable might be non-linear, and the residuals
     don't follow a normal distribution.
+    Below is a table listing the default and available solvers for each regularizer.
+
+    | Regularizer   | Default Solver   | Available Solvers                                           |
+    | ------------- | ---------------- | ----------------------------------------------------------- |
+    | UnRegularized | GradientDescent  | GradientDescent, BFGS, LBFGS, NonlinearCG, ProximalGradient, LBFGSB |
```
BalzaniEdoardo, line 58: Aren't we dropping LBFGSB?

src/nemos/glm.py
```diff
@@ -70,57 +91,47 @@ class GLM(BaseRegressor):
         Basis coefficients for the model.
     solver_state :
         State of the solver after fitting. May include details like optimization error.
-    scale:
+    scale_:
         Scale parameter for the model. The scale parameter is the constant $\Phi$, for which
         $\text{Var} \left( y \right) = \Phi V(\mu)$. This parameter, together with the estimate
         of the mean $\mu$ fully specifies the distribution of the activity $y$.
+    dof:
```
BalzaniEdoardo, line 98:
```suggestion
    dof_resid_:
```

src/nemos/glm.py
```diff
@@ -70,57 +91,47 @@ class GLM(BaseRegressor):
...
-        # test solver instantiation on the GLM loss
-        try:
-            regularizer.instantiate_solver(self._predict_and_compute_loss)
-        except Exception:
-            raise TypeError(
-                "The provided `solver` cannot be instantiated on "
-                "the GLM log-likelihood."
-            )
-        self._regularizer = regularizer
+        self.dof = None
```
BalzaniEdoardo, line 134:
```suggestion
        self.dof_resid_ = None
```

src/nemos/glm.py
```diff
@@ -679,9 +705,9 @@ def fit(
 
         self._set_coef_and_intercept(params)
 
-        dof_resid = self.estimate_resid_degrees_of_freedom(X)
-        self.scale = self.observation_model.estimate_scale(
-            self._predict(params, data), y, dof_resid=dof_resid
+        self.dof = self._estimate_resid_degrees_of_freedom(X)
```
BalzaniEdoardo, line 708:
```suggestion
        self.dof_resid_ = self._estimate_resid_degrees_of_freedom(X)
```

src/nemos/glm.py
```diff
@@ -976,9 +1032,9 @@ def update(
         self.solver_state = opt_step[1]
 
         # estimate the scale
-        dof_resid = self.estimate_resid_degrees_of_freedom(X, n_samples=n_samples)
-        self.scale = self.observation_model.estimate_scale(
-            self._predict(params, data), y, dof_resid=dof_resid
+        self.dof = self._estimate_resid_degrees_of_freedom(X, n_samples=n_samples)
```
BalzaniEdoardo, line 1035:
```suggestion
        self.dof_resid_ = self._estimate_resid_degrees_of_freedom(X, n_samples=n_samples)
```

src/nemos/glm.py
```diff
@@ -976,9 +1032,9 @@ def update(
...
+        self.scale_ = self.observation_model.estimate_scale(
+            self._predict(params, data), y, dof_resid=self.dof
```
BalzaniEdoardo, line 1037:
```suggestion
            self._predict(params, data), y, dof_resid=self.dof_resid_
```

src/nemos/glm.py
```diff
@@ -994,16 +1050,36 @@ class PopulationGLM(GLM):
     It is suitable for scenarios where the relationship between predictors and the response
     variable might be non-linear, and the residuals  don't follow a normal distribution. The predictors must be
     stored in tabular format, shape (n_timebins, num_features) or as [FeaturePytree](../pytrees).
+    Below is a table listing the default and available solvers for each regularizer.
+
+    | Regularizer   | Default Solver   | Available Solvers                                           |
+    | ------------- | ---------------- | ----------------------------------------------------------- |
+    | UnRegularized | GradientDescent  | GradientDescent, BFGS, LBFGS, NonlinearCG, ProximalGradient, LBFGSB |
+    | Ridge         | GradientDescent  | GradientDescent, BFGS, LBFGS, NonlinearCG, ProximalGradient, LBFGSB |
```
BalzaniEdoardo, line 1058: Dropping LBFGSB?

src/nemos/glm.py
```diff
@@ -994,16 +1050,36 @@ class PopulationGLM(GLM):
...
         Default is the Poisson model.
     regularizer :
-        Regularization to use for model optimization. Defines the regularization scheme, the optimization algorithm,
+        Regularization to use for model optimization. Defines the regularization scheme
         and related parameters.
-        Default is UnRegularized regression with gradient descent.
+        Default is UnRegularized regression.
+    regularizer_strength :
+        Float that is default None. Sets the regularizer strength. If a user does not pass a value, and it is needed for
+        regularization, a warning will be raised and the strength will default to 1.0.
```
BalzaniEdoardo, line 1073: Change description if we allow pytrees.

src/nemos/observation_models.py
```diff
@@ -797,9 +797,7 @@ def deviance(
         where $ y $ is the observed data, $ \hat{y} $ is the predicted data, and $\text{LL}$ is the model
         log-likelihood. Lower values of deviance indicate a better fit.
         """
-        y_mu = jnp.clip(
-            neural_activity / predicted_rate, min=jnp.finfo(predicted_rate.dtype).eps
-        )
+        y_mu = jnp.clip(neural_activity / predicted_rate, min=jnp.finfo(float).eps)
```
BalzaniEdoardo, line 800: @billbrod this matches statsmodels computation. it is more precise than what we had previously (which for jax is usually float32), and resulted in discrepancies with statsmodels due to very low values.

src/nemos/regularizer.py
```diff
@@ -78,177 +38,58 @@ class Regularizer(Base, abc.ABC):
...
-        Parameters
-        ----------
-        solver_name :
-            Name of the solver.
-        solver_kwargs :
-            Additional keyword arguments for the solver.
+        loss :
+            Callable loss function.
+        regularizer_strength :
+            Float the indicates the regularization strength.
```
BalzaniEdoardo, line 71: revisit description if pytree

src/nemos/base_regressor.py
BalzaniEdoardo: Ignore this and related comments. other PR

src/nemos/_regularizer_builder.py
```diff
@@ -0,0 +1,45 @@
...
+
+
+def create_regularizer(name: str):
+    """
+    Create a regularizer from a given name.
+
+    Parameters
+    ----------
+    name :
+        The string name of the regularizer to create. Must be one of: 'unregularized', 'ridge', 'lasso', 'group_lasso'.
```
BalzaniEdoardo, line 13:
```suggestion
        The string name of the regularizer to create. Must be one of: 'UnRegularized', 'Ridge', 'Lasso', 'GroupLasso'.
```

src/nemos/base_regressor.py
BalzaniEdoardo: Yeah, I think this should be a common attribute of any BaseRegressor

src/nemos/base_regressor.py
billbrod: Resolved by dropping support for LBFGSB, will add support back if needed.

src/nemos/base_regressor.py
billbrod: I think this makes sense, with some additional explanatory comments that Edoardo is adding.

src/nemos/base_regressor.py
billbrod: now setting attributes here.

billbrod (approved):
This looks good to me! :tada:

BalzaniEdoardo (approved):

PR #184, opened 2024-07-10
Add (Prox-)SVRG as a solver for GLMs

bagibence:
Add a preliminary implementation of [SVRG](https://papers.nips.cc/paper_files/paper/2013/hash/ac1dd209cbcc5e5d1c6e28598e8cbbe8-Abstract.html) and [Prox-SVRG](https://arxiv.org/abs/1403.4699v1) compatible with the jaxopt API, so nemos can use it as a solver:

- the `run` method (used in `GLM.fit`) assumes the full data is loaded into memory and runs optimization until convergence.
- the `update` method (used in `GLM.update`) updates the parameters on a single data point or batch. Note that currently, when fitting on batches and calling `glm.update`, some steps of the algorithm have to be implemented outside.

`ProxSVRG` takes a `prox` argument for the proximal operator which should be a callable with the same signature as the functions defined in `jaxopt.prox` (e.g. `jaxopt.prox.prox_lasso`).
`SVRG` inherits from `ProxSVRG`, uses the identity as the proximal operator and adjusts some calls to account for the different number of arguments.


src/nemos/solvers.py
```diff
@@ -0,0 +1,726 @@
...
+            ind = random.randint(subkey, (self.batch_size,), 0, N)
+
+            # perform a single update on the mini-batch or data point
+            xk = self._xk_update(
+                xk,
+                xs,
+                df_xs,
+                state.stepsize,
+                prox_lambda,
+                *(tree_slice(arg, ind) for arg in args),
```
BalzaniEdoardo, line 517:
```suggestion
                *tree_slice(args, ind),
```

This is sufficient because you are sure that args is a tuple, the output of a tree_map is going to be a tuple (same tree struct as args), and you can unpack output of tree slice directly.

src/nemos/solvers.py
```diff
@@ -0,0 +1,726 @@
...
+        state = state._replace(
+            iter_num=state.iter_num + 1,
+            key=key,
+            x_av=tree_scalar_mul(1 / m, x_sum),
+        )
+
+        # the next anchor point is the parameters at the end of the inner loop
+        # or the average over the inner loop
+        return OptStep(params=xk, state=state)
+        # return OptStep(params=state.x_av, state=state)
```
BalzaniEdoardo, line 547:
```suggestion
```

src/nemos/solvers.py
```diff
@@ -0,0 +1,726 @@
...
+
+    # @staticmethod
+    # def _error(x, x_prev, stepsize):
+    #    diff_norm = tree_l2_norm(tree_sub(x, x_prev))
+    #    return diff_norm / stepsize
```
BalzaniEdoardo, line 552:
```suggestion
```

src/nemos/solvers.py
```diff
@@ -0,0 +1,726 @@
...
+    @staticmethod
+    def _error(x, x_prev, stepsize):
+        return tree_l2_norm(tree_sub(x, x_prev)) / tree_l2_norm(x_prev)
+
+    # @staticmethod
+    # def _error(x, x_prev, stepsize):
+    #    # adapted from scikit-learn's SAG solver
+    #    diff = tree_sub(x, x_prev)
+    #    flat_diff, _ = jax.flatten_util.ravel_pytree(diff)
+    #    return jnp.max(jnp.abs(flat_diff))
```
BalzaniEdoardo, line 563:
```suggestion
```

src/nemos/solvers.py
```diff
@@ -0,0 +1,726 @@
...
+    def loss_fn(params, X, y):
+        ...
+
+    svrg = SVRG(loss_fn)
+    params, state = svrg.run(init_params, X, y)
+
+    References
+    ----------
+    Prox-SVRG - https://arxiv.org/abs/1403.4699v1
+    SVRG - https://papers.nips.cc/paper_files/paper/2013/hash/ac1dd209cbcc5e5d1c6e28598e8cbbe8-Abstract.html
```
BalzaniEdoardo, line 599: reference as in the suggestion of ProxSVRG

src/nemos/solvers.py
```diff
@@ -0,0 +1,726 @@
...
+        ----------
+        init_params : Any
+            Initial parameters to start from.
+        args:
+            Positional arguments passed to loss function `fun` and its gradient.
+            For GLMs it is:
+                X :
+                    Input data.
+                y :
+                    Output data.
```
BalzaniEdoardo, line 706: remember to update the docstrings as in proxSVRG

src/nemos/solvers_naive.py
BalzaniEdoardo, line 1: should we remove this? I don't think we are using it

BalzaniEdoardo (changes requested):
This is looking great already. My comments are pretty minor, very good job! 

This PR is quite big already, so I would not add a tutorial for the docs in this one, we should open a dedicated PR for it.

I'll add an issue describing what I am envisioning for that!

src/nemos/solvers_naive.py
bagibence: Yes, I'll remove that.

src/nemos/solvers.py
```diff
@@ -0,0 +1,726 @@
...
+
+      objective(params, hyperparams_prox, *args, **kwargs) =
+        fun(params, *args, **kwargs) + non_smooth(params, hyperparams_prox)
+
+    Attributes
+    ----------
+    fun: Callable
+        smooth function of the form ``fun(x, *args, **kwargs)``.
+    prox: Callable
+        proximity operator associated with the function ``non_smooth``.
```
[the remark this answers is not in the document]
bagibence: I think I copied it from jaxopt, but I also like proximal better.

src/nemos/solvers.py
```diff
@@ -0,0 +1,726 @@
...
+KeyArrayLike = ArrayLike
+
+
+class SVRGState(NamedTuple):
+    iter_num: int
+    key: KeyArrayLike
+    error: float
+    stepsize: float
+    # loss_log: jnp.ndarray
+    xs: Optional[tuple] = None
```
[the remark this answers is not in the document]
bagibence: I think it could be any pytree, it was just `(Ws, b)` for GLMs.

src/nemos/solvers.py
bagibence: Cool!

src/nemos/solvers.py
```diff
@@ -0,0 +1,726 @@
...
+    df_xs: Optional[tuple] = None
```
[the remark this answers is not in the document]
bagibence: I also like explicit names, but in this case I might prefer these as they are closer to the math and the pseudo-code in the papers. How about explaining what they are in multiple places?

src/nemos/solvers.py
```diff
@@ -0,0 +1,726 @@
...
+        state : SVRGState
+            Initialized optimizer state
+        """
+        df_xs = None
+        if init_full_gradient:
+            df_xs = self.loss_gradient(init_params, *args)
+
+        state = SVRGState(
+            iter_num=0,
+            key=self.key if self.key is not None else random.key(0),
```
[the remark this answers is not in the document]
bagibence: Done.

src/nemos/solvers.py
bagibence: Removed.

src/nemos/solvers.py
bagibence: Removed.

src/nemos/solvers.py
```diff
@@ -0,0 +1,726 @@
...
+                    Output data.
+
+
+        Returns
+        -------
+        OptStep
+            xs : Any
+                Parameters after taking one step defined in the inner loop of Prox-SVRG.
+            state : SVRGState
+                Updated state.
```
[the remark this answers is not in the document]
bagibence: Added.

src/nemos/solvers.py
```diff
@@ -0,0 +1,726 @@
...
+        hyperparams_prox : float
+            Parameters of the proximal operator, in our case the regularization strength.
+            Not used here, but required to be consistent with the jaxopt API.
+        args:
+            Positional arguments passed to loss function `fun` and its gradient.
+            For GLMs it is:
+                X :
+                    Input data.
+                y :
+                    Output data.
```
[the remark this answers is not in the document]
bagibence: I changed this everywhere. Let me know if it's good like this. I also thought about having the full description only in one place, then referring to it from other methods. How would you like that?

src/nemos/solvers.py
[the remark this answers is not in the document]
bagibence: Changed it to `Pytree`.

src/nemos/solvers.py
```diff
@@ -0,0 +1,726 @@
...
+
+        Parameters
+        ----------
+        loss_log : jnp.ndarray
+            1D array storing the loss values.
+        i : int
+            Index at which to update, most likely the current iteration number.
+        params : Any
+            Parameters with which to evaluate the loss.
+        args:
```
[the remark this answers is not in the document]
bagibence: Done.

src/nemos/solvers.py
```diff
@@ -0,0 +1,726 @@
...
+        OptStep
+            xs : Any
+                Parameters at the end of the last inner loop.
+                (... or the average of the parameters over the last inner loop)
+            state : SVRGState
+                Updated state.
+        """
+        n_points_per_arg = {leaf.shape[0] for leaf in jax.tree.leaves(args)}
+        if not len(n_points_per_arg) == 1:
+            raise ValueError("All arguments must have the same sized first dimension.")
```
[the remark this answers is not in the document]
bagibence: I just tried to measure this by moving it up into the `_run` method, and I don't think it matters, so I would leave it here.
I removed the calculations for `x_av` which is never used now, and that might actually save some time, but I didn't measure that.

src/nemos/solvers.py
```diff
@@ -0,0 +1,726 @@
...
+            For GLMs it is:
+                X :
+                    Input datapoint or mini-batch.
+                y :
+                    Output datapoint or mini-batch.
+        """
+        dfik_xk = self.loss_gradient(xk, *args)
+        dfik_xs = self.loss_gradient(xs, *args)
+
+        gk = jax.tree_util.tree_map(lambda a, b, c: a - b + c, dfik_xk, dfik_xs, df_xs)
```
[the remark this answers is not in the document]
bagibence: Added the comment.

src/nemos/solvers.py
```diff
@@ -0,0 +1,726 @@
...
+        args:
+            Hyperparameters passed to `prox` and positional arguments passed to loss function `fun` and its gradient.
+            For GLMs it is:
+                X :
+                    Input datapoint or mini-batch.
+                y :
+                    Output datapoint or mini-batch.
+        """
+        dfik_xk = self.loss_gradient(xk, *args)
+        dfik_xs = self.loss_gradient(xs, *args)
```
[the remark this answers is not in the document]
bagibence: I renamed these to `grad_of_fik_at_xk` and `grad_of_fik_at_xs`. We might want to rename `df_xs` to `grad_of_f_at_xs` as well. These might be a good compromise that's kind of close to that math but easier to interpret. What do you think?

I also renamed `final_xs` to `final_params` where it occurred.

src/nemos/solvers.py
```diff
@@ -0,0 +1,726 @@
...
+                    Output data.
+
+        Returns
+        -------
+        Updated loss log.
+        """
+        return loss_log.at[i].set(self.fun(params, *args))
+
+    @partial(jit, static_argnums=(0,))
+    def _xk_update(
```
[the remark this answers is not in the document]
bagibence: I renamed this to `_xk_update_step` to signal that it's a single gradient step. I think I'm leaning towards leaving it like this instead of saying parameter because it's more explicit which step of the SVRG algorithm it refers to, whereas parameter step/update is a bit more vague to me. I feel like it might also discourage misuse or confusion with `update`.

src/nemos/solvers.py
```diff
@@ -0,0 +1,726 @@
...
+                    Input datapoint or mini-batch.
+                y :
+                    Output datapoint or mini-batch.
+        """
+        dfik_xk = self.loss_gradient(xk, *args)
+        dfik_xs = self.loss_gradient(xs, *args)
+
+        gk = jax.tree_util.tree_map(lambda a, b, c: a - b + c, dfik_xk, dfik_xs, df_xs)
+
+        next_xk = tree_add_scalar_mul(xk, -stepsize, gk)
```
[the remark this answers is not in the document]
bagibence: How about `x_k_plus_one` / `x__k_plus_one`?

src/nemos/solvers.py
```diff
@@ -0,0 +1,725 @@
...
+from jax import grad, jit, lax, random
+from jaxopt import OptStep
+from jaxopt._src import loop
+from jaxopt.prox import prox_none
+
+from .tree_utils import tree_add_scalar_mul, tree_l2_norm, tree_slice, tree_sub
+from .typing import KeyArrayLike, Pytree
+
+
+class SVRGState(NamedTuple):
```
billbrod, line 16: should we be using a NamedTuple for this or a [dataclass](https://docs.python.org/3/library/dataclasses.html)?

From [this stackoverflow](https://stackoverflow.com/questions/51671699/data-classes-vs-typing-namedtuple-primary-use-cases), it looks like NamedTuples make sense when the data structure is immutable, but the state does get changed, right?

src/nemos/solvers.py
```diff
@@ -0,0 +1,725 @@
...
+        Number of data points to sample per inner loop iteration.
+
+    Example
+    -------
+    def loss_fn(params, X, y):
+        ...
+
+    svrg = ProxSVRG(loss_fn, prox_fun)
+    params, state = svrg.run(init_params, prox_lambda, X, y)
+
```
billbrod, line 61: Example is not rendered correctly. I believe it should be [formatted like](https://numpydoc.readthedocs.io/en/latest/format.html#examples):
```suggestion
    Examples
    --------
    >>> def loss_fn(params, X, y):
    >>>    ...
    >>>
    >>> svrg = ProxSVRG(loss_fn, prox_fun)
    >>> params, state = svrg.run(init_params, prox_lambda, X, y)

```

src/nemos/solvers.py
```diff
@@ -0,0 +1,725 @@
...
+        init_full_gradient: bool = False,
+    ) -> SVRGState:
+        """
+        Initialize the solver state
+
+        Parameters
+        ----------
+        init_params :
+            Pytree containing the initial parameters.
+            For GLMs it's a tuple of (W, b)
```
billbrod, line 101: this docstirng is irrelevant, right? shouldn't it be describing what the SVRG needs?

src/nemos/solvers.py
```diff
@@ -0,0 +1,725 @@
...
+        hyperparams_prox :
```
billbrod, line 102: since this is unused, should we raise an error or warning if it gets set?

src/nemos/solvers.py
```diff
@@ -0,0 +1,725 @@
...
+from .typing import KeyArrayLike, Pytree
+
+
+class SVRGState(NamedTuple):
+    iter_num: int
+    key: KeyArrayLike
+    error: float
+    stepsize: float
+    xs: Optional[Pytree] = None
+    df_xs: Optional[Pytree] = None
```
billbrod, line 22: I think this naming is unclear. is it documented anywhere? can we give this class a docstring explaining what they are? and is this naming matching a convention? if so, should point that out. if not, I think it should be changed to something more clear

src/nemos/solvers.py
```diff
@@ -0,0 +1,725 @@
...
+        prox_lambda: Union[float, None],
+        *args,
+    ) -> OptStep:
+        """
+        Perform a single parameter update on the passed data (no random sampling or loops)
+        and increment `state.iter_num`.
+
+        Please note that this is called by `GLM.update`, but repeated calls to `GLM.update`
+        on mini-batches passed to it will not result in running the full (Prox-)SVRG,
+        and parts of the algorithm will have to be implemented outside.
```
billbrod, line 211: - currently, this gets called by `GLM.update`, but it will get called by any model update step, right? I would say something like "This gets called by `BaseRegressor.solver_update` (e.g., as called by `GLM.update`)..."
- For "repeated calls ... will not result in running the full SVRG": in order to run the full SVRG, they should call `.run`, right? We should say that instead of making it sound like they need to implement it themselves

src/nemos/solvers.py
```diff
@@ -0,0 +1,725 @@
...
+
+        Returns
+        -------
+        OptStep
+            xs :
+                Parameters after taking one step defined in the inner loop of Prox-SVRG.
+            state :
+                Updated state.
+        """
+        # NOTE this doesn't update state.xs, state.df_xs, that has to be done outside
```
billbrod, line 296: this is already in the docstring, so can be removed

src/nemos/solvers.py
```diff
@@ -0,0 +1,725 @@
...
+        self,
+        current_params: Pytree,
+        state: SVRGState,
+        prox_lambda: Union[float, None],
+        *args,
+    ) -> OptStep:
+        """
+        Update parameters given a mini-batch of data and increment iteration/epoch number in state.
+
+        Note that this method doesn't update state.xs, state.df_xs, that has to be done outside.
```
billbrod, line 266: would clarify: "e.g., as is done in ..."

src/nemos/solvers.py
```diff
@@ -0,0 +1,725 @@
...
+    @partial(jit, static_argnums=(0,))
+    def run(
+        self,
+        init_params: Pytree,
+        prox_lambda: Union[float, None],
+        *args,
+    ) -> OptStep:
+        """
+        Run a whole optimization until convergence or until `maxiter` epochs are reached.
+        Called from `GLM.fit` and assumes that X and y are the full data set.
```
billbrod, line 317: same point about GLM being too specific

src/nemos/solvers.py
```diff
@@ -0,0 +1,725 @@
...
+    def _run(
+        self,
+        init_params: Pytree,
+        init_state: SVRGState,
+        prox_lambda: Union[float, None],
+        *args,
+    ) -> OptStep:
+        """
+        Run a whole optimization until convergence or until `maxiter` epochs are reached.
+        Called from `GLM.fit` and assumes that X and y are the full data set.
```
billbrod, line 365: same point about GLM being too specific

src/nemos/solvers.py
```diff
@@ -0,0 +1,725 @@
...
+    stepsize : float
+        Constant step size to use.
+    tol: float
+        Tolerance level for the error when comparing parameters
+        at the end of consecutive epochs to check for convergence.
+    batch_size: int
+        Number of data points to sample per inner loop iteration.
+
+    Example
+    -------
```
billbrod, line 571: same point about this not rendering correctly

src/nemos/tree_utils.py
```diff
@@ -140,3 +141,65 @@ def pytree_map_and_reduce(
     cond_tree = jax.tree_util.tree_map(map_fn, *pytrees, is_leaf=is_leaf)
     # for some reason, tree_reduce doesn't work well with any.
     return reduce_fn(jax.tree_util.tree_leaves(cond_tree))
+
+
+def tree_slice(data: Any, idx):
```
billbrod, line 146: I thought slicing already worked on our pytrees, why was this necessary?

billbrod (changes requested):
I don't understand the algorithm, so I'm not looking at its details here. I'm trusting you and Edoardo have that taken care of!

Most of my comments are about documentation.

- Without the example, it's a little hard for me to see how this all works, but I agree with Edoardo that it's fine to have a separate PR for that. I don't see an issue for that right now -- @bagibence can you create one and drop any relevant notes in there before you leave?
- Why are the tree changes in this PR as well? Are they related to the solver?

src/nemos/solvers.py
bagibence: The state does get updated, but by creating a new one (`._replace` creates a copy) instead of modifying the original one.
I basically just copied [how jaxopt defines optimizer states](https://github.com/google/jaxopt/blob/6d882923a54feb6f68b46193d9787cc73bfdadd3/jaxopt/_src/proximal_gradient.py#L84), which might have been inspired by [this example](https://jax.readthedocs.io/en/latest/stateful-computations.html#simple-worked-example-linear-regression).

src/nemos/solvers.py
bagibence: Changed it.

src/nemos/solvers.py
bagibence: Updated.

src/nemos/solvers.py
bagibence: What do you mean by it getting set?

[This is how jaxopt treats this](https://github.com/google/jaxopt/blob/6d882923a54feb6f68b46193d9787cc73bfdadd3/jaxopt/_src/proximal_gradient.py#L179). I wanted to be consistent with them.

src/nemos/solvers.py
bagibence: - Re `BaseRegressor._solver_update`: Yes, I'll change it.
- Re `.run`: `.run` and `.fit` assumes you have the full data in memory and can index into it. If you don't, you have to rely on `.update` and you do have to implement some parts of the SVRG algorithm yourself. This is in contrast with SGD (shown in the batching example), where you just have to call `.update` on each new minibatch. We have tests that the full SVRG algorithm calling `.update`, but we still need to make a tutorial in the docs. In the future this would be hidden from the user with an alternative interface for batched fitting. These might be relevant:
https://github.com/flatironinstitute/nemos/issues/194
https://github.com/flatironinstitute/nemos/issues/203

bagibence: I addressed the first point. Let me know how you like the new description.

src/nemos/solvers.py
bagibence: Changed it. Let me know if you like it the new way.

src/nemos/solvers.py
bagibence: Changed it. Let me know if you like it the new way.

src/nemos/solvers.py
bagibence: Removed it.

src/nemos/solvers.py
bagibence: Do you mean "as is done in `ProxSVRG.run`"? Or how `state.xs` and `state.df_xs` can be updated?

src/nemos/solvers.py
bagibence: They are documented in the implementation of the algorithm, but we can add a docstring here as well.
The naming is trying to match the pseudocode provided in [this paper](https://arxiv.org/abs/2010.00892).

src/nemos/tree_utils.py
bagibence: It's for slicing any pytree, not just `FeaturePytree`.

src/nemos/solvers.py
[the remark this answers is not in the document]
bagibence: I renamed it to `_inner_loop_param_update_step`

src/nemos/solvers.py
```diff
@@ -0,0 +1,726 @@
...
+        Updated loss log.
+        """
+        return loss_log.at[i].set(self.fun(params, *args))
+
+    @partial(jit, static_argnums=(0,))
+    def _xk_update(
+        self,
+        xk: Any,
+        xs: Any,
+        df_xs: Any,
```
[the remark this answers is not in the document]
bagibence: Changed it.

src/nemos/solvers.py
```diff
@@ -0,0 +1,726 @@
...
+        Returns
+        -------
+        Updated loss log.
+        """
+        return loss_log.at[i].set(self.fun(params, *args))
+
+    @partial(jit, static_argnums=(0,))
+    def _xk_update(
+        self,
+        xk: Any,
```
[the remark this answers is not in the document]
bagibence: Renamed to `params` everywhere.

src/nemos/solvers.py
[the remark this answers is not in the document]
bagibence: I ended up with `next_params`.

src/nemos/solvers.py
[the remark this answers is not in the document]
bagibence: We ended up calling them `minibatch_grad_at_current_params` and `minibatch_grad_at_reference_point` with some comments about how they relate to the pseudocode in Gower et al. 2020.

src/nemos/solvers.py
```diff
@@ -0,0 +1,726 @@
...
+        on mini-batches passed to it will not result in running the full (Prox-)SVRG,
+        and parts of the algorithm will have to be implemented outside.
+
+        Parameters
+        ----------
+        current_params : Any
+            Parameters at the end of the previous update, used as the starting point for the current update.
+        state : SVRGState
+            Optimizer state at the end of the previous update.
+            Needs to have the current anchor point (xs) and the gradient at the anchor point (df_xs) already set.
```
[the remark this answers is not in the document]
bagibence: Done.

src/nemos/solvers.py
bagibence: Updated!

src/nemos/solvers.py
billbrod: I think that's fine, but what I meant was that if these are specified, we should raise a warning or error. as in `if hyperparams_prox is not None`. but actually, I think I'm confused -- why isn't this argument optional?

src/nemos/solvers.py
billbrod: ah that's right, I forgot about that. as long as we remember to show it in the tutorial, that's fine.

src/nemos/solvers.py
billbrod: I think this is related to the tutorial for `update` that we're going to add. If I understand correctly, the updating "that has to be done outside" isn't currently shown anywhere (except for a test), right? so this will just need to point to that new tutorial that we write

src/nemos/solvers.py
```diff
@@ -14,12 +14,39 @@
...
+    key :
+        Random key to use when sampling data points or mini-batches.
+    error :
+        Scaled difference (~distance) between subsequent parameter values
+        used to monitor convergence.
+    stepsize :
+        Step size of the individual gradient steps.
+    reference_point :
+        Anchor/reference/snapshot point where the full gradient is calculated in the SVRG algorithm.
+        Corresponds to `x_{s}` in the pseudocode in [1]
```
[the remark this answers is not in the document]
billbrod: @BalzaniEdoardo did you figure this out for mkdocs?

billbrod (changes requested):
This generally looks very good! Mainly some minor changes here, and I responded in a couple of my previous comments, so make sure to check those again.

Also, and I know this is annoying, but `test_solvers` still has references to `xs` and `df_xs`. I think that should be updated to match the new nomenclature, which will make it easier for us in the future.

src/nemos/solvers.py
BalzaniEdoardo: unsure why, but in the general `jaxopt.ProximalGradient` API, both `init_state` and `update` require a `hyperparameter_prox` input. 

This is an annoying choice, they should have make it keyword and optional, but in their case it made sense to have an hyperparameter variable at state initialization since you cannot anticipate if a specific proximal method would require it or not.

For us, I don't think we should care about this parameter for this ProxSVRG since we are not implementing a base class for any stochastic solver (which we should, but not in this PR). I'll remove it.

An idea for a refinement of this solver class is to have a general interface for stochastic solvers (with inner/outer loop structure ) and the inner loop should be an abstract class, as well as equipping the class with a `run_iterator` for batching. I'll add this idea to Bence's issue #194.

src/nemos/solvers.py
BalzaniEdoardo: yes, I agree

BalzaniEdoardo: I added to the issue on the SVRG tutorial

src/nemos/solvers.py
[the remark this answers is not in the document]
BalzaniEdoardo: I figured that our links for refs did not work because the References is not considered a markdown section. I did make it a section and this seems to be rendered well. Do you like this solution?

src/nemos/solvers.py
```diff
@@ -30,15 +30,15 @@ class SVRGState(NamedTuple):
         Step size of the individual gradient steps.
     reference_point :
         Anchor/reference/snapshot point where the full gradient is calculated in the SVRG algorithm.
-        Corresponds to `x_{s}` in the pseudocode in [1]
+        Corresponds to `x_{s}` in the pseudocode[$^{[1]}$](#references).
```
billbrod, line 33:
```suggestion
        Corresponds to $x_{s}$ in the pseudocode[$^{[1]}$](#references).
```

billbrod (approved):
Two small issues and then this is ready to merge!

- Docstring of `ProxSVRG` mentions `hyperparams_prox`, which is no longer an argument. So that should be removed.

src/nemos/solvers.py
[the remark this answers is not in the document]
billbrod: I think this solution is fine for now, but we should figure out how to get them rendering correctly. Not as part of this PR, I opened an issue for it: #212

BalzaniEdoardo:
> * `hyperparams_prox`

The docstrings are ok, because that is referring to any hyperparameter that a general proximal operator gets as input. For example, what we call regularization_strength is one such hyper-parameter for ridge.

BalzaniEdoardo (approved):

Issue #178, opened 2024-07-08
pre-defined range/knots for basis


wulfdewolf:
**Is your feature request related to a problem? Please describe.**
Sometimes we want to be able to define a spline basis over a fixed range, instead of over the range of the given data.
For example, in [Hardcastle et al. (2017)](https://linkinghub.elsevier.com/retrieve/pii/S0896-6273(17)30237-4), fixed ranges of position, speed, head direction, and theta phase are imposed.
Ultimately, this allows for learning 'model-derived' tuning curves that are defined across the same range, as in some sessions, animals do not sample all bins.

**Describe the solution you'd like**
A simple solution would be to allow passing an absolute min and max when initialising the basis such that those are used to define the knots.
Another option is to simply allow passing the knots themselves.

**Describe alternatives you've considered**
Alternatives for me would be to integrate scikit-learn basis transformers into the nemos code, as they allow the passing knots.
They have no API for making multiplicative bases though, which I find very handy.

BalzaniEdoardo:
Thanks for the suggestion, we'll work on it!

Do you have suggestions on how to handle out-of-bounds observations if users passes the min-max? 

My inclination would be to return NaNs for out of bound values (instead of zero, which would create artifactual jumps in the firing rate).

wulfdewolf:
NaNs would indeed be good!
Typically, I would have bounded the data already (or it just is impossible for actual acquisition to go beyond the range).
So in practice, it should not happen often that these out of bounds occur but NaNs seem like the correct way to handle when they do.

BalzaniEdoardo:
Closed with PR #191

Issue #74, opened 2023-12-20
"How to choose basis" tutorial


billbrod:
We'll want to write a tutorial to guide people on how to choose their basis. This should describe a variety of basis properties that you might want and why, and which basis object has them. We could have a table listing all the basis objects and all the properties, with :heavy_check_mark: or :x: to show which basis has which property. Should also describe feature domains that a basis makes sense for.

Feature domains:
- Arbitrary visual input ("image-computable")
- Visual spatial frequency and orientation / Fourier decomposition
- Head position and other circular variables
- Spiking
- Position in 2d space

Basis properties:
- Orthogonality
- Uniform tiling
- Linear vs. log-scaled / foveated (e.g., raised cosine basis)
- Anti-aliasing / ability to interpolate

BalzaniEdoardo:
I add the "integrates to 1" property which only MSpline have

PR #191, opened 2024-07-16
Provide a range for basis

BalzaniEdoardo:
## Setting basis range
This PR addresses #178, allowing to set a fixed range for the basis. 

This is a useful when,
1. We want to discard outliers, for example position tracking artifacts or similar
2. We want to fit the same model with the same basis configuration over different runs of an experiment.

The vmin/vmax are set at initialization of the basis, so that multiple calls to `Basis.compute_features` will return same result every time and do not require extra arguments. 

vmin and vmax are get only properties.

### Behavior of (`__call__` and `compute_features`):
- **The default behavior is unchanged.** If vmin/vmax  are not provided, they are set to the nanmin/nanmax respectively when the samples are provided.
- If vmin/vmax are provided, the range starts/ends at vmin/vmax. Any sample out of range will return NaN.

## Behavior of`evaluate_on_grid` (differs since it does not receive any sample):
- if both vmin/vmax are provided, evaluates on an equispaced grid between vmin and vmax.
- If only vmin is provided,  evaluates on an equispaced grid between vmin and vmin + 1.
- If only vmax is provided, evaluates on an equispaced grid between vmax - 1 and vmax.
- If no value is provided, evaluates on an equispaced grid between 0 and 1.

```python
from nemos.basis import BSplineBasis
import numpy as np


x = np.arange(10)
basis = BSplineBasis(5, vmin=1, vmax=8)

# this uses the values passed at initializaiton vmin=1, vmax=8
y = basis.compute_features(x)
y = basis(x)

```


BalzaniEdoardo:
TODO

- [ ] Add Tests

src/nemos/basis.py
```diff
@@ -493,7 +566,7 @@ def _check_transform_input(
             # make sure array is at least 1d (so that we succeed when only
             # passed a scalar)
             xi = tuple(np.atleast_1d(np.asarray(x, dtype=float)) for x in xi)
-        except TypeError:
+        except (TypeError, ValueError):
```
BalzaniEdoardo, line 569: this was needed to surface exception obtained passing samples like `np.array(["a",  "b"])`

billbrod (changes requested):
Basic ideas here all look good to me, but I've got some higher level questions:

- Not sure how I feel about setting just one of vmin/vmax. Picking vmin+1 / vmax-1 feels very arbitrary. why not require both? we could do that by changing the argument to `vrange` and requiring it to be a 2-tuple
- this PR pulls the eval and conv modes of the basis further apart. do we want to consider making them separate classes? or something else, but it feels weird to have fully half of the arguments be ignored in one mode and the other half in the other mode.
- not necessarily part of this PR, but: I don't know how to pull up the docstring of `__call__`. what comes up is the docstring of the class. that seems ... like a problem. if I'm missing something, let me know, otherwise I can create an issue
- for the raised cosine basis objects, they only give reasonable values on the interval [0,1]. So vmin/vmax should lie within that interval as well.
- for those objects, it's weird that the log never rescales and the linear can optionally do so. shouldn't they both always rescale? did we discuss this at some point? the other basis objects never rescale, right? but it feels like rescaling means something different

docs/background/plot_01_1D_basis_function.py
```diff
@@ -52,8 +82,8 @@
 #
 # Let's see how this two modalities operate.
 
-eval_mode = nmo.basis.BSplineBasis(n_basis_funcs=n_basis, mode="eval")
-conv_mode = nmo.basis.BSplineBasis(n_basis_funcs=n_basis, mode="conv", window_size=100)
+eval_mode = nmo.basis.MSplineBasis(n_basis_funcs=n_basis, mode="eval")
```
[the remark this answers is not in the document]
BalzaniEdoardo: to switch it up, and not show a single case. before it was MSpline then BSpline for the same reason. Now I prefer BSpline for the first example, because the MSpline scales differently with the range.

Eventually, we will need a dedicated doc page for basis in which we show each one.

src/nemos/validation.py
```diff
@@ -275,3 +276,33 @@ def check_tree_structure(tree_1: Any, tree_2: Any, err_message: str):
     """
     if jax.tree_util.tree_structure(tree_1) != jax.tree_util.tree_structure(tree_2):
         raise TypeError(err_message)
+
+
+def check_fraction_valid_samples(*tree: Any, err_msg: str, warn_msg: str) -> None:
+    """
+    Check the fraction of valid entries.
```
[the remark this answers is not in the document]
BalzaniEdoardo: it checks for nans and infs specifically. Anything that can be converted to a finite/non nan float is valid

src/nemos/basis.py
```diff
@@ -1133,6 +1227,8 @@ def __call__(self, sample_pts: ArrayLike) -> FeatureMatrix:
             ],
             axis=1,
         )
+        # re-normalize so that it integrates to 1 over the range.
+        X /= scaling
```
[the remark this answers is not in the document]
BalzaniEdoardo: the docstrings yes, the docs also but in a PR in which we detail each baisis function.

src/nemos/basis.py
```diff
@@ -1263,12 +1371,20 @@ def __call__(self, sample_pts: ArrayLike) -> FeatureMatrix:
...
-        basis_eval = bspline(
-            sample_pts, knot_locs, order=self.order, der=0, outer_ok=False
-        )
+        valid = ~np.isnan(sample_pts)
+        if np.any(valid):
+            basis_eval = bspline(
+                sample_pts, knot_locs, order=self.order, der=0, outer_ok=False
+            )
+        else:
+            basis_eval = np.full((sample_pts.shape[0], self.n_basis_funcs), np.nan)
```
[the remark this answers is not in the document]
BalzaniEdoardo: yep, It wasn't the case at the beginning  but then I changed my mind and this is a leftover

src/nemos/basis.py
```diff
@@ -1511,18 +1670,22 @@ def _check_width(width: float) -> None:
...
         Parameters
         ----------
         sample_pts :
             Spacing for basis functions, holding elements on interval [0, 1], Shape (number of samples, ).
-
-        Returns
-        -------
-        basis_funcs :
-            Raised cosine basis functions, shape (n_samples, n_basis_funcs).
+        rescale_samples :
```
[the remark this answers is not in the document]
BalzaniEdoardo: removed, and added a  private flag at init that define if the RaisedCosineLinear should rescale or not (if called by the subclass it should not, otherwise it always have to)

BalzaniEdoardo:
> Basic ideas here all look good to me, but I've got some higher level questions:
> 
>     * Not sure how I feel about setting just one of vmin/vmax. Picking vmin+1 / vmax-1 feels very arbitrary. why not require both? we could do that by changing the argument to `vrange` and requiring it to be a 2-tuple
> 
This has been addressed (`bounds` is the name of the variable)

>     * this PR pulls the eval and conv modes of the basis further apart. do we want to consider making them separate classes? or something else, but it feels weird to have fully half of the arguments be ignored in one mode and the other half in the other mode.
> 
We should discuss this with Sofia and @gviejo. But definitely not something we should tackle in this PR.

>     * not necessarily part of this PR, but: I don't know how to pull up the docstring of `__call__`. what comes up is the docstring of the class. that seems ... like a problem. if I'm missing something, let me know, otherwise I can create an issue
> 
Interesting...

>     * for the raised cosine basis objects, they only give reasonable values on the interval [0,1]. So vmin/vmax should lie within that interval as well.
> 
This has been resolved. RaisedCosineLinear should always be called with the rescale_sample flag set to true. The Log instead must call the super class __call__ with the rescale_sample set to False.

>     * for those objects, it's weird that the log never rescales and the linear can optionally do so. shouldn't they both always rescale? did we discuss this at some point? the other basis objects never rescale, right? but it feels like rescaling means something different

They both have a single possible behavior, I was exposing the parameter by mistake.

src/nemos/basis.py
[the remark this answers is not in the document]
BalzaniEdoardo: we should take care of this when we take on this issue #74

src/nemos/basis.py
```diff
@@ -58,15 +60,52 @@ def wrapper(self: Basis, *xi: ArrayLike, **kwargs):
...
+    bounds: Optional[Tuple[float, float]] = None,
+) -> Tuple[NDArray, float]:
+    """Rescale samples to [0,1].
+
+    Parameters
+    ----------
+    sample_pts:
+        The original samples.
+    bounds:
+        Sample bounds. `bounds[0]` abd `bounds[1]` are mapped to 0/1 respectively.
```
billbrod, line 74:
```suggestion
        Sample bounds. `bounds[0]` and `bounds[1]` are mapped to 0 and 1, respectively.
```

src/nemos/basis.py
```diff
@@ -58,15 +60,52 @@ def wrapper(self: Basis, *xi: ArrayLike, **kwargs):
...
+    Warns
+    -----
+    UserWarning
+        If more than 90% of the sample points contain NaNs or Infs.
+    """
+    sample_pts = sample_pts.astype(float)
+    vmin = np.nanmin(sample_pts) if bounds is None else bounds[0]
+    vmax = np.nanmax(sample_pts) if bounds is None else bounds[1]
+    if vmin and vmax and vmax <= vmin:
+        raise ValueError("Invalid value range. `vmax` must be larger then `vmin`!")
```
billbrod, line 92:
```suggestion
    if vmin and vmax and vmax <= vmin:
        raise ValueError("Invalid value range. `bounds[1]` must be larger then `bounds[0]`!")
```

since name of args have changed

src/nemos/basis.py
```diff
@@ -205,13 +248,23 @@ def __init__(
...
         self._conv_args = args
         self._conv_kwargs = kwargs
+
+        if bounds is not None and len(bounds) != 2:
+            raise ValueError(
+                f"The provided `bounds` must be of length two. Length {len(bounds)} provided instead!"
+            )
+
+        # convert to float and store
+        self._bounds = bounds if bounds is None else tuple(map(float, bounds))
```
billbrod, line 266: presumably this raises an error if the value of bounds aren't numeric. is it a legible error? or should we raise an easy-to-understand one for users?

billbrod: and does all this validation logic belong in a setter?

src/nemos/basis.py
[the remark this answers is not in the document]
billbrod: still think this comment is needed

src/nemos/basis.py
```diff
@@ -990,21 +1079,11 @@ def _generate_knots(
         if is_cyclic:
             num_interior_knots += self.order - 1
 
-        assert 0 <= perc_low <= 1, "Specify `perc_low` as a float between 0 and 1."
-        assert 0 <= perc_high <= 1, "Specify `perc_high` as a float between 0 and 1."
-        assert perc_low < perc_high, "perc_low must be less than perc_high."
-
-        # clip to avoid numerical errors in case of percentile numerical precision close to 0 and 1
-        # Spline basis have support on the semi-open [a, b)  interval, we add a small epsilon
-        # to mx so that the so that basis_element(max(samples)) != 0
```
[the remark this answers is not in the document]
billbrod: still think it's necessary

src/nemos/basis.py
```diff
@@ -990,21 +1079,11 @@ def _generate_knots(
...
-        mn = np.nanpercentile(sample_pts, np.clip(perc_low * 100, 0, 100))
-        mx = np.nanpercentile(sample_pts, np.clip(perc_high * 100, 0, 100)) + 10**-8
-
         knot_locs = np.concatenate(
             (
-                mn * np.ones(self.order - 1),
-                np.linspace(mn, mx, num_interior_knots + 2),
-                mx * np.ones(self.order - 1),
+                np.zeros(self.order - 1),
+                np.linspace(0, (1 + 10**-8), num_interior_knots + 2),
```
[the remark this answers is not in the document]
billbrod: still curious

src/nemos/basis.py
```diff
@@ -1075,6 +1143,11 @@ class MSplineBasis(SplineBasis):
     ----------
     [1] Ramsay, J. O. (1988). Monotone regression splines in action. Statistical science,
         3(4), 425-441.
+
+    Notes
+    -----
+    MSplines must integrate to 1 over their domain. Therefore, if the domain (x-axis) of an MSpline
+    basis is dilated by a factor of $\alpha$, the co-domain (y-axis) values will shrink by a factor of $1/\alpha$.
```
billbrod, line 1150: should make it clear how this point maps to the input arguments. Something like: "For example, by default, with `bounds=(0,1)`, the maximum height is 1. However, if `bounds=(0,2)`, then the maximum height is .5". Or the like

src/nemos/basis.py
```diff
@@ -1471,14 +1573,24 @@ def __init__(
...
+            window_size=window_size,
+            bounds=bounds,
+            **kwargs,
         )
         self._n_input_dimensionality = 1
         self._check_width(width)
         self._width = width
+        # if linear raised cosine are initialized
+        # this flag is always true, the samples
+        # must be rescaled to 0 and 1.
```
billbrod, line 1592:
```suggestion
        # for these linear raised-cosine basis functions,
        # the samples must be rescaled to 0 and 1.
```

the "if" in previous wording is confusing

src/nemos/basis.py
```diff
@@ -1663,8 +1778,14 @@ def __init__(
             mode=mode,
             width=width,
             window_size=window_size,
+            bounds=bounds,
             **kwargs,
         )
+        # overwrite the flag for scaling the samples to [0,1] in the super.__call__(...).
+        # The samples are scaled appropriately in the self._transform_samples which scales
+        # and applies the log-stretch, no additional transform is needed.
```
billbrod, line 1786:
```suggestion
        # The samples are scaled appropriately in the self._transform_samples which scales
        # and applies the log-stretch, no additional transform is needed.
```

billbrod (approved):
I have a couple points dropped throughout that I think need addressing, basically juts by adding a comment, but then I think this is good to go. I think using a single `bounds` argument is much cleaner.

src/nemos/basis.py
BalzaniEdoardo: moved to a setter, exception surfaced

src/nemos/basis.py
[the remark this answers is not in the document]
BalzaniEdoardo: it doesn't really matter, it just need to be slightly larger than 1, the knots arrive to 1 by default after the normalization. any small epsilon after one would have the same effect. I'll set it to numerical precision of floats if that's more standard

src/nemos/basis.py
BalzaniEdoardo: it's the area under the curve (like the cumsum/dx) that changes

BalzaniEdoardo: I'll try to explain

PR #207, opened 2024-08-08
Automatic step sizes for SVRG

bagibence:
Attempt to automatically determine the batch- and step sizes for SVRG when fitting a GLM with Poisson observations and a softplus inverse link function.
Based on [this paper](https://papers.nips.cc/paper_files/paper/2019/hash/cd00692c3bfe59267d5ecfac5310286c-Abstract.html).


bagibence:
This needs to be more exhaustively tested to make sure it works on real datasets.
I have even encountered toy examples where my current implementation didn't work either.

bagibence:
The regularization strength might have to be added to the `L` and `L_max` constants determined.
It could also be useful as bound for the convexity when using ridge.

BalzaniEdoardo:
This PR provides the infrastructure for computing an optimal stepsize and batch_size  for SVRG based on the GLM configurations. 

The optimal hyperparameters depends on the loss function L-smoothness. This means that for each model configuration (observation noise, link function, regularization), one may need to to compute a different estimate of the smoothness parameters. 

Here, I implemented a look-up table that should be easy to extend whenever new estimates becomes available (for example if we derive the L-smoothness for Gamma + softplus observations).

tests/test_glm.py
```diff
@@ -1697,17 +1758,308 @@ def test_reg_set_params(self, params, warns):
...
+        opt_state = model.initialize_state(X, y, true_params)
+        solver = inspect.getclosurevars(model._solver_run).nonlocals["solver"]
+
+        if stepsize is not None:
+            assert opt_state.stepsize == stepsize
+            assert solver.stepsize == stepsize
+        else:
+            assert opt_state.stepsize > 0
+            assert isinstance(opt_state.stepsize, float)
+            model.fit(X, y)
```
sjvenditto, line 1825: I would consider removing `model.fit()` calls in this test (as well as in `test_glm_optimal_config_set_initial_state_pytree`, and both functions with the same names in the `TestPopulationGLM` class) since its not being consistently called across all cases, and since no checks are happening after the model is fit

tests/test_svrg_defaults.py
```diff
@@ -0,0 +1,264 @@
...
+            "batch_size" in result and result["batch_size"] > 0
+        ), "Batch size should be computed since it was not provided."
+    if expected_stepsize is not None:
+        assert (
+            result["stepsize"] == expected_stepsize
+        ), "Provided stepsize should be returned as-is."
+    else:
+        assert (
+            "stepsize" in result and result["stepsize"] > 0
+        ), "Stepsize should be computed since it was not provided."
```
sjvenditto, line 174: I don't know if I missed it in one of the previous tests, but I didn't notice any test that explicitly tests the result of a computed `stepsize`. They only check that `stepsize > 0`. Is it worth it to have a test check the result explicitly, similar to `test_calculate_optimal_batch_size_svrg_all_config` at the end?

src/nemos/glm.py
```diff
@@ -55,10 +56,35 @@ class GLM(BaseRegressor):
...
-    | GroupLasso    | ProximalGradient | ProximalGradient                                            |
+    | UnRegularized | GradientDescent  | GradientDescent, BFGS, LBFGS, NonlinearCG, SVRG, ProximalGradient, ProxSVRG |
+    | Ridge         | GradientDescent  | GradientDescent, BFGS, LBFGS, NonlinearCG, SVRG, ProximalGradient, ProxSVRG |
+    | Lasso         | ProximalGradient | ProximalGradient, ProxSVRG                                           |
+    | GroupLasso    | ProximalGradient | ProximalGradient, , ProxSVRG                                         |
+
+
+    **Fitting Large Models**
+
+    For very large models, you may consider using the Stochastic Variance Reduced Gradient
```
billbrod, line 67: would be nice to point to example in the docs here

src/nemos/glm.py
```diff
@@ -55,10 +56,35 @@ class GLM(BaseRegressor):
...
+    ([ProxSVRG](../solvers/_svrg/#nemos.solvers._svrg.ProxSVRG)) solver,
+    which take advantage of batched computation.
+
+    The performance of the SVRG solver depends critically on the choice of `batch_size` and `stepsize`
+    hyperparameters. These parameters control the size of the mini-batches used for gradient computations
+    and the step size for each iteration, respectively. Improper selection of these parameters can lead to slow
+    convergence or even divergence of the optimization process.
+
+    To assist with this, for certain GLM configurations, we provide recommended `batch_size` and `stepsize`
+    values that are theoretically guaranteed to ensure fast convergence.
```
billbrod, line 78: does the user have to do anything to use the recommended values?

Overlaps with #249 , but would be good to have an example for how to use SVRG and set these options

src/nemos/glm.py
```diff
@@ -1003,10 +1035,35 @@ class PopulationGLM(GLM):
...
-    | GroupLasso    | ProximalGradient | ProximalGradient                                            |
+    | UnRegularized | GradientDescent  | GradientDescent, BFGS, LBFGS, NonlinearCG, SVRG, ProximalGradient, ProxSVRG |
+    | Ridge         | GradientDescent  | GradientDescent, BFGS, LBFGS, NonlinearCG, SVRG, ProximalGradient, ProxSVRG |
+    | Lasso         | ProximalGradient | ProximalGradient, ProxSVRG                                           |
+    | GroupLasso    | ProximalGradient | ProximalGradient, ProxSVRG                                            |
+
+
+    **Fitting Large Models**
+
+    For very large models, you may consider using the Stochastic Variance Reduced Gradient
```
billbrod, line 1046: same point about example and doc link

src/nemos/solvers/_compute_defaults.py
```diff
@@ -0,0 +1,71 @@
...
+    optimal step size and batch size for variance reduced gradient (SVRG and
+    ProxSVRG) algorithms. If the model is configured with specific solver names,
+    the appropriate computation functions are returned. Additionally, it determines the
+    smoothness and strong convexity constants based on the model's observation and regularizer.
+
+    Parameters
+    ----------
+    model :
+        The generalized linear model object for which the optimal step size and batch
+        configuration need to be computed. The model should have attributes like
```
billbrod, line 34: "attributes like" or it needs to have those attributes?

billbrod: probably don't need that second sentence at all, the type-hinting above specifies it must be an instance of `GLM` or `PopulationGLM`

src/nemos/solvers/_svrg_defaults.py
```diff
@@ -0,0 +1,548 @@
...
+
+    @wraps(func)
+    def wrapper(*args, **kwargs):
+        args, kwargs = jax.tree_util.tree_map(float, (args, kwargs))
+        return func(*args, **kwargs)
+
+    return wrapper
+
+
+# not using the previous one to avoid calculating L and L_max twice
```
billbrod, line 39: what's this comment?

src/nemos/solvers/_svrg_defaults.py
```diff
@@ -0,0 +1,548 @@
...
+    n_power_iters :
+        Number of power iterations to use when finding the largest eigenvalue. If None,
+        the eigenvalue is calculated directly.
+
+    Returns
+    -------
+    :
+        Smoothness constant `L`.
+    """
+    if n_power_iters is None:
```
billbrod, line 300: do we want to raise a warning (or an exception) if it looks like this a bad idea because the matrices are too large?

src/nemos/solvers/_svrg_defaults.py
```diff
@@ -0,0 +1,548 @@
...
+        if jnp.allclose(v_prev, v):
+            break
+
+    # Final eigenvalue calculation
+    v /= jnp.linalg.norm(v)
+    return _glm_softplus_poisson_l_smooth_multiply(X, y, v, batch_size).dot(v)
+
+
+def _glm_softplus_poisson_l_smooth(
+    X: NDArray, y: NDArray, batch_size: int, n_power_iters: Optional[int] = None
```
billbrod, line 275: do we want this to default to using the power method (safer)?

src/nemos/solvers/_svrg_defaults.py
```diff
@@ -0,0 +1,548 @@
...
+    numerator = 0.5 * batch_size * (num_samples - 1)
+    denominator = (
+        3 * (num_samples - batch_size) * l_smooth_max
+        + num_samples * (batch_size - 1) * l_smooth
+    )
+    return numerator / denominator
+
+
+@_convert_to_float
+def _calculate_stepsize_saga(
```
billbrod, line 383: what's saga? I don't see this function used anywhere

tests/test_glm.py
billbrod: yes, I'm confused about this `model.fit` too -- is it supposed to be in the else block?

tests/test_svrg_defaults.py
```diff
@@ -0,0 +1,257 @@
...
+    assert isinstance(batch_size, int)
+
+
+@pytest.mark.parametrize(
+    "num_samples, l_smooth_max, l_smooth, expected_b_hat",
+    [
+        (100, 10.0, 2.0, 2.8697202),
+    ],
+)
+def test_calculate_b_hat(num_samples, l_smooth_max, l_smooth, expected_b_hat):
```
billbrod, line 117: why only a single set to test here?

tests/test_svrg_defaults.py
```diff
@@ -0,0 +1,257 @@
...
+    assert jnp.isclose(b_hat, expected_b_hat)
+
+
+@pytest.mark.parametrize(
+    "num_samples, l_smooth_max, l_smooth, strong_convexity, expected_b_tilde",
+    [
+        (100, 10.0, 2.0, 0.1, 3.4146341463414633),
+    ],
+)
+def test_calculate_b_tilde(
```
billbrod, line 129: why only a single set to test here?

billbrod (changes requested):
- It is hard for me to review this (not knowing the math) without an example. Do we want to add an example as part of this PR or do a separate one to add an example for "how to use SVRG"?
- We really do need to get the math for this up somewhere.
- same point as #252 where it looks like black was run on some test scripts for the first time
- The new test glm (and test population glm) functions are hard to parse. they look like they're doing a lot -- would it make sense to break them up?

src/nemos/solvers/_svrg_defaults.py
BalzaniEdoardo: what is the threshold for this? If the matrix is too big the user will get a runtime error, we can try catch that, and raise a more understandable RuntimeError, saying: "Computing the optimal stepsize resulted in a RuntimeError. This is likely because the model design matrix is too large...." and add a possible solution.

BalzaniEdoardo:
This is ready for another round, I should have addressed everything. The I added the pdf as an asset, just for us to have it in a place that is easy to find. However, I would not add it to the documentation or any public site yet, since it is very much a work in progress. I  want to close this PR quickly, and before the math is polished

BalzaniEdoardo (commented):
@billbrod some of the things I resolved add pending comments that I did not submit. I added the SVRG example to one of the issue in the docs project. The rest should be addressed. If you feel like the SVRG description in the GLM class can still be improved, go ahead and change it, I think that's the upper-bound of my English writing skills :)

src/nemos/solvers/_svrg_defaults.py
```diff
@@ -298,9 +298,21 @@ def _glm_softplus_poisson_l_smooth(
...
     """
     if n_power_iters is None:
-        # Calculate the Hessian directly and find the largest eigenvalue
-        XDX = X.T.dot((0.17 * y.reshape(y.shape[0], 1) + 0.25) * X) / y.shape[0]
-        return jnp.sort(jnp.linalg.eigvalsh(XDX))[-1]
+        try:
+            warnings.warn(
+                "Direct computation of the eigenvalues requires a substantial amount of resources. "
+                "Please, consider using the power method by setting the `n_power_iters` parameter "
+                "(default behavior).",
```
billbrod, line 305:
```suggestion
                "Please, consider using the power method by setting the `n_power_iters` parameter ",
```

tests/test_svrg_defaults.py
```diff
@@ -204,9 +194,17 @@ def test_warnigns_svrg_optimal_batch_and_stepsize(
...
 
 
-@pytest.mark.parametrize("n_power_iter", [None, 1, 10])
-def test_glm_softplus_poisson_l_smooth_power_iter(x_sample, y_sample, n_power_iter):
-    _svrg_defaults._glm_softplus_poisson_l_smooth(x_sample, y_sample, n_power_iter)
+@pytest.mark.parametrize(
+    "n_power_iter, expectation",
+    [
+        (None, pytest.warns(UserWarning, match="Direct computation of the eigenvalues")),
+        (1, does_not_raise()),
```
billbrod, line 201: what's the behavior with `n_power_iter=0`

billbrod (approved):
This is good to go once we add the RuntimeError for the eigenvalue to the docstrings.

I think the glm docstring is clear enough for now -- once we add the relevant doc, I think we can basically remove that info (moving it to the tutorial) from the docstring and point to the tutorial

PR #211, opened 2024-08-14
Update python versions

BalzaniEdoardo:
Updated minimal requirement to python>=3.10, tested 3.12.


BalzaniEdoardo:
what do you guys think? should we support only 3.10, 3.11, 3.12 or should we keep supporting 3.9?

billbrod:
For plenoptic, I switched to only 3.10 - 3.12, as part of my review with pyopensci. Their advice was to follow [SPEC0](https://scientific-python.org/specs/spec-0000/), which basically has you support any version of python for 3 years after release. I think that's reasonable.

PR #216, opened 2024-08-27
Use pooch

BalzaniEdoardo:
Start transition to pooch for fetching data.

The initial intent was to get rid completely of the example_utils in the docs, so that each notebook could be downloaded and run directly. The first step to achieve this is to move the code for fetching the data in src/nemos, and use `pooch` to download from osf.

There is one tutorial that relies on dandi, and that's a big dependency that I would rather not have as main dependency of nemos. 

What should we do for that dataset?


billbrod (commented):
In general, this looks good. Thoughts:

- Why not just put `download_dandi_data` into our fetch module? currently, we're only using it once, but it's feasible we'd use it more
- I don't love using pooch to grab code -- pooch feels more like it's meant for binary files, and just dropping random scripts into the users working directory feels fragile. I'm not sure of a better solution right now, but I posted in the US-RSE slack to see if anyone has suggestions
- If we *do* end up going with pooch to get the utility scripts, I think we shouldn't use `if not path.exists`, but a try/except on the import.

BalzaniEdoardo:
> In general, this looks good. Thoughts:
> 
>     * Why not just put `download_dandi_data` into our fetch module? currently, we're only using it once, but it's feasible we'd use it more
I thought about it, it's a package with a lot of dependencies, and I was hesitant, moving it into the main package would make it a requirement. pynapple has it has an optional dependency, @gviejo what do you think about making dandi a dependency?  
My personal inclination is not to add it as a requirement for downloading a single dataset.

>     * I don't love using pooch to grab code -- pooch feels more like it's meant for binary files, and just dropping random scripts into the users working directory feels fragile. I'm not sure of a better solution right now, but I posted in the US-RSE slack to see if anyone has suggestions

I agree with you, I did not like people having to manually download the scripts, and place it in the correct folder structure but I did not find a clear optimal solution. This one at least is safer because of the hashing and the download of the script based on the package version.

>     * If we _do_ end up going with pooch to get the utility scripts, I think we shouldn't use `if not path.exists`, but a try/except on the import.

Ok! let's wait until we get an opinion from the US-RSE people and finish this off.

billbrod:
Just pushed some updates here, following conversation on US-RSE slack, which we should review:
- Pooch is now an optional dependency, which changes a bit of how `fetch_data.py` is structured. If the user tries to call `fetch_data` without pooch installed, they'll get an error telling them to install it.
- moves `download_dandi_data` into `fetch_data.py`, which behaves similarly to `fetch_data`, but with dandi as the optional dependency instead of pooch
- big restructure: makes the functions contained in the different `*_utils/plotting.py` scripts part of a hidden module, `_documentation_utils`. This is not imported in `nemos/__init__.py`  and so needs to be specifically imported: `from nemos import _documentation_utils`. This means it can contain `import matplotlib` (also an optional dependency), though we make sure to raise a more informative error message if the user manages to call it without it installed.

I think this ends up being cleaner, both for us and for users.

Those plotting functions in `_documentation_utils` should probably all have full docstrings now though, which should, among other things say `WARNING: this function is intended for use in the documentation, and is not generalizable. It may not work well with other data` or something similar

src/nemos/_documentation_utils/plotting.py
```diff
@@ -3,33 +3,186 @@
...
+
+try:
+    import matplotlib as mpl
+except ImportError:
+    raise ImportError(
+        "Missing optional dependency 'matplotlib'."
+        " Please use pip or "
+        "conda to install 'matplotlib'."
+    )
 import matplotlib.pyplot as plt
```
billbrod, line 15:
```suggestion
import matplotlib.pyplot as plt
import warnings
warnings.warn("plotting functions contained within `_documentation_utils` are intended for nemos's documentation. Feel free to use them, but they will probably not work as intended with other datasets / in other contexts.")
```

src/nemos/fetch/fetch_data.py
```diff
@@ -130,48 +85,10 @@ def _create_retriever(path: Optional[pathlib.Path] = None) -> pooch.Pooch:
         registry=REGISTRY_DATA,
         retry_if_failed=2,
         allow_updates="POOCH_ALLOW_UPDATES",
-        env="NEMOS_DATA_DIR",
+        env=_NEMOS_ENV,
```
billbrod, line 88: do you have a use case in mind for this?

billbrod (approved):
Generally looks good to me. It would be nice to have the reminder about the additional requirements at the top of each file, rather than the readme for the section. I'm not sure there's a way to do that cleanly (by editing the template), so maybe it's okay for now.

src/nemos/fetch/fetch_data.py
BalzaniEdoardo: I do use it to have the data in my repo docs/data directory instead of a cache directory harder to locate.

Issue #229, opened 2024-09-19
Deploying docs requires ffmpeg installation


sjvenditto:
The head-direction example requires the ffmpeg to generate the animation. If ffmpeg is installed, deploying the docs after installing nemos with the "docs" dependencies works as-is.  If ffmpeg is not installed, it fails with the error:
```
ERROR   -  Here is a summary of the problems encountered when running the examples

           Unexpected failing examples:
           /Users/svenditto/Documents/nemos/docs/tutorials/plot_02_head_direction.py failed leaving traceback:
           Traceback (most recent call last):
             File "/Users/svenditto/Documents/nemos/docs/tutorials/plot_02_head_direction.py", line 182, in
           <module>
               doc_plots.run_animation(neuron_count, epoch_one_spk.start[0])
             File "/Users/svenditto/Documents/nemos/src/nemos/_documentation_utils/plotting.py", line 506, in
           run_animation
               return HTML(anim.to_html5_video())
                           ^^^^^^^^^^^^^^^^^^^^^
             File "/opt/miniconda3/envs/nemos-test/lib/python3.12/site-packages/matplotlib/animation.py", line
           1265, in to_html5_video
               Writer = writers[mpl.rcParams['animation.writer']]
                        ~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
             File "/opt/miniconda3/envs/nemos-test/lib/python3.12/site-packages/matplotlib/animation.py", line
           128, in __getitem__
               raise RuntimeError(f"Requested MovieWriter ({name}) not available")
           RuntimeError: Requested MovieWriter (ffmpeg) not available


           -------------------------------------------------------------------------------

Aborted with a BuildError!
```
If there isn't a global ffmpeg installation and if using a conda environment, installing it via conda before pip-installing nemos will also work.

billbrod:
Agreed on avoiding ffmpeg if possible. When you say turn the animation into a gif, what are you thinking?

A possibility is to set `plt.rcParams['animation.html'] = 'jshtml'`, which should work in the browser without anything additional required). 

[...]

PR #230, opened 2024-09-24
set up javascript animation

BalzaniEdoardo:
This solves #229 by setting up jshtml for the animation output, which will not require ffmpeg to run nor Pillow as an example dependency.

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -11,9 +11,6 @@ build:
   tools:
     python: "3.10"
 
-  apt_packages:
-    - ffmpeg
-
 mkdocs:
   configuration: mkdocs.yml
 
--- a/docs/tutorials/plot_02_head_direction.py
+++ b/docs/tutorials/plot_02_head_direction.py
@@ -20,9 +20,6 @@ import nemos as nmo
 # some helper plotting functions
 from nemos import _documentation_utils as doc_plots
 
-# set the output format for animations in Jupyter notebooks to use JavaScript
-plt.rcParams['animation.html'] = 'jshtml'
-
 # configure pynapple to ignore conversion warning
 nap.nap_config.suppress_conversion_warnings = True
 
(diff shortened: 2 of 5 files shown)


billbrod (changes requested):
Two notes:
- We could move this configuration into the `nemos/styles/nemos.mplstyle` file so that it's easy to apply this configuration to all our notebooks (in case we end up adding additional animations).
- We should remove the `apt_packages: ffmpeg` lines from `.readthedocs.yaml`, since we're no longer using it. I'm slightly worried that we're actually still using the ffmpeg writer (if I recall correctly, the animation looks different when using jshtml), and removing the ffmpeg install will check this for us

sjvenditto:
```
ERROR   -  Here is a summary of the problems encountered when running the examples

           Unexpected failing examples:
           /mnt/home/svenditto/GitHub/nemos/docs/tutorials/plot_02_head_direction.py failed leaving traceback:
           Traceback (most recent call last):
             File "/mnt/home/svenditto/GitHub/nemos/docs/tutorials/plot_02_head_direction.py", line 185, in <module>
               doc_plots.run_animation(neuron_count, epoch_one_spk.start[0])
             File "/mnt/home/svenditto/GitHub/nemos/src/nemos/_documentation_utils/plotting.py", line 506, in run_animation
               return HTML(anim.to_html5_video())
             File "/mnt/home/svenditto/python_venvs/nemos/lib/python3.10/site-packages/matplotlib/animation.py", line 1265, in to_html5_video
               Writer = writers[mpl.rcParams['animation.writer']]
             File "/mnt/home/svenditto/python_venvs/nemos/lib/python3.10/site-packages/matplotlib/animation.py", line 128, in __getitem__
               raise RuntimeError(f"Requested MovieWriter ({name}) not available")
           RuntimeError: Requested MovieWriter (ffmpeg) not available
```
Still requiring ffmpeg on a fresh build

billbrod:
Oh I think this may be fault (since I think I wrote the animation code in `_documentation_utils/plotting.py`. Hold up, let me see if I can fix it.

billbrod:
@sjvenditto try it now

sjvenditto:
This works now!

PR #235, opened 2024-10-02
Fix transformer

gviejo:
This PR adress issues with Transformers. When creating a pipeline, the number of basis should be equal to the number of dimensions for the feature. 
If the number of columns for the feature is very high, it become tedious to add basis by hand. I added multiplication by integers for the class Basis.

There was also an issue where the class AdditiveBasis was trying to get the attributes `basis1` and `basis2` but they were only available as `_basis1` and `_basis2`. As a quick hack, i added two properties for AdditiveBasis that return basis 1 and basis 2.

Extra: I also added the method `__len__` for Basis where it returns basis.n_basis_funcs. It's more convenient.


BalzaniEdoardo:
Thanks @gviejo, I incorporated the change on the basis attribute in the #276 PR. 
I will add the `__len__` method too but I am closing this since I made a lot of changes in the `Basis` api and included the fixes for the `TransformerBasis`.

Thanks starting off this process.

Issue #226, opened 2024-09-13
Check links automatically


BalzaniEdoardo:
**Is your feature request related to a problem? Please describe.**
Some links (especially relative links) break as we refactor the docs. we should have a way to automatize the checks

**Describe the solution you'd like**
For external links, the [urlchecker action](https://github.com/marketplace/actions/urlchecker-action) will do it. For internal links, mkdocs raises warnings, so we need to check them / turn them into errors.

PR #238, opened 2024-10-03
prevent absolute links to docs (except badge)

BalzaniEdoardo:
Added automation to the CI that checks for no absolute links to NeMoS docs. 

The problem that absolute links may cause is described here #226. 

Quoting
> All links to NeMoS docs pages **must be relative**.  If absolute links to docs pages are added, structural changes in the HTML may break the absolute links, but this break won't be detected the first time the CI is run (this is because the htmlproofer would check the absolute link that will point to an older version of the website - "https://nemos.readthedocs.io/stable" or ".../latest" -, instead of the current built). 

> [!NOTE] 
> The `CONTRIBUTING.md` contains a note that clarifies to use only relative links already. Look at the end of the file.


billbrod (commented):
I think this is fine.
- Note that if you run `chmod +x prevent_absolute_links_to_docs.sh` locally and then commit those changes, you won't need to run it in the CI (git tracks permissions as well).
- I would move this script from the root directory. Maybe to a new folder called `scripts/` or include it in the `.github` directory? so as not to clutter up the root (I'm pretty sure that you'll have to change the path it's called on then).
- maybe replace `grep -r nemos.readthedocs.io` with `grep -r -E https?://nemos.*`, since we're going to change the domain name soon. (the `-E` allows you to use extended regex (not sure how that differs from regular regex, it's what I use), and `s?` means it will match `http://nemos` or https://nemos`).
- I'd probably add an explanation in `CONTRIBUTING` as to what relative vs. absolute means. I'm not sure it's obvious to the average scientist. E.g., "links must be relative, not absolute. That is, they should be of the form `./path/to/page` rather than `https://nemos.readthedocs.io/path/to/page`." or something like that

BalzaniEdoardo:
> I think this is fine.
> 
>     * Note that if you run `chmod +x prevent_absolute_links_to_docs.sh` locally and then commit those changes, you won't need to run it in the CI (git tracks permissions as well).
that's cool

>     * I would move this script from the root directory. Maybe to a new folder called `scripts/` or include it in the `.github` directory? so as not to clutter up the root (I'm pretty sure that you'll have to change the path it's called on then).
ok!

>     * maybe replace `grep -r nemos.readthedocs.io` with `grep -r -E https?://nemos.*`, since we're going to change the domain name soon. (the `-E` allows you to use extended regex (not sure how that differs from regular regex, it's what I use), and `s?` means it will match `http://nemos` or [https://nemos`](https://nemos%60)).
nice regex trick

>     * I'd probably add an explanation in `CONTRIBUTING` as to what relative vs. absolute means. I'm not sure it's obvious to the average scientist. E.g., "links must be relative, not absolute. That is, they should be of the form `./path/to/page` rather than `https://nemos.readthedocs.io/path/to/page`." or something like that
will do

BalzaniEdoardo:
After a second thought, I think it is safer to keep the explicit permission setting in the ci, and the execution time is basically minimal

Issue #202, opened 2024-07-26
Should we separate basis objects' eval and conv mode?


billbrod:
With the changes in #191 , the eval and conv modes of the basis object are becoming more and more separate, such that there a handful of initialization arguments that are ignored depending on which one you choose. This feels weird to me -- should we have them be separate objects somehow? Or initialize with shared arguments and then call a `eval()` or `conv()` method to set the mode, passing the relevant arguments then?

BalzaniEdoardo:
the power of the current approach is that you can combine and call the same method "compute_features", which will create any combination of convolution or evaluation for you once the parameters are set at initialization.

To achieve the behavior, we need to set configurations that are not passed before the actual transformation takes place. I don't like the idea that a basis is initialized in a default mode and then one needs to call an extra method to finish off the initialization to switch mode and pass hyperparameters.

Having separate classes may be an option, but we have already a number of basis (possibly expanding), i think it would be a lot of api to remember.

The solution we have now is not super elegant, but keeps the api to a minimum, open to suggestion on how to deal with this init more elegantly!

BalzaniEdoardo:
Revisiting my opinion: 
After adding a utility function for computing slices that applies to the feature axis, I needed to add an extra parameter that is required only for basis in `mode="conv"`, this is an argument to keep the classes separated, I will take care of that in a dedicated PR.

BalzaniEdoardo:
After the PR #247, basis will naturally have an equivalent of "fit" (setting the input shapes, and the conv kernels) and. a `transform` method (creating the features, the private `_compute_features`), and a "fit_transform" (`compute_features`) method.

I would make this more explicit when I'll split the PR, and this will make the map to TransformerBasis even cleaner

PR #247, opened 2024-10-09
Start facilitating coefficient splitting

BalzaniEdoardo:
## Summary

This PR starts the process of facilitating slicing the feature axis of either the model coefficients, or of the design matrix.

In particular,

### Added To `Basis`

- **`label` parameter:** name the variable, write only. Label of `AdditiveBasis` and `MultiplicativeBasis` combines the labels of 1d basis. 
- **`n_basis_input` parameter:** name the variable, write only, the shape of axis 1 of a 2D input to basis. This happens in convolve, for example when we convolve the counts of a neural population.
- **_get_feature_slicing:** a recursion that returns a dictionary of slices, that can be applied to the feature axis. Labels (or combination of labels) are used as keys.
- **split_by_feature:** the user facing method that splits an array into feature components.
- **n_output_features:** a read-only property that returns the number of output feature dimension.
- **_input_shape:** stores the shape of the first input passed to compute features. Any subsequent input shape will be checked against this. Consistent input shape will guaranteed that we can split correctly a feature axis.

`_get_feature_slicing` methods are for internal use, but will make our life very easy. In  `split_feature_axis`  we could use a  `jax.tree_utils.tree_map` to apply the slicing to any array, and get automatically a dictionary, well labeled in a meaningful way, containing the coefficients. 

If arrays are numpy, the use of slice is very efficient, since it will create a dict of views of the array.

It will also facilitate creating a "FeaturePytree" from an input in array or TsdFrame form.

## Example of `_get_feature_slicing`:

```python
>>> import nemos as nmo
>>> import jax

>>> bas1 = nmo.basis.RaisedCosineBasisLinear(3, mode="conv", n_basis_input=2, window_size=5, label="position")

>>> bas2 = nmo.basis.MSplineBasis(4, mode="conv", n_basis_input=3, window_size=5,  label="velocity")

>>> bas3 = bas1 + bas2 + bas1 * bas2

>>> # slice each individual input, default behavior.
>>> slice_dict = bas3._get_feature_slicing()[0]  
>>> slice_dict
{'position': {'0': slice(0, 3, None), '1': slice(3, 6, None)},
 'velocity': {'0': slice(6, 10, None),
  '1': slice(10, 14, None),
  '2': slice(14, 18, None)},
 '(position * velocity)': slice(18, 90, None)}

>>> # slice each additive component instead.
>>> bas3._get_feature_slicing(split_by_input=False)[0] 
{'position': slice(0, 6, None),
 'velocity': slice(6, 18, None),
 '(position * velocity)': slice(18, 90, None)}

>>> # splitting a design matrix becomes trivial
>>> x1 = np.random.normal(size=(10, 2))
>>> x2 = np.random.normal(size=(10, 3))
>>> X = bas3.compute_features(x1, x2, x1, x2)
>>> splits = jax.tree_util.tree_map(lambda sl: X[:, sl], slice_dict)

```

## Example of `split_by_feature`

```python
>>> import numpy as np
>>> from nemos.basis import BSplineBasis
>>> from nemos.glm import GLM

>>> # Define an additive basis
>>> basis = (
...     BSplineBasis(n_basis_funcs=5, mode="conv", window_size=10, label="feature_1") +
...     BSplineBasis(n_basis_funcs=6, mode="conv", window_size=10, label="feature_2")
... )

>>> # split an arbitrarily shaped array
>>> basis.compute_features(np.random.randn(10,2), np.random.randn(10,3))
>>> array = np.ones((1, 1, basis.num_output_features, 1))
>>> basis.split_by_feature(array, axis=2)
{'feature_1': array([[[[[1.],
           [1.],
           [1.],
           [1.],
           [1.]],
 
          [[1.],
           [1.],
           [1.],
           [1.],
           [1.]]]]]),
 'feature_2': array([[[[[1.],
           [1.],
           [1.],
           [1.],
           [1.],
           [1.]],
 
          [[1.],
           [1.],
           [1.],
           [1.],
           [1.],
           [1.]],
 
          [[1.],
           [1.],
           [1.],
           [1.],
           [1.],
           [1.]]]]])}


>>> # example of usage in combination with GLM
>>> X = np.random.normal(size=(20, basis.num_output_features))
>>> y = np.random.poisson(size=(20, ))
>>> basis.split_by_feature(GLM().fit(X, y).coef_, axis=0)
{'feature_1': Array([-0.02247754,  0.49239248, -0.09706223, -0.30416837,  0.04843776],      dtype=float32),
 'feature_2': Array([-0.29889402, -0.0040512 , -0.28740323,  0.5222396 ,  0.55201346,
        -0.13157026], dtype=float32)}
```

### Moved  To `nemos.identifiability_constraint`

- **apply_identifiability_constraints:** function that receives a matrix as input and drops columns until a minimal set of linearly independent columns is left.
- **apply_identifiability_constraints_by_basis_component:** a method that assumes that each feature component is independent, and `apply_identifiability_constraints` to each feature. This doesn't guarantee full rank but can be way faster, especially when the number of features is large, and can be applied in sequence: first a pass by component, then a full pass to the design matrix. It has the advantage of being more stable numerically, since the dimensionality of each component can be much smaller than the overall design matrix dimensionality, making the singular value computation and thresholding easier.

> [NOTE] 
> The use of parentheses guarantees that the `label` fully specifies the order of operation in a composite basis.


BalzaniEdoardo:
Note to myself, if the number of output features is pre-computed, we avoid a nested recursion,  and we can get rid of the method for computing the number of outputs.

src/nemos/basis.py
```diff
@@ -659,7 +692,9 @@ def _compute_features(self, *xi: ArrayLike) -> FeatureMatrix:
         Raises
         ------
         ValueError:
-            If an invalid mode is specified or necessary parameters for the chosen mode are missing.
+            - If an invalid mode is specified or necessary parameters for the chosen mode are missing.
+            - In mode "conv", if the number of inputs to be convolved, doesn't match the number of inputs
```
BalzaniEdoardo, line 698: The init of `mode="conv"` and `mode="eval"` starts to diverge  even more, and this is another argument in favor of having distinct classes for basis that performs convolution and basis for evaluation. 

Separate PR I would say, since there is a lot to change if we include the test. Do you have suggestion on an API that may work and also naming conventions?

Most of the machinery is shared, so probably a base class which they both inherit from, with an abstract method `compute_features`, and the two variants simply re-implement that

BalzaniEdoardo: Refer to #202 for a discussion on the basis API

src/nemos/identifiability_constraints.py
```diff
@@ -0,0 +1,372 @@
...
+    For GLMs, this rank deficiency means that different sets of coefficients might yield
+    identical predicted rates and log-likelihood, complicating parameter learning, especially
+    in the absence of regularization.
+
+    For very large feature matrices generated by a sum of low-dimensional basis components, consider
+    `apply_identifiability_constraints_by_basis_component`.
+
+    Parameters
+    ----------
+    feature_matrix:
```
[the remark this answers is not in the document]
billbrod: I think not in the docstring, just in the signature. (if we have them in both places it's hard to keep them synced)

src/nemos/identifiability_constraints.py
```diff
@@ -0,0 +1,372 @@
...
+        Indices of the columns that are kept. This should be used for applying the same transformation
+        to a feature matrix generated from different a set of inputs (as for a test set).
+
+    Examples
+    --------
+    >>> import numpy as np
+    >>> from nemos.identifiability_constraints import apply_identifiability_constraints_by_basis_component
+    >>> from nemos.basis import BSplineBasis
+    >>> from nemos.glm import GLM
+
```
billbrod, line 303: same point about blank lines

tests/test_basis.py
```diff
@@ -100,6 +100,107 @@ def test_compute_features_input(self, eval_input):
...
+        [
+            ((20,), 1, does_not_raise()),
+            ((20, 1), 1, does_not_raise()),
+            ((20, 2), 2, does_not_raise()),
+            ((20, 1, 2), 2, does_not_raise()),
+            ((20, 2, 1), 2, does_not_raise()),
+            ((20, 2, 2), 4, does_not_raise()),
+        ],
+    )
+    def test_compute_features_conv_input(
```
billbrod, line 182: if you're never going to raise an error, why use expectation?

tests/test_basis.py
```diff
@@ -661,20 +822,20 @@ def test_convolution_is_performed(self):
...
-            bas.identifiability_constraints = "True"
-
-    def test_identifiability_constraint_apply(self):
-        bas = self.cls(5, mode="conv", window_size=10)
-        bas.identifiability_constraints = True
-        # identifiability constraint should mean center only for this basis
-        X = bas(np.linspace(0, 1, 20))
-        assert np.allclose(X.mean(axis=0), np.zeros(X.shape[1]))
-        assert X.shape[1] == bas.n_basis_funcs
+    # def test_identifiability_constraint_setter(self):
```
billbrod, line 825: either delete or uncomment

tests/test_basis.py
```diff
@@ -821,6 +988,96 @@ def test_compute_features_input(self, eval_input):
...
+        [
+            ((20,), 1, does_not_raise()),
+            ((20, 1), 1, does_not_raise()),
+            ((20, 2), 2, does_not_raise()),
+            ((20, 1, 2), 2, does_not_raise()),
+            ((20, 2, 1), 2, does_not_raise()),
+            ((20, 2, 2), 4, does_not_raise()),
+        ],
+    )
+    def test_compute_features_conv_input(
```
billbrod, line 1068: same point about expectation

tests/test_basis.py
```diff
@@ -1337,19 +1646,19 @@ def test_convolution_is_performed(self):
...
-        with pytest.raises(TypeError, match="`identifiability_constraints` must be"):
-            bas.identifiability_constraints = "True"
-
-    def test_identifiability_constraint_apply(self):
-        bas = self.cls(5)
-        bas.identifiability_constraints = True
-        X = bas(np.linspace(0, 1, 20))
-        assert np.allclose(X.mean(axis=0), np.zeros(X.shape[1]))
-        assert X.shape[1] == bas.n_basis_funcs - 1
+    # def test_identifiability_constraint_setter(self):
```
billbrod, line 1649: either uncomment or delete

tests/test_basis.py
```diff
@@ -2012,19 +2481,19 @@ def test_convolution_is_performed(self):
...
```
billbrod, line 2484: either delete or uncomment

tests/test_basis.py
```diff
@@ -2804,19 +3392,19 @@ def test_convolution_is_performed(self):
...
-        with pytest.raises(TypeError, match="`identifiability_constraints` must be"):
-            bas.identifiability_constraints = "True"
-
-    def test_identifiability_constraint_apply(self):
-        bas = self.cls(5, decay_rates=[1, 2, 3, 4, 5])
-        bas.identifiability_constraints = True
-        X = bas(np.linspace(0, 1, 20))
-        assert np.allclose(X.mean(axis=0), np.zeros(X.shape[1]))
-        assert X.shape[1] == bas.n_basis_funcs
+    # def test_identifiability_constraint_setter(self):
```
billbrod, line 3395: either delete or uncomment

tests/test_basis.py
```diff
@@ -2862,6 +3450,68 @@ def test_compute_features_input(self, eval_input):
...
+        [
+            ((20,), 1, does_not_raise()),
+            ((20, 1), 1, does_not_raise()),
+            ((20, 2), 2, does_not_raise()),
+            ((20, 1, 2), 2, does_not_raise()),
+            ((20, 2, 1), 2, does_not_raise()),
+            ((20, 2, 2), 4, does_not_raise()),
+        ],
+    )
+    def test_compute_features_conv_input(
```
billbrod, line 3495: same point about expectation

tests/test_basis.py
```diff
@@ -3400,19 +4126,19 @@ def test_convolution_is_performed(self):
...
-        with pytest.raises(TypeError, match="`identifiability_constraints` must be"):
-            bas.identifiability_constraints = "True"
-
-    def test_identifiability_constraint_apply(self):
-        bas = self.cls(5)
-        bas.identifiability_constraints = True
-        X = bas(np.linspace(0, 1, 20))
-        assert np.allclose(X.mean(axis=0), np.zeros(X.shape[1]))
-        assert X.shape[1] == bas.n_basis_funcs - 1
+    # def test_identifiability_constraint_setter(self):
```
billbrod, line 4129: either delete or uncomment

tests/test_basis.py
```diff
@@ -3559,6 +4291,68 @@ def test_compute_features_input(self, eval_input):
...
+        "input_shape, expected_n_input, expectation",
+        [
+            ((20,), 1, does_not_raise()),
+            ((20, 1), 1, does_not_raise()),
+            ((20, 2), 2, does_not_raise()),
+            ((20, 1, 2), 2, does_not_raise()),
+            ((20, 2, 1), 2, does_not_raise()),
+            ((20, 2, 2), 4, does_not_raise()),
+        ],
+    )
```
billbrod, line 4335: same point about expectation

tests/test_basis.py
```diff
@@ -4059,19 +4877,19 @@ def test_convolution_is_performed(self):
...
-        with pytest.raises(TypeError, match="`identifiability_constraints` must be"):
-            bas.identifiability_constraints = "True"
-
-    def test_identifiability_constraint_apply(self):
-        bas = self.cls(5)
-        bas.identifiability_constraints = True
-        X = bas(np.linspace(0, 1, 20))
-        assert np.allclose(X.mean(axis=0), np.zeros(X.shape[1]))
-        assert X.shape[1] == bas.n_basis_funcs - 1
+    # def test_identifiability_constraint_setter(self):
```
billbrod, line 4880: either delete or uncomment

docs/how_to_guide/plot_05_batch_glm.py
```diff
@@ -54,6 +54,7 @@
 #
 # Here we instantiate the basis. `ws` is 40 time bins. It corresponds to a 200 ms windows
 ws = 40
+
```
billbrod, line 57: ??

billbrod (changes requested):
I'm confused by the behavior of `split_by_feature`:
- Why does it return 3d arrays? If I look at the example in the docstring, it returns arrays of shape (20,1,5) and (20,1,6). What's the middle dimension? This is also the behavior for a single basis or multiplicative basis.
- This method only really make sense for the additive basis right? Otherwise it doesn't do anything but add the middle dimension? In that case, should it exist for the other bases objects? Or am I missing something?

I also don't think the tutorial we have on identifiability constraints right now is sufficient. Might not need to be addressed here, but should be added to the docs project if not. Should explain why it's bad to be rank-deficient, show what you gain from being full rank, etc.

src/nemos/basis.py
```diff
@@ -543,10 +537,12 @@ def __init__(
         mode: Literal["eval", "conv"] = "eval",
         window_size: Optional[int] = None,
         bounds: Optional[Tuple[float, float]] = None,
+        label: Optional[str] = None,
```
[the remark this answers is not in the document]
BalzaniEdoardo: I like this idea!

src/nemos/identifiability_constraints.py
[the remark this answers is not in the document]
BalzaniEdoardo: I agree, let's keep it in one place only

src/nemos/basis.py
```diff
@@ -93,10 +88,6 @@ def min_max_rescale_samples(
     sample_pts = sample_pts.astype(float)
     vmin = np.nanmin(sample_pts) if bounds is None else bounds[0]
     vmax = np.nanmax(sample_pts) if bounds is None else bounds[1]
-    if vmin and vmax and vmax <= vmin:
```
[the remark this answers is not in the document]
BalzaniEdoardo: not necessary, bounds are checked in the property setter and, when not None, it must hold that bounds[1] > bounds[0]. This implies that vmin <= vmax.

src/nemos/basis.py
```diff
@@ -611,6 +615,28 @@ def _check_convolution_kwargs(self):
                 f"Allowed convolution keyword arguments are: {convolve_configs}."
             )
 
+    @property
+    def n_output_features(self) -> int | None:
+        """Read-only property returning the number of features returned by the basis."""
```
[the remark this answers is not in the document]
BalzaniEdoardo: I added a note explaining.

src/nemos/basis.py
```diff
@@ -1221,10 +1577,27 @@ def __init__(self, basis1: Basis, basis2: Basis) -> None:
         self._n_input_dimensionality = (
             basis1._n_input_dimensionality + basis2._n_input_dimensionality
         )
+        self._n_basis_input = None
+        self._n_output_features = None
+        self._label = "(" + basis1.label + " + " + basis2.label + ")"
```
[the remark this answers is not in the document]
BalzaniEdoardo: We definitely need the parenthesis for the multiplication, otherwise this 

```python
a,b,c = BSplineBasis(5),BSplineBasis(5),BSplineBasis(5)
example_2 = (a + b) * c
example_3 = a + (b * c)
```

Will result in the same label. I can drop the addition one

src/nemos/basis.py
```diff
@@ -611,6 +615,28 @@ def _check_convolution_kwargs(self):
...
+    @property
+    def n_basis_input(self) -> tuple | int | None:
+        # Needed for sklearn to work properly; cloning of sklearn requires that the
+        # parameter provided at initialization is returned as is: if an int is provided
+        # the same int is expected.
+        # Internally, for code consistency, the parameter is stored as a tuple for
+        # all basis.
+        if self._n_basis_input is None:
+            return
+        elif len(self._n_basis_input) == 1:
```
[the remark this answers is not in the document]
BalzaniEdoardo: It's a tuple for composite basis, additive and multiplicative. Having the parameter being a tuple makes the recursions easy, I don't have to deal with special cases, and this works for both composite and non-composite

```python
    def _set_num_output_features(self, *xi: NDArray) -> Basis:
        self._n_basis_input = (
            *self._basis1._set_num_output_features(
                *xi[: self._basis1._n_input_dimensionality]
            )._n_basis_input,
            *self._basis2._set_num_output_features(
                *xi[self._basis1._n_input_dimensionality :]
            )._n_basis_input,
        )
        self._n_output_features = (
            self._basis1.n_output_features + self._basis2.n_output_features
        )
        return self
```


In my first implementation, this was mandatory.  The `n_basis_input` was passed at initialization for the non-composite basis as an integer (most natural choice) but sklearn does not allow changing  the type of the parameter (to tuple) for their estimator API to work robustly. 

Now I can do whatever I want with the parameter. I'll just return the tuple

docs/tutorials/plot_06_calcium_imaging.py
```diff
@@ -166,9 +192,24 @@
 # ## Design matrix
 #
 # We can now create the design matrix by combining the head-direction of the animal and the activity of all other neurons.
-#
 X = basis.compute_features(head_direction, Y[:, selected_neurons])
 
+# %%
+# A design matrix including a CyclicBSpline and an intercept term is rank deficient.
```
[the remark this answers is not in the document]
BalzaniEdoardo: I like how you changed it, The explanation is at the right level

BalzaniEdoardo:
> I'm confused by the behavior of `split_by_feature`:
> 
>     * Why does it return 3d arrays? If I look at the example in the docstring, it returns arrays of shape (20,1,5) and (20,1,6). What's the middle dimension? This is also the behavior for a single basis or multiplicative basis.

Basis in conv mode allow for mutlidimensional inputs (intended for counts which are a TsdFrame in pynapple). The extra dimension. Example below,

```python
>>> import nemos as nmo
>>> import pynapple as nap
>>> import numpy as np
>>> counts = nap.TsdFrame(t=np.arange(100), d=np.random.poisson(size=(100, 2)))
>>> basis = nmo.basis.BSplineBasis(5, mode="conv", window_size=50)
>>> X = basis.compute_features(counts)
>>> X.shape
(100, 10)
>>> basis.split_by_feature(X, axis=1)["BSplineBasis"].shape
(100, 2, 5)

```

>     * This method only really make sense for the additive basis right? Otherwise it doesn't do anything but add the middle dimension? In that case, should it exist for the other bases objects? Or am I missing something?

I prefer that any basis should have the method for consistency. I can see a case in which one create the additive basis in a script, but the number of component is variable, from 1 to N, if you have the method, you can use the same exact code to process a regular basis ( 1 component) and the rest.
 
> I also don't think the tutorial we have on identifiability constraints right now is sufficient. Might not need to be addressed here, but should be added to the docs project if not. Should explain why it's bad to be rank-deficient, show what you gain from being full rank, etc.

Yes, totally

src/nemos/basis.py
```diff
@@ -1327,16 +1330,16 @@ def split_by_feature(
...
         each of shape:
 
-        \[
+        $$
         (..., n_i, b_i, ...)
-        \]
+        $$
 
         Here:
         - $n_i$ is the number of inputs processed by the **i-th basis component**.
```
billbrod, line 1345:
```suggestion

        - $n_i$ is the number of inputs processed by the **i-th basis component**.
```
Need an empty line here to render the docs correctly

billbrod (approved):
Just the small changes to the docstring we discussed, then this looks good to me!

PR #249, opened 2024-10-13
Added glm.py examples in docstrings

pranmod01:
This PR adds examples for the `glm.py` public-facing [commands](https://nemos.readthedocs.io/en/latest/reference/nemos/glm/#nemos.glm.GLM.predict). I added examples of how the function might be used with different parameters to get different results. 

To be discussed: These are my first set of examples for functions. Because I might not thoroughly understand how these examples might be used in the context of neural data, I'm unsure if I should add more details and scenarios of how the functions can be used.


billbrod:
Pranati was having an issue where running `isort` locally was giving inconsistent results with the CI and with my local machine. Turned out the issue was https://github.com/PyCQA/isort/issues/1790: there was a local folder named `jax` in her nemos directory, which made isort consider jax as an internal package, and changing how it was sorted. `jax` was a git submodule, and we're unsure how it ended up there. Deleting the directory caused isort to behave as expected.

BalzaniEdoardo (changes requested):
This is almost good to go but we should have an example in the `fit` method of PopulationGLM, since it overrides the GLM one.

pranmod01:
Updates made: 
1. Added example to the `fit` method of PopulationGLM
2. Changed some wording based on Edoardo's comments
3. Removed the basis example that overlapped with this branch to keep this branch clean of only GLM examples
4. Regular formatting tox fixes

src/nemos/basis.py
```diff
@@ -204,6 +204,7 @@ def fit(self, X: FeatureMatrix, y=None):
         -------
         self :
             The transformer object.
+
```
billbrod, line 207:
```suggestion
```
Surprised this isn't caught by ruff, but maybe because we're not including the docstrings?

src/nemos/glm.py
```diff
@@ -307,6 +326,21 @@ def predict(self, X: DESIGN_INPUT_TYPE) -> jnp.ndarray:
...
+        >>> X, y = np.random.normal(size=(10, 2)), np.random.uniform(size=10)
+        >>> Xnew = np.random.normal(size=(20, ) + X.shape[1:])
+
+        >>> # define and fit a GLM
+        >>> import nemos as nmo
+        >>> model = nmo.glm.GLM()
+        >>> model = model.fit(X, y)
+
+        >>> # predict new spike data
+        >>> predicted_spikes = model.predict(Xnew)
```
billbrod, line 342:
```suggestion
        >>> X, y = np.random.normal(size=(10, 2)), np.random.uniform(size=10)

        >>> # define and fit a GLM
        >>> import nemos as nmo
        >>> model = nmo.glm.GLM()
        >>> model = model.fit(X, y)

        >>> # predict new spike data
        >>> Xnew = np.random.normal(size=(20, X.shape[1]))
        >>> predicted_spikes = model.predict(Xnew)
```
I think it makes it a bit clearer to move `Xnew` to where it gets used, and to change that size syntax (X is only 2d anyway)

src/nemos/glm.py
```diff
@@ -407,6 +441,20 @@ def score(
...
+        >>> import numpy as np
+        >>> X, y = np.random.normal(size=(10, 2)), np.random.uniform(size=10)
+
+        >>> import nemos as nmo
+        >>> model = nmo.glm.GLM()
+        >>> model = model.fit(X, y)
+
+        >>> # get model score
+        >>> log_likelihood_score = model.score(X, y)
+        >>> pseudo_r2_score = model.score(X, y, score_type='pseudo-r2-McFadden')
```
billbrod, line 456: do we need to show both? not sure it adds anything

src/nemos/glm.py
```diff
@@ -605,6 +653,20 @@ def fit(
...
+        >>> import numpy as np
+        >>> X, y = np.random.normal(size=(10, 2)), np.random.uniform(size=10)
+
+        >>> # fit a ridge regression Poisson GLM
+        >>> import nemos as nmo
+        >>> model = nmo.glm.GLM(regularizer="Ridge", regularizer_strength=0.1)
+        >>> model = model.fit(X, y)
+
+        >>> # get model weights
+        >>> model_weights = model.coef_
```
billbrod, line 668: maybe show intercept as well?

src/nemos/glm.py
```diff
@@ -698,6 +760,21 @@ def simulate(
...
+        >>> Xnew = np.random.normal(size=(20, ) + X.shape[1:])
+
+        >>> # define and fit model
+        >>> import nemos as nmo
+        >>> model = nmo.glm.GLM()
+        >>> model = model.fit(X, y)
+
+        >>> # generate spikes and rates
+        >>> random_key = jax.random.key(123)
+        >>> spikes, rates = model.simulate(random_key, Xnew)
```
billbrod, line 777:
```suggestion

        >>> # define and fit model
        >>> import nemos as nmo
        >>> model = nmo.glm.GLM()
        >>> model = model.fit(X, y)

        >>> # generate spikes and rates
        >>> Xnew = np.random.normal(size=(20, X.shape[1]))
        >>> random_key = jax.random.key(123)
        >>> spikes, rates = model.simulate(random_key, Xnew)
```
same point about Xnew

src/nemos/glm.py
```diff
@@ -1376,6 +1453,26 @@ def fit(
         - If the mask is a `FeaturePytree`, then `"feature_name"` is a predictor of neuron `j` if
         `feature_mask["feature_name"][j] == 1`.
 
+        Examples
```
billbrod, line 1456: I find this example confusing. maybe add a couple of comments explaining what weights are and what they're used for here? feature_mask is fine, because it's explained elsewhere in the docstring

billbrod (changes requested):
This looks good! Couple more user-facing functions need examples, though @BalzaniEdoardo maybe some of these can be made private?
- `cast_to_jax`
- `initialize_state`

General notes:
- While it's fine to have the imports in the examples right above where they get used, I think it's more standard to have all imports at the top (that's how scripts and notebooks are generally written) -- any particular reason you didn't do it that way?

pranmod01:
For the comment about ordering the import statements -- I didn't put them all at the top and instead put them in the order that they are called just because I thought it would be easier for someone who didn't really understand the imports to see exactly how they are being used. But because the examples aren't too lengthy or technical, I can change the imports to the regular standard. Let me know!

Added an example to `initialize_state` but not `cast_to_jax` because the latter seems to be only used internally. Other than that, everything else is fixed from my end.

src/nemos/glm.py
```diff
@@ -104,6 +104,25 @@ class GLM(BaseRegressor):
     TypeError
         If provided `regularizer` or `observation_model` are not valid.
 
+    Examples
+    --------
+    >>> import nemos as nmo
+
+    >>> # define simple GLM model
```
BalzaniEdoardo, line 111:
```suggestion
    >>> # define single neuron GLM model
```

src/nemos/glm.py
```diff
@@ -407,6 +441,19 @@ def score(
...
+        >>> # example input
+        >>> import numpy as np
+        >>> X, y = np.random.normal(size=(10, 2)), np.random.uniform(size=10)
+
+        >>> import nemos as nmo
+        >>> model = nmo.glm.GLM()
+        >>> model = model.fit(X, y)
+
+        >>> # get model score
+        >>> log_likelihood_score = model.score(X, y)
```
BalzaniEdoardo, line 455:
```suggestion
        >>> log_likelihood_score = model.score(X, y)
        
        >>> # get a pseudo-R2 score
        >>> pseudo_r2 = model.score(X, y, score_type="pseudo-r2-McFadden")
```

src/nemos/glm.py
```diff
@@ -407,6 +441,19 @@ def score(
             If X structure doesn't match the params, and if X and y have different
             number of samples.
 
+        Examples
+        --------
+        >>> # example input
+        >>> import numpy as np
+        >>> X, y = np.random.normal(size=(10, 2)), np.random.uniform(size=10)
```
BalzaniEdoardo, line 448:
```suggestion
        >>> X, y = np.random.normal(size=(10, 2)), np.random.poisson(size=10)
```

The default PoissonObservations model discrete variables, if you pass a continuous variable, the function you are maximizing is not a likelihood anymore.

src/nemos/glm.py
```diff
@@ -307,6 +326,21 @@ def predict(self, X: DESIGN_INPUT_TYPE) -> jnp.ndarray:
             - If `X` is not three-dimensional.
             - If there's an inconsistent number of features between spike basis coefficients and `X`.
 
+        Examples
+        --------
+        >>> # example input
+        >>> import numpy as np
+        >>> X, y = np.random.normal(size=(10, 2)), np.random.uniform(size=10)
```
BalzaniEdoardo, line 333:
```suggestion
        >>> X, y = np.random.normal(size=(10, 2)), np.random.poisson(size=10)
```

src/nemos/glm.py
```diff
@@ -605,6 +652,21 @@ def fit(
             - If `init_params` are not array-like
             - If `init_params[i]` cannot be converted to jnp.ndarray for all i
 
+        Examples
+        -------
+        >>> # example input
+        >>> import numpy as np
+        >>> X, y = np.random.normal(size=(10, 2)), np.random.uniform(size=10)
```
BalzaniEdoardo, line 659:
```suggestion
        >>> X, y = np.random.normal(size=(10, 2)), np.random.poisson(size=10)
```

src/nemos/glm.py
```diff
@@ -698,6 +760,21 @@ def simulate(
         ValueError
             - If the instance has not been previously fitted.
 
+        Examples
+        --------
+        >>> # example input
+        >>> import numpy as np
+        >>> X, y = np.random.normal(size=(10, 2)), np.random.uniform(size=10)
```
BalzaniEdoardo, line 767:
```suggestion
        >>> X, y = np.random.normal(size=(10, 2)), np.random.poisson(size=10)
```

BalzaniEdoardo (changes requested):
Last very small round of comments, and then this is good to be merged. Thanks Pranati, very good job on this

src/nemos/glm.py
```diff
@@ -307,6 +326,21 @@ def predict(self, X: DESIGN_INPUT_TYPE) -> jnp.ndarray:
...
+
+        >>> # define and fit a GLM
+        >>> import nemos as nmo
+        >>> model = nmo.glm.GLM()
+        >>> model = model.fit(X, y)
+
+        >>> # predict new spike data
+        >>> Xnew = np.random.normal(size=(20, ) + X.shape[1:])
```
billbrod, line 341:
```suggestion
        >>> Xnew = np.random.normal(size=(20, X.shape[1]))
```

src/nemos/glm.py
billbrod: I still don't think we need to show both here

src/nemos/glm.py
```diff
@@ -698,6 +760,21 @@ def simulate(
...
+        >>> X, y = np.random.normal(size=(10, 2)), np.random.uniform(size=10)
+
+        >>> # define and fit model
+        >>> import nemos as nmo
+        >>> model = nmo.glm.GLM()
+        >>> model = model.fit(X, y)
+
+        >>> # generate spikes and rates
+        >>> random_key = jax.random.key(123)
+        >>> Xnew = np.random.normal(size=(20, ) + X.shape[1:])
```
billbrod, line 776:
```suggestion
        >>> Xnew = np.random.normal(size=(20, X.shape[1]))
```

billbrod (changes requested):
I still think the definition of size for Xnew should be changed, but otherwise this looks good to me!

src/nemos/glm.py
BalzaniEdoardo: I disagree

BalzaniEdoardo: It is an important parameter and it's a few extra lines, let's keep it

BalzaniEdoardo:
> This looks good! Couple more user-facing functions need examples, though @BalzaniEdoardo maybe some of these can be made private?
> 
> * `cast_to_jax`
> * `initialize_state`
> 
> General notes:
> 
> * While it's fine to have the imports in the examples right above where they get used, I think it's more standard to have all imports at the top (that's how scripts and notebooks are generally written) -- any particular reason you didn't do it that way?

Let's make 'cast_to_jax' private, keep 'initialize_state' public since batching as we have it now requires the user to call the method, if you check the relative how to guide in the docs.

billbrod (approved):
All looks good to me, thanks for your work on this Pranati!

@BalzaniEdoardo should we open an issue about making `cast_to_jax` private? (doing that probably doesn't belong here)

BalzaniEdoardo (approved):

BalzaniEdoardo:
Thanks @pranmod01!

PR #250, opened 2024-10-14
Splits tox into multiple jobs

billbrod:
With this PR, we now have three possible tox environments: the default (run pytest), `fix` (runs black and isort commands that edit `src/` and some of `docs`), and `check` (just check those directories and raise error).

Then additionally made changes to `ci.yml`:
- adds new `check` job, which just runs the `check` tox environment.
- remove `pip install -e .[dev]` during set up of tox (because tox will install all requirements into its own 
- final `check` job now requires all of `tests`, `prevent_docs_absolute_links`, and `check`environments)

Thus our CI should be slightly faster (as we're no longer installing unnecessary packages and only running black+isort+flake8 with one OS / python version, instead of all)

I think that by editing `envlist` at the top of the `tox.ini`, we can make it so the default tox action is to run the test with a single python version and also run `fix`, do we want that?

Finally, CONTRIBUTING still needs to be updated. if this all looks good to you, I can go ahead and update it to reflect these changes.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,7 +18,7 @@ on:
        - main
 
 jobs:
-  tox:
+  tests:
     if: ${{ !github.event.pull_request.draft }}
     strategy:
       matrix:
@@ -38,7 +38,6 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
-          python -m pip install .
           pip install tox
 
       - name: Run tox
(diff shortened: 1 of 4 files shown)


billbrod:
Most recent pushes:
- Update CONTRIBUTING to describe recent changes to tox
- adds parallelization for pytest, using `pytest-xdist`
- change tox's default behavior to run test with one python version and `fix`

BalzaniEdoardo (approved):
this is great! thanks Billy! I would keep the toxenv as is now, but no strong preference.

PR #251, opened 2024-10-21
Added docstring examples to basis functions

pranmod01:
What I added: Simple examples for the public-facing functions of the Basis classes. Because most of the class definitions are derived the abstract classes, I didn't add examples to each class/function to avoid redundancy. 

Potential discussion: Similar to glm examples, I think the examples can be more in depth and/or add different contexts where the functions might be used. But I'm unsure on the domain knowledge so would need more guidance there. This is just my initial idea, and I’m open to feedback to make the examples more detailed.


src/nemos/basis.py
```diff
@@ -223,6 +236,21 @@ def transform(self, X: FeatureMatrix, y=None) -> FeatureMatrix:
...
+        --------
+        # Example input
+        >>> import numpy as np
+        >>> X, y = np.random.normal(size=(100, 2)), np.random.uniform(size=100)
+
+        # Define and fit tranformation basis
+        >>> from nemos.basis import MSplineBasis, TransformerBasis
+        >>> basis = MSplineBasis(10)
+        >>> transformer = TransformerBasis(basis)
+        >>> transformer_fitted = transformer.fit(X) # input must be a 2d array
```
BalzaniEdoardo, line 250: I would comment about what "fit" is doing here. Say something like: fit in mode"conv" creates the set of basis that will be convolve with the input. In mode "eval" does nothing, but we need the method to fully comply with sklearn API.

src/nemos/basis.py
```diff
@@ -1071,7 +1132,28 @@ class AdditiveBasis(Basis):
     n_basis_funcs : int
         Number of basis functions.
 
-
+    Examples
+    --------
+    # Generate sample data
+    >>> import numpy as np
+    >>> import nemos as nmo
+    >>> X, y = np.random.normal(size=(30, 2)), np.random.poisson(size=30)
```
BalzaniEdoardo, line 1140: add a comment saying 
```python
>>> # X.shape is (n_samples, n_inputs), where n_inputs is the number required by the basis
```

src/nemos/basis.py
```diff
@@ -1183,6 +1265,27 @@ class MultiplicativeBasis(Basis):
...
+
+    # define two basis and multiply
+    >>> basis_1 = nmo.basis.BSplineBasis(10)
+    >>> basis_2 = nmo.basis.RaisedCosineBasisLinear(15)
+    >>> multiplicative_basis = nmo.basis.MultiplicativeBasis(basis1=basis_1, basis2=basis_2)
+    >>> transformed_X = multiplicative_basis.to_transformer().transform(X[:, 0:2])
+    >>> print(transformed_X.shape)
+    (30, 150)
+
+    # can multiply or add another basis to the AdditiveBasis object
```
BalzaniEdoardo, line 1283: warn that the number of output features of the basis grows as the product of the n_features of the 1d basis. if you have 10, 15 and 100, the final basis will output 15000, as you are showing

src/nemos/basis.py
```diff
@@ -223,6 +236,21 @@ def transform(self, X: FeatureMatrix, y=None) -> FeatureMatrix:
...
+
+        Examples
+        --------
+        # Example input
+        >>> import numpy as np
+        >>> X, y = np.random.normal(size=(100, 2)), np.random.uniform(size=100)
+
+        # Define and fit tranformation basis
+        >>> from nemos.basis import MSplineBasis, TransformerBasis
+        >>> basis = MSplineBasis(10)
```
BalzaniEdoardo, line 248:
```suggestion
        >>> basis = MSplineBasis(10, mode="conv", window_size=200)
```

src/nemos/basis.py
BalzaniEdoardo: Assume that the
```suggestion
        >>> # Before calling `fit` the convolution kernel is not set
        >>> transformer.kernel_
        
        >>> transformer_fitted = transformer.fit(X) # input must be a 2d array
        >>> # Now the convolution kernel is initialized and has shape (window_size, n_basis_funcs)
        >>> transformer_fitted.kernel.shape
        (200, 10)
        
```

src/nemos/basis.py
```diff
@@ -248,6 +276,20 @@ def fit_transform(self, X: FeatureMatrix, y=None) -> FeatureMatrix:
         array-like
             The data transformed by the basis functions, after fitting the basis
             functions to the data.
+
+        Examples
+        --------
+        # Example input
+        >>> import numpy as np
+        >>> X, y = np.random.normal(size=(100, 2)), np.random.uniform(size=100)
```
BalzaniEdoardo, line 284:
```suggestion
        >>> X, y = np.random.normal(size=(100, 1)), np.random.uniform(size=100)
```

src/nemos/basis.py
```diff
@@ -248,6 +276,20 @@ def fit_transform(self, X: FeatureMatrix, y=None) -> FeatureMatrix:
...
+
+        # Define tranformation basis
+        >>> from nemos.basis import MSplineBasis, TransformerBasis
+        >>> basis = MSplineBasis(10)
+        >>> transformer = TransformerBasis(basis)
+
+        # Fit and transform basis
+        >>> feature_transformed = transformer.fit_transform(X[:, 0:1]) # input must be a 2d array, (num_samples, 1)
```
BalzaniEdoardo, line 292:
```suggestion
        >>> feature_transformed = transformer.fit_transform(X) # input must be a 2d array, (num_samples, 1)
```

BalzaniEdoardo (changes requested):
Hi Pranati,
First of all, nice work! The examples are looking good.

We are still missing a few of user facing methods, for example,
- compute_features for Basis
- __call__ for all concrete classes
- evaluate_on_grid for all concrete classes.

Might sound like repetition, but we should think of docstrings of these classes as an helper for the user, so repetition in this case is good.
It's likely that the user will look for an example on the method of the concrete sub-class that he is currently using and the info should be there.

After this, we are ready to merge this one

BalzaniEdoardo:
> What I added: Simple examples for the public-facing functions of the Basis classes. Because most of the class definitions are derived the abstract classes, I didn't add examples to each class/function to avoid redundancy.
> 
> Potential discussion: Similar to glm examples, I think the examples can be more in depth and/or add different contexts where the functions might be used. But I'm unsure on the domain knowledge so would need more guidance there. This is just my initial idea, and I’m open to feedback to make the examples more detailed.

I think this is a good point, but probably, we should keep the example simple in the docstrings and have a link to other parts of the documentation in which we showcase the use of basis more in depth!

pranmod01:
I have started this branch from the glm branch instead of the `main` branch. So I am deleting this branch and will start a new branch with a similar name containing basis examples and no glm examples—closing this PR.

PR #254, opened 2024-10-23
require sample axis to be axis=0

BalzaniEdoardo:
This PR fixes a corner-case bug related to basis in `mode="conv"`. 

## Issue

Before this PR, one could pass at basis initialization any kwargs from `nemos.convolve.create_convolutional_predictor`. This included `axis` meaning that one could specify what's the axis along which the convolution is applied.

What happened in `basis.compute_features` is that the axis is transposed to the first, and the rest of the axis are flattened to the second axis, so that the compute_features returns an 2darray. However, when the basis is composite, the we check that the number of samples matches for all input, and this check relied on the fact that the sample axis is `axis=0`.

This caused the following corner case:
```python
>>> b = nmo.basis.RaisedCosineBasisLog(5, mode="conv", window_size=100, axis=1)

>>> # this works because there is one input (the check on the sample axis passes)
>>> b.compute_features(np.ones((2, 200)))
Array([[       nan,        nan,        nan, ...,        nan,        nan,
               nan],
       [       nan,        nan,        nan, ...,        nan,        nan,
               nan],
       [       nan,        nan,        nan, ...,        nan,        nan,
               nan],
       ...,
       [ 2.4915924,  2.4915924,  5.5825534, ..., 20.70655  , 39.878582 ,
        39.878582 ],
       [ 2.4915924,  2.4915924,  5.5825534, ..., 20.70655  , 39.878582 ,
        39.878582 ],
       [ 2.4915924,  2.4915924,  5.5825534, ..., 20.706553 , 39.878582 ,
        39.878586 ]], dtype=float32)

>>> # this fails because the check on the sample axis will raise an error since 3 != 2
>>> (b + b).compute_features(np.ones((2, 200)), np.ones((3, 200)))
```

## Solution

Instead of complicating the check logic, or move it after the transposition happens for all 1D basis, I simplified our life and enforced that the sample axis is always the first one. This is always true for pynapple, and requires the user to transpose an array if the data are organized differently;

Now the initialization would raise an error if axis different from 0 is passed.
```python
>>> b = nmo.basis.RaisedCosineBasisLog(5, mode="conv", window_size=100, axis=1)
ValueError: Invalid `axis=1` provided. Basis requires  requires the convolution to be applied along the first axis (`axis=0`).
Please transpose your input so that the desired axis for convolution is the first dimension (axis=0).
```

### DIFF
--- a/src/nemos/basis.py
+++ b/src/nemos/basis.py
@@ -545,7 +545,7 @@ class Basis(Base, abc.ABC):
             if param.default
             # prevent user from passing
             # `basis_matrix` or `time_series` in kwargs.
-            is not inspect.Parameter.empty  
+            is not inspect.Parameter.empty
         }
         if not set(self._conv_kwargs.keys()).issubset(convolve_configs):
             # do not encourage to set axis.
--- a/src/nemos/basis.py
+++ b/src/nemos/basis.py
@@ -529,7 +529,9 @@ class Basis(Base, abc.ABC):
         convolve_configs = convolve_configs.difference({"axis"})
         if not set(kwargs.keys()).issubset(convolve_configs):
             # remove axis in case axis=0, was passed which is allowed.
-            invalid = set(kwargs.keys()).difference(convolve_configs).difference({"axis"})
+            invalid = (
+                set(kwargs.keys()).difference(convolve_configs).difference({"axis"})
+            )
             raise ValueError(
                 f"Unrecognized keyword arguments: {invalid}. "
                 f"Allowed convolution keyword arguments are: {convolve_configs}."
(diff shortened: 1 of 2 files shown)


src/nemos/basis.py
```diff
@@ -496,14 +509,57 @@ def __init__(
...
+                "Please transpose your input so that the desired axis for "
+                "convolution is the first dimension (axis=0)."
+            )
+        convolve_params = inspect.signature(create_convolutional_predictor).parameters
+        convolve_configs = {
+            key
+            for key, param in convolve_params.items()
+            if param.default
+            is not inspect.Parameter.empty  # prevent user from passing directly
+            # `basis_matrix` or `time_series` in kwargs.
```
billbrod, line 547:
```suggestion
            # prevent user from passing
            # `basis_matrix` or `time_series` in kwargs.
            is not inspect.Parameter.empty  
```

billbrod (commented):
My vote would be to have axis just not be allowed at all. Since the behavior cannot be modified, it feels a bit confusing to set it.

PR #275, opened 2024-12-04
Multi dim eval basis

BalzaniEdoardo:
This PR follows #274, additional edits to the basis for generalizing the behavior and removing the `__call__` method.

- Allow for N-dim inputs for both conv and eval basis
- Rename the `__call__` method of basis to `_evaluate`
- Make sure that the notebook exec error break the build (conf.py config)

EDIT: 

- I made the "set_kernel" method public because I am using it in the doc mostly
- The set_kernel will be removed from Eval basis in PR #276 (as discussed with @billbrod )

PR #276, opened 2024-12-05
Set shape basis method

BalzaniEdoardo:
Add a `set_input_shape` method that initializes the basis.

The method can accept a list of:
- Integers: assumes that the input will be flat (if the integer n = 1) or 2D (for n > 1). In the first case the basis allows inputs of shape (n_samples,) in the second (n_samples, n).
- Tuple of integers: the tuple corresponding to the input shape trimming the first axis (sample axis). If input is  (n, m, ...) then the basis expects inputs of shape (n_samples, n, m,...)
- Arrays: directly the inputs that are going to be provided from compute features.

It prepares the basis and precomputes some internal quantities that are useful for subsequent functionalities: knowing how to split the feature axis, i.e. `split_by_features`. In the next PR, `TransformerBasis` will use the pre-computed shapes to split out the concatenated inputs before processing. 

Additionally, it will compute and store the number of output features, this could be an information that a user may want to have after building a complex composite basis.

This PR follows #275 

EDIT:
# PR Summary

In this PR, I refined the class structure to better separate basis attributes and methods, delegating their validation logic to mixin classes wherever possible.

## New Classes

- **`CompositeBasisMixin`**:  
  This mixin is inherited by additive and multiplicative bases. It implements methods for traversing the composite basis tree, such as `__sklearn_clone__` and `setup_basis`.

- **`AtomicBasisMixin`**:  
  Designed for non-composite (atomic) bases, this mixin stores the `n_basis_funcs` parameter and implements selected methods like `__sklearn_clone__`, which have uniform implementations across all atomic bases.

## New Abstract Methods

- **`set_input_shape`**:  
  This method stores the expected input shape, a state attribute required for compatibility with transformers. Parameters set by this method are carried over during cloning, such as in cross-validation. Concrete implementations are provided in `AtomicBasisMixin` and `CompositeBasisMixin`.

- **`_set_input_independent_states`**:  
  Responsible for setting all state variables that depend on class parameters (provided at initialization and retrievable with `get_params`).

- **`setup_basis`**:  
  Computes all state variables, both input-dependent (e.g., input shape) and input-independent (e.g., kernels for convolutional bases). Concrete implementations are found in `Eval`, `Conv`, and `Composite` basis mixins.

## Clone Method for Bases

- **`__sklearn_clone__`**:  
  Implements cloning logic to retain input-dependent states (such as input shape), which would otherwise be discarded by `sklearn.base.clone`. This is implemented in both the `CompositeBasisMixin` and `AtomicBasisMixin`.

## Modified/Moved Attributes and Methods

- **Kernel-related logic**:  
  The `kernel_` attribute, along with the `_check_has_kernel` method and `set_kernel`, has been moved to the `Conv` mixin.

- **Input shape validation**:  
  The `_check_input_shape_consistency` method has been relocated to `AtomicBasisMixin` and `CompositeBasisMixin`.

- **Composite basis setters**:  
  Setters for `basis1` and `basis2` in composite bases now support cross-validation scenarios.

## Inheritance of New Mixins

- **`CompositeBasisMixin`**:  
  Inherited by `AdditiveBasis` and `MultiplicativeBasis`.

- **`AtomicBasisMixin`**:  
  Inherited by the following classes:  
  - `SplineBasis` (the superclass for all splines)  
  - `RaisedCosineLinear` (the superclass for all raised cosines)  
  - `OrthExponentialBasis` (the superclass for orthogonal exponential bases)


BalzaniEdoardo:
In the next PR I'll improve the API of the transformer and move the TransformerBasis tests in a dedicated script

docs/how_to_guide/plot_06_sklearn_pipeline_cv_demo.md
```diff
@@ -538,7 +529,7 @@ The plot confirms that the firing rate distribution is accurately captured by ou
 :::{warning}
 Please note that because it would lead to unexpected behavior, mixing the two ways of defining values for the parameter grid is not allowed. The following would lead to an error:
 
-```python
+```{code} ipython
```
sjvenditto, line 532: `.set_transformer_input(1)` should be added to all basis definitions in this code block

sjvenditto (changes requested):
Looking good! I just have a couple of suggested changes to clarify in documentation/error messages where `set_input_shape` will have an impact. Also, some code in a Warning admonition in plot_06_sklearn_pipeline_cv_demo needs to be updated with the new syntax

docs/how_to_guide/plot_05_transformer_basis.md
```diff
@@ -0,0 +1,183 @@
...
+    trans_bas.fit_transform(inp)
+except RuntimeError as e:
+    print(repr(e))
+```
+
+...Unfortunately, not yet. The problem is that the basis has never interacted with the two separate inputs, and therefore doesn't know which columns of `inp` should be processed by `count_basis` and which by `speed_basis`.
+
+You can provide this information by calling the `set_input_shape` method of the basis. 
+
+This can be called before or after the transformer basis is defined. The method extracts and stores the array shapes excluding the sample axis (which won't be affected in the concatenation).
```
billbrod, line 120:
```suggestion
This can be called before or after the transformer basis is defined. The method extracts and stores the number of columns for each input. There are multiple ways to call this method:

- It directly accepts the input: `composite_basis.set_input_shape(counts, speed)`
...
```

It's too many lines to make a good looking suggestion, but I think this should be a list like the above, with only a single line per example (rather than a full code block, though ideally still with the copy button and syntax highlighting).

I don't think we need to show the `out = composite_basis.to_transformer().fit_transform(inp)` line in each of them, just show it at the end.

Similarly, should have an admonition (or another small paragraph?) mentioning that we are looking at all dimensions except the first (since the samples are irrelevant here), and show a 4-dimensional example.

docs/how_to_guide/plot_05_transformer_basis.md
```diff
@@ -0,0 +1,183 @@
...
+out = trans_bas.fit_transform(inp)
+```
+
+:::{note}
+
+If you define a basis and call `compute_features` on your inputs, internally, it will store its shapes, 
+and the `TransformerBasis` will be ready to process without any direct call to `set_input_shape`.
+:::
+
+If for some reason you need to provide an input of different shape to an already set-up transformer, you must reset the
```
billbrod, line 165: I think this point should become an admonition (warning), without the example.

docs/how_to_guide/plot_05_transformer_basis.md
```diff
@@ -0,0 +1,183 @@
...
+
+trans_bas = composite_basis.to_transformer()
+trans_bas.set_input_shape(3, (2, 3)) 
+out2 = trans_bas.fit_transform(inp2)
+```
+
+
+### Learn more
+
+If you want to learn more about how to select basis' hyperparameters with `sklearn` pipelining and cross-validation, check out [this how-to guide](sklearn-how-to).
```
billbrod, line 182:
```suggestion
If you want to see how `TransformerBasis` allows us to make use of  `sklearn` pipelining and cross-validation machinery to select basis' hyperparameters, check out [this how-to guide](sklearn-how-to).
```

docs/how_to_guide/plot_06_sklearn_pipeline_cv_demo.md
```diff
@@ -92,7 +93,7 @@ Here we used a placeholder `"label_i"` for demonstration; you should choose a mo
 :::
 
 Calling `pipe.fit(X, y)` will perform the following computations:
-```python
+```{code} ipython3
```
billbrod, line 96:
```suggestion
# we don't want to run this, just use syntax highlighting
```{code} ipython3
```
because I got confused

docs/quickstart.md
billbrod, line 1: Looks like the only change here is to run this, right?

src/nemos/basis/_decaying_exponential.py
```diff
@@ -40,8 +41,8 @@ def __init__(
         mode="eval",
         label: Optional[str] = "OrthExponentialBasis",
     ):
+        AtomicBasisMixin.__init__(self, n_basis_funcs=n_basis_funcs)
```
billbrod, line 44: I think we've already discussed this, but there's no way to have `super().__init__()` initialize both super classes?

src/nemos/basis/_basis.py
```diff
@@ -135,28 +133,26 @@ class Basis(Base, abc.ABC, BasisTransformerMixin):
 
     def __init__(
         self,
-        n_basis_funcs: int,
         mode: Literal["eval", "conv"] = "eval",
         label: Optional[str] = None,
     ) -> None:
-        self.n_basis_funcs = n_basis_funcs
-        self._n_input_dimensionality = 0
+        self._n_basis_funcs = getattr(self, "_n_basis_funcs", None)
```
billbrod, line 139: why is this just not in the AtomicBasis mixing?

src/nemos/basis/_basis.py
```diff
@@ -135,28 +133,26 @@ class Basis(Base, abc.ABC, BasisTransformerMixin):
...
-
         if label is None:
             self._label = self.__class__.__name__
         else:
             self._label = str(label)
 
-        self.kernel_ = None
+        # specified only after inputs/input shapes are provided
+        self._n_basis_input_ = getattr(self, "_n_basis_input_", None)
+        self._input_shape_ = getattr(self, "_input_shape_", None)
```
billbrod, line 151: what's the context in which these will be set here?

src/nemos/basis/_basis.py
```diff
@@ -835,33 +742,57 @@ class AdditiveBasis(Basis):
     """
 
     def __init__(self, basis1: Basis, basis2: Basis) -> None:
-        self.n_basis_funcs = basis1.n_basis_funcs + basis2.n_basis_funcs
-        super().__init__(self.n_basis_funcs, mode="eval")
+        CompositeBasisMixin.__init__(self, basis1, basis2)
+        Basis.__init__(self, mode="eval")
```
billbrod, line 746: why is this eval? shouldn't it be based on basis1 / basis2?

src/nemos/basis/_basis.py
```diff
@@ -1246,38 +1153,30 @@ class MultiplicativeBasis(Basis):
     """
 
     def __init__(self, basis1: Basis, basis2: Basis) -> None:
-        self.n_basis_funcs = basis1.n_basis_funcs * basis2.n_basis_funcs
-        super().__init__(self.n_basis_funcs, mode="eval")
+        CompositeBasisMixin.__init__(self, basis1, basis2)
+        Basis.__init__(self, mode="eval")
```
billbrod, line 1157: same question about mode

src/nemos/basis/_basis_mixin.py
```diff
@@ -13,6 +17,166 @@
 from ..convolve import create_convolutional_predictor
 from ._transformer_basis import TransformerBasis
 
+if TYPE_CHECKING:
+    from ._basis import Basis
+
+
+def set_input_shape_state(method):
```
billbrod, line 24: why not make this more general and accept a list of attributes to copy, which default to `["_n_basis_input_", "_input_shape_"]`?

src/nemos/basis/_basis_mixin.py
```diff
@@ -48,9 +212,32 @@ def _compute_features(self, *xi: ArrayLike | Tsd | TsdFrame | TsdTensor):
...
+
+        Returns
+        -------
+        :
+            The basis with ready for evaluation.
+        """
+        self.set_input_shape(*xi)
+        return self
+
+    def _set_input_independent_states(self) -> "EvalBasisMixin":
```
billbrod, line 238: currently this only appears to call `set_kernel` in the convolutional mixin and does nothing anywhere else -- do we really need it?

src/nemos/basis/_basis_mixin.py
```diff
@@ -255,4 +489,182 @@ def to_transformer(self) -> TransformerBasis:
...
+            *(bas2._input_shape_ for bas2 in basis2._iterate_over_components()),
+        )
+        # if all bases where set, then set input for composition.
+        set_bases = [s is not None for s in shapes]
+
+        if all(set_bases):
+            # pass down the input shapes
+            self.set_input_shape(*shapes)
+        elif any(set_bases):
+            warnings.warn(
```
billbrod, line 523: this isn't actually necessary, the following runs fine. should it?
```python
b = nmo.basis.BSplineEval(5)
b.setup_basis(np.ones((100, 5)))
b2 = nmo.basis.BSplineEval(5)
c = b+b2
c.compute_features(np.ones((100, 5)), np.ones((100,2)))

```

src/nemos/basis/_basis_mixin.py
```diff
@@ -255,4 +489,182 @@ def to_transformer(self) -> TransformerBasis:
...
+            *xi[self._basis1._n_input_dimensionality :]
+        )
+
+    @property
+    def basis1(self):
+        return self._basis1
+
+    @basis1.setter
+    def basis1(self, bas: Basis):
+        self._basis1 = bas
```
billbrod, line 590: why are these properties if we do nothing special in the getter or setter? why not just use `self.basis1` instead of `self._basis1`?

src/nemos/basis/basis.py
```diff
@@ -1514,3 +1793,56 @@ def split_by_feature(
...
+        # if n_basis_funcs is not yet initialized, skip check
+        n_basis = getattr(self, "n_basis_funcs", None)
+        if n_basis and window_size < n_basis:
+            raise ValueError(
+                "OrthExponentialConv basis requires at least a window_size larger then the number "
+                f"of basis functions. window_size is {window_size}, n_basis_funcs while"
+                f"is {self.n_basis_funcs}."
+            )
+
+    def set_kernel(self):
```
billbrod, line 1834: will this ever get reached? since `_check_window_size` gets called during init?

billbrod (changes requested):
Haven't had a chance to look at tests, but here's my first go through. Some tweaking in the tutorials, and also:

- I think we can probably move away from using numbers in the names of the tutorials. The only reason to do so is for the automatic sorting to work, but we're no longer making use of automatic sorting, right?
- I am confused with what lives in the `Basis` superclass and what lives in `AtomicBasisMixin`. For example, why does anything  to do with n_basis_input or n_basis_funcs live in Basis?

docs/how_to_guide/plot_05_transformer_basis.md
```diff
@@ -0,0 +1,183 @@
...
+
+All of `scikit-learn`'s machinery relies on strict assumptions about input structure. In particular, all `scikit-learn` 
+objects require inputs as arrays of at most two dimensions, where the first dimension represents the time (or samples) 
+axis, and the second dimension represents features. 
+While this may feel rigid, it enables transformations to be seamlessly chained together, greatly simplifying the 
+process of building stable, complex pipelines.
+
+On the other hand, `NeMoS` takes a different approach to feature construction. `NeMoS`' bases are composable constructors that allow for more flexibility in the required input structure. 
+Depending on the basis type, it can accept one or more input arrays or `pynapple` time series data, each of which can take any shape as long as the time (or sample) axis is the first of each array;
+`NeMoS` design favours object composability: one can combine any two or more bases to compute complex features, with a user-friendly interface that can accept a separate array/time series for each input type (e.g., an array with the spike counts, an array for the animal's position, etc.).
```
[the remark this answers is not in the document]
BalzaniEdoardo: correct!

docs/how_to_guide/plot_05_transformer_basis.md
```diff
@@ -0,0 +1,183 @@
...
+
+# concatenate the inputs
+inp = np.concatenate([counts, speed[:, np.newaxis]], axis=1)
+print(inp.shape)
+
+try:
+    trans_bas.fit_transform(inp)
+except RuntimeError as e:
+    print(repr(e))
+```
```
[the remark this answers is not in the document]
BalzaniEdoardo: cool

docs/quickstart.md
BalzaniEdoardo: yes

src/nemos/basis/_decaying_exponential.py
BalzaniEdoardo: This modification is quite hard and requires some precision, so we can leave it for the future.

The method that in theory can work is the first one `super().__init__()` however, for this to work, for all the classes inheriting directly `AtomicBasis` and `Basis`, we need to invert the inheritance order. 

This also doesn't work directly, because `AtomicBasis` is using some parameters that are defined at basis initialization.

I think this all can be smoothed out but requires precision and a dedicate PR. I'll add it to an issue

src/nemos/basis/_basis.py
BalzaniEdoardo: no reason, moved it

src/nemos/basis/_basis.py
BalzaniEdoardo: never

src/nemos/basis/_basis.py
BalzaniEdoardo: we should probably say "composite", it is unspecified, one can be eval and one conv, or it could be eva/composite->eval/conv or any complex struct

src/nemos/basis/_basis_mixin.py
BalzaniEdoardo: yes, and also for the future: I want to funnel every other setting through this method (for basis that do not exist yet :))

src/nemos/basis/_basis_mixin.py
BalzaniEdoardo: this always works, but the transformer version no, I can move the warning to where it is more appropriate

BalzaniEdoardo: actually, whatever, if one tries to use the transformer it will have an error and then figure out that the shape was not initialized

docs/how_to_guide/plot_05_transformer_basis.md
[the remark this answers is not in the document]
BalzaniEdoardo: I tried, but it prints the way too much stuff, so I am rolling back to my solution for now

PR #282, opened 2024-12-19
Identity basis

BalzaniEdoardo:
In this PR I add an identity basis function and a basis function that capture raw history.

I added a few how-to guide that showcase the use of these basis for:

1. Including custom predictors in a composite basis.
2. Create a design from categorical variables.
3. Fit a coupled GLM.


src/nemos/basis/_identity.py
```diff
@@ -0,0 +1,183 @@
...
+        vmax = np.nanmax(sample_pts, axis=0) if self.bounds is None else self.bounds[1]
+        sample_pts[(sample_pts < vmin) | (sample_pts > vmax)] = np.nan
+        check_fraction_valid_samples(
+            sample_pts,
+            err_msg="All the samples lie outside the [vmin, vmax] range.",
+            warn_msg="More than 90% of the samples lie outside the [vmin, vmax] range.",
+        )
+        return sample_pts[..., np.newaxis]
+
+    def evaluate_on_grid(self, n_samples: int) -> Tuple[Tuple[NDArray], NDArray]:
```
billbrod, line 72: I think return type is wrong? shouldn't it just be `Tuple[NDarray, NDArray]`?

src/nemos/basis/_identity.py
```diff
@@ -0,0 +1,183 @@
...
+            The equi-spaced sample points X as a 2D array (n_samples, 1).
+        """
+        return super().evaluate_on_grid(n_samples)
+
+    def _check_n_basis_min(self):
+        """No checks necessary."""
+        pass
+
+    @property
+    def n_basis_funcs(self) -> tuple | None:
```
billbrod, line 96: why is this necessary? because it's read-only?

src/nemos/basis/_identity.py
```diff
@@ -0,0 +1,183 @@
...
+        """Read-only property for identity."""
```
billbrod, line 97:
```suggestion
        """Number of basis functions (read-only)."""
```

src/nemos/basis/basis.py
```diff
@@ -1830,3 +1825,248 @@ def _check_window_size(self, window_size: int):
...
+        >>> _ = basis.set_input_shape(x)
+        >>> basis.n_output_features
+        20
+
+        """
+        return AtomicBasisMixin.set_input_shape(self, xi)
+
+
+class HistoryConv(ConvBasisMixin, HistoryBasis):
+    """Identity basis function.
```
billbrod, line 1947: Need a better one-liner here, at least to distinguish with IdentityEval

docs/background/basis/README.md
```diff
@@ -64,6 +64,21 @@
      - :class:`~nemos.basis.OrthExponentialEval`  :raw-html:`<br />`
        :class:`~nemos.basis.OrthExponentialConv`
      - 🟢 Eval
+   * - **Identity Function**
```
billbrod, line 67: Feels odd to have these two be separate, as they're then the only ones that don't have both modes. Why not include them in the same row?

docs/how_to_guide/categorical_predictors.md
```diff
@@ -0,0 +1,120 @@
...
+    jupytext_version: 1.16.4
+kernelspec:
+  display_name: Python 3
+  language: python
+  name: python3
+---
+
+# Construct Design Matrices for Categorical Features
+
+## Design Matrix with `patsy`
```
billbrod, line 16: should be clear here that `patsy` is optional, like scikit learn, but that might be hlpeful

docs/how_to_guide/categorical_predictors.md
```diff
@@ -0,0 +1,120 @@
...
+
+In the design matrix:
+- `T.` indicates **treatment coding**, which is used to encode categorical variables. For example, `stimulus[T.s2]` represents the presence (`1`) or absence (`0`) of the category `s2` compared to the reference category `s1`.
+- The reference category (`s1`) is dropped to avoid collinearity, as including all categories would result in redundancy.
+- Similarly, for `context`, columns `context[c1]` and `T.context[c2]` represent the presence of each category. 
+- The interaction term `stimulus[T.s2]:context[T.c2]` represents the combined effect of `stimulus = s2` and `context = c2`. Only one column is needed for the interaction, as the other combinations are implicitly represented by the reference categories (`s1` and `c1`).
+:::
+- 
+NeMoS GLMs, however, already specify an intercept term, making the first column of this dataframe redundant. Let's drop it.
+
```
billbrod, line 69: this is rendered weirdly -- what's the `-` doing here? I also think this should be part of hte paragraph befor ethe admonition.

I think the admonition should be moved to right after the code box, the title needs to be corrected (it's currently not rendering as the title), and the admonition should probably have a link to the relevant part of the patsy docs: "See patsy docs for more details"

docs/how_to_guide/categorical_predictors.md
```diff
@@ -0,0 +1,120 @@
...
+The feature in the design matrix are 1-hot encoded categories.
+Note that `patsy` adds an intercept and drops the reference `s1`. This is done by design to avoid perfect collinearity (the sum of the `s1` and the `s2` column would be equal to the intercept). 
+
+:::{note} Understanding `patsy`'s Output
+:class: dropdown
+
+In the design matrix:
+- `T.` indicates **treatment coding**, which is used to encode categorical variables. For example, `stimulus[T.s2]` represents the presence (`1`) or absence (`0`) of the category `s2` compared to the reference category `s1`.
+- The reference category (`s1`) is dropped to avoid collinearity, as including all categories would result in redundancy.
+- Similarly, for `context`, columns `context[c1]` and `T.context[c2]` represent the presence of each category.
```
billbrod, line 64:
```suggestion
- Similarly, for `context`, columns `context[c1]` and `T.context[c2]` represent the presence of each category, and `context[T.c2]` is dropped to avoid collinearity. 
```

docs/how_to_guide/categorical_predictors.md
```diff
@@ -0,0 +1,120 @@
...
+Let's assume we have a simple dataset with four samples and two labels assigned to each sample. You can think of these labels as characteristics of the trial to which the samples belong, such as stimulus 1 vs stimulus 2 or context 1 vs context 2.
+
+```{code-cell} ipython3
+# Example data
+import pandas as pd
+from patsy import dmatrix
+import numpy as np
+
+# Example data 
+data = pd.DataFrame({
```
billbrod, line 29: this example is very abstract. I think we'd better off using specific stimulus / context, and pointing out that the general principle applies

docs/how_to_guide/categorical_predictors.md
```diff
@@ -0,0 +1,120 @@
...
+:class: dropdown
+
+In the design matrix:
+- `T.` indicates **treatment coding**, which is used to encode categorical variables. For example, `stimulus[T.s2]` represents the presence (`1`) or absence (`0`) of the category `s2` compared to the reference category `s1`.
+- The reference category (`s1`) is dropped to avoid collinearity, as including all categories would result in redundancy.
+- Similarly, for `context`, columns `context[c1]` and `T.context[c2]` represent the presence of each category. 
+- The interaction term `stimulus[T.s2]:context[T.c2]` represents the combined effect of `stimulus = s2` and `context = c2`. Only one column is needed for the interaction, as the other combinations are implicitly represented by the reference categories (`s1` and `c1`).
+:::
+- 
+NeMoS GLMs, however, already specify an intercept term, making the first column of this dataframe redundant. Let's drop it.
```
billbrod, line 68:
```suggestion
NeMoS GLMs, however, implicitly specify an intercept term, making the first column of this dataframe redundant. Let's drop it.
```

docs/how_to_guide/categorical_predictors.md
```diff
@@ -0,0 +1,120 @@
...
+design_df.drop(columns=["Intercept"], inplace=True)
+
+design_df
+```
+
+
+
+
+### Fit the GLM
+
```
billbrod, line 80: need some text here explaining what's happening

docs/how_to_guide/categorical_predictors.md
```diff
@@ -0,0 +1,120 @@
...
+```{code-cell} ipython3
+import nemos as nmo
+
+# Fit the GLM model
+model = nmo.glm.GLM().fit(design_df, counts)
```
billbrod, line 85: this is a bit confusing, because we don't talk about pandas dataframes anywhere else. I'm assuming it works because we just grab `design_df.values`, the array representation, but let's make that explicit.

docs/how_to_guide/categorical_predictors.md
```diff
@@ -0,0 +1,120 @@
...
+## Categorical Predictors and Basis Functions
+
+Assume you have an additional time series that you want to process with a basis:
+
+```{code-cell} ipython3
+# Example time series
+speed = np.array([10., 3., 2., 20.])
+```
+
+### Composite Basis for Multiple Predictors
```
billbrod, line 97: this header doesn't add anything

docs/how_to_guide/categorical_predictors.md
```diff
@@ -0,0 +1,120 @@
...
+
+You can create a composite basis to combine predictors:
+
+```{code-cell} ipython3
+# Identity basis combined with a B-spline basis
+bas = nmo.basis.IdentityEval() + nmo.basis.BSplineEval(5)
+
+# Compute features
+X = bas.compute_features(design_df, speed)
```
billbrod, line 106: should describe what this does: identity basis just passes through the design_df (again, I think we should add `.values`), while `BSplineEval` works with speed.

and probably link to the bit about additive bases

maybe show the array? or plot it?

docs/how_to_guide/categorical_predictors.md
```diff
@@ -0,0 +1,120 @@
...
+```
+
+### Interaction with Basis algebra
+
+To model the interaction between two variables, you can take advantage of the basis multiplication:
```
billbrod, line 111: this is an additional interaction right? so maybe say something like 
```suggestion
To add the interaction between the `c2` context and speed to your design matrix, you can take advantage of basis multiplication:
```

docs/how_to_guide/categorical_predictors.md
```diff
@@ -0,0 +1,120 @@
...
+
+```{code-cell} ipython3
+# Interaction basis
+bas = nmo.basis.IdentityEval() * nmo.basis.BSplineEval(5)
+
+# Compute features for interaction
+X2 = bas.compute_features(design_df["context[T.c2]"], speed)
+```
```
billbrod, line 119: and again, I think we should visualize this output

docs/how_to_guide/custom_predictors.md
```diff
@@ -0,0 +1,61 @@
...
+    format_version: 0.13
+    jupytext_version: 1.16.4
+kernelspec:
+  display_name: Python 3
+  language: python
+  name: python3
+---
+
+(custom-features)=
+# Custom Features
```
billbrod, line 15: I don't think "custom" is the right word here, but I don't have a better one.

docs/how_to_guide/custom_predictors.md
```diff
@@ -0,0 +1,61 @@
...
+
+## Incorporating Custom Features into GLM Design
+
+In some cases, your data may contain features that are not directly computable through the available bases in NeMoS, for instance a Principal Component Analysis (PCA) of high-dimensional signals. 
+The `IdentityEval` basis allows you to incorporate these precomputed features into the model design, and combine them with other predictors through basis composition.
+
+For example, let's assume that you want to compute a Principal Component Analysis (PCA) of some signals, and use 
+the first 2 Principal Components (PCs) as GLM features.
```
billbrod, line 23:
```suggestion
the first 2 Principal Components as GLM features. 
```

docs/how_to_guide/custom_predictors.md
```diff
@@ -0,0 +1,61 @@
...
+
+## Incorporating Custom Features into GLM Design
+
+In some cases, your data may contain features that are not directly computable through the available bases in NeMoS, for instance a Principal Component Analysis (PCA) of high-dimensional signals. 
+The `IdentityEval` basis allows you to incorporate these precomputed features into the model design, and combine them with other predictors through basis composition.
+
+For example, let's assume that you want to compute a Principal Component Analysis (PCA) of some signals, and use 
+the first 2 Principal Components (PCs) as GLM features. 
+
+Currently, NeMoS doesn't provide a PCA basis, but you can compute the PCs in `sklearn`.
```
billbrod, line 25:
```suggestion
Currently, NeMoS doesn't provide a PCA basis, but you can compute the principle components in `sklearn`.
```

docs/how_to_guide/custom_predictors.md
```diff
@@ -0,0 +1,61 @@
...
+
+# generate some counts
+counts = np.random.poisson(size=n_samples)
+
+# compute the first 2 pcs
+pcs = PCA(2).fit_transform(high_dim_singals)
+
+```
+
+Now, let's see how to use the `IdentityEval` basis to model jointly the PCs and a spike history filter.
```
billbrod, line 45:
```suggestion
Now, let's see how to use the `IdentityEval` basis to model jointly the principle components and a spike history filter.
```

docs/how_to_guide/custom_predictors.md
```diff
@@ -0,0 +1,61 @@
...
+```{code-cell}
+import nemos as nmo
+
+# create a composite basis 
+pc_basis = nmo.basis.IdentityEval()
+history_basis = nmo.basis.RaisedCosineLogConv(3, window_size=10)
+composite_basis = pc_basis + history_basis
+
+# create the model design
+X = composite_basis.compute_features(pcs, counts)
```
billbrod, line 56: I feel like we should visualize this, or at least show the shape?

docs/how_to_guide/raw_history_feature.md
```diff
@@ -0,0 +1,167 @@
...
+
+plt.title("Basis Feature Matrix", fontsize=16)
+plt.imshow(X2[10:], cmap="Greys")
+plt.xticks([1, 4, 7, 10], [0, 1, 2, 3])
+plt.xlabel("Neurons", fontsize=12)
+plt.ylabel("Samples", fontsize=12)
+plt.show()
+```
+
+This model can be fit with the usual `model_reduced = nmo.glm.GLM().fit(X2, counts)`.
```
billbrod, line 93: should be population glm, right?

docs/how_to_guide/raw_history_feature.md
```diff
@@ -0,0 +1,167 @@
...
+mask[1, 0] = 0
+
+# remove the directional coupling from 2 to 3
+mask[2, 3] = 0
+
+# repeat over the rows by the number of basis functions
+mask = np.repeat(mask, basis.n_basis_funcs, axis=0)
+```
+
+We are now readi to fit the GLM masking the coefficients.
```
billbrod, line 142:
```suggestion
We are now ready to fit the GLM masking the coefficients.
```

docs/how_to_guide/raw_history_feature.md
billbrod, line 1: why not do this one using the data from https://nemos--282.org.readthedocs.build/en/282/tutorials/plot_02_head_direction.html? then we can visualize the outputs (and point them to the full tutorial for more details)

billbrod: regardless, I think we should visualize the model coefficients we get

billbrod (changes requested):
It looks like this PR is showing the changes from #280 in the diff, do you know why?

Otherwise, these looks generally good, though I think this means we're going to have to go through the existing how-tos ,etc to see where to use the new bases (at least in the head direction tutorial). Can we make an issue for that?

The new how-tos all looks like they match what they are in diataxis, i.e., very short and answering a single question. I think we'll need to go back through the existing ones and update / move them then, and I think we should consider changing the layout of the landing page for them: make the cards smaller, only show the top-level title, which should be very descriptive. This should also be a separate issue.

src/nemos/basis/_identity.py
BalzaniEdoardo: Yes!

src/nemos/basis/_identity.py
BalzaniEdoardo: yes, to make it read-only

src/nemos/basis/basis.py
BalzaniEdoardo: yep, copy-paste issue

docs/background/basis/README.md
BalzaniEdoardo: My only point here is that history is like a lagged version of the identity, so it is not exactly the "conv" version of identity. 

If it was that, i would expect that the kernel  for convolution would be just a single `np.linsapce(0,1,n_samples)` (which is f(x) = x when x is the linspace).

Instead the history basis is a bunch of delta function that peaks at different lags. 

Also the image in the table of identity doesn't apply to the history so we would not be able to have an image that makes sense for both.

What do you think?

docs/how_to_guide/categorical_predictors.md
BalzaniEdoardo: I added a dropdown admonition

docs/how_to_guide/categorical_predictors.md
BalzaniEdoardo: I tried to make the stim and context more concrete, let me know if it works

docs/how_to_guide/categorical_predictors.md
BalzaniEdoardo: added an admonition about it

BalzaniEdoardo: let me know if it works

docs/how_to_guide/categorical_predictors.md
BalzaniEdoardo: the `.values` is not needed, i think it's extra syntax that they should not care about. If the dataframe contains non-numerical values it will fail.

I'll add a description and plot the content

docs/how_to_guide/categorical_predictors.md
BalzaniEdoardo: I wrapped the output into a DataFrame to add back column names and have a more readable output

docs/how_to_guide/custom_predictors.md
BalzaniEdoardo: Precomputed?

docs/how_to_guide/custom_predictors.md
BalzaniEdoardo: I think i simulated only a 1D array actually

docs/how_to_guide/raw_history_feature.md
BalzaniEdoardo: i think it requires too much pre-processing  for this session. I'll print the coefficients

BalzaniEdoardo: actually, I take it back... i can just grab the first sec or so and use that data

BalzaniEdoardo: also, i would need more explaining about the data, maybe filter a subset of unit and samples... I think it is better to go straight to the point. If we had a sample data which has already a sub-population and a short duration we could load that (like the iris dataset for sklearn)

docs/background/basis/README.md
billbrod: I think that makes sense.

docs/how_to_guide/categorical_predictors.md
billbrod: I think this is good!

docs/how_to_guide/categorical_predictors.md
billbrod: added some suggetions

docs/how_to_guide/categorical_predictors.md
```diff
@@ -0,0 +1,163 @@
...
+import nemos as nmo
+
+# Fit the GLM model
+model = nmo.glm.GLM().fit(design_df, counts)
+```
+:::{note}
+
+You can directly pass a [`pandas.DataFrame`](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.html) to the 
+`fit` method, as long as it can be internally converted to a floating-point array. The conversion is handled automatically, 
+so no additional steps are required on your part.
```
billbrod, line 115:
```suggestion
You can directly pass a [`pandas.DataFrame`](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.html) to the 
`fit` method, as long as it can be internally converted to a floating-point array (that means, for example, that you can't have a column whose values are strings). The conversion is handled automatically, by calling `design_df.values`,
so no additional steps are required on your part.
```

docs/how_to_guide/categorical_predictors.md
billbrod: but it will fail if they try to do use a dataframe with non-numeric values anyway, right?

billbrod: I actually don't think we should display it as a dataframe -- we should show what the output looks like: an appropriately shaped array.

docs/how_to_guide/categorical_predictors.md
billbrod: same point about not showing it as a dataframe. just show it as an array and explain what the columns mean in words

docs/how_to_guide/raw_history_feature.md
billbrod: Okay, I think this is fine, but it would be nice to think about whether there is a meaningful dataset we could use. because it's nice to plot the coefficients and have them look meaningful.

billbrod (approved):

docs/how_to_guide/raw_history_feature.md
BalzaniEdoardo: you are right

BalzaniEdoardo:
> ?



> It looks like this PR is showing the changes from #280 in the diff, do you know why?
No idea, I might have branched from `set_shape_basis_method` and this confused the history. I am not sure it makes sense

> Otherwise, these looks generally good, though I think this means we're going to have to go through the existing how-tos ,etc to see where to use them. Can we make an issue for that?
done

> The new how-tos all looks like they match what they are in diataxis, i.e., very short and answering a single question. I think we'll need to go back through the existing ones and update / move them then, and I think we should consider changing the layout of the landing page for them: make the cards smaller, only show the top-level title, which should be very descriptive. This should also be a separate issue.

I agree with both points

PR #286, opened 2025-01-09
Adds __repr__ for our custom classes

billbrod:
This adds a `__repr__` method for our custom classes: basis, glm, observation models, and regularizers. The logic for almost all of these are included in the new `nmo.utils.format_repr` function, summarized as:

    - The returned string includes the output of ``obj.get_params(deep=False)`` if value
      (that is, if it's not False, None, empty, etc.).

    - Any key in ``exclude_keys`` are excluded.

    - If the key is in ``use_name_keys``, we use ``value.__name__``. Else we use
      ``value.__repr__``

TransformerBasis and CompositeBasis are different:
- TransformerBasis is `Transformer(self.basis)`
- CompositeBasis is

```
AdditiveBasis(
    basis1=self.basis1,
    basis2=self.basis2,
)
```

and same for Multiplicative, with composites of composites using additional tab indenting


src/nemos/utils.py
```diff
@@ -459,6 +460,32 @@ def assert_scalar_func(func: Callable, inputs: List[jnp.ndarray], func_name: str
...
+    - Any key in ``exclude_keys`` are excluded.
+
+    - If the key is in ``use_name_keys``, we use ``value.__name__``. Else we use
+      ``value.__repr__``
+
+    """
+    init_params = list(inspect.signature(obj.__init__).parameters.keys())
+    disp_params = []
+    for k, v in obj.get_params(deep=False).items():
+        if k not in exclude_keys and v:
```
BalzaniEdoardo, line 480: This doesn't work if v is an array (like `decay_rates` for `OrthExponentialBasis`), we must generalize this to handle array values

src/nemos/utils.py
```diff
@@ -459,6 +460,32 @@ def assert_scalar_func(func: Callable, inputs: List[jnp.ndarray], func_name: str
         )
 
 
+def format_repr(
+    obj: object, exclude_keys: List[str] = [], use_name_keys: List[str] = []
```
BalzaniEdoardo, line 464:
```suggestion
    obj: Base, exclude_keys: List[str] = [], use_name_keys: List[str] = []
```

and add `from .base_class import Base`.

src/nemos/utils.py
```diff
@@ -459,6 +460,32 @@ def assert_scalar_func(func: Callable, inputs: List[jnp.ndarray], func_name: str
...
+):
+    """Format object representation string (__repr__).
```
BalzaniEdoardo, line 466: can you format this in numpydoc?

BalzaniEdoardo (changes requested):
thanks for the first pass at this!

I added a couple of comments. I would include a max depth or something similar for additive bases in the repr.

BalzaniEdoardo:
## Added Features
- Slightly changed the `format_repr` utility function:
    - Ignore parameters that have the attribute `shape` (intended for arrays).
    - Do not ignore parameters with value `False` and `0`, while ignore other falsey values like `None` or `""`.
- Added basic repr tests.
- Doctests for basis.

> Note: I did not add doctests for GLM repr because the output is way too long, but I have included actual tests.

src/nemos/glm.py
```diff
@@ -1116,6 +1117,9 @@ def _get_optimal_solver_params_config(self):
         """Return the functions for computing default step and batch size for the solver."""
         return glm_compute_optimal_stepsize_configs(self)
 
+    def __repr__(self):
+        return format_repr(self, ["feature_mask"])
```
billbrod, line 1121:
```suggestion
        return format_repr(self)
```
Since format_repr now doesn't show arrays

src/nemos/utils.py
```diff
@@ -459,6 +461,75 @@ def assert_scalar_func(func: Callable, inputs: List[jnp.ndarray], func_name: str
         )
 
 
+def format_repr(
+    obj: Base, exclude_keys: Optional[List[str]] = None, use_name_keys: List[str] = []
```
billbrod, line 465: why is `exclude_keys=None`, but `use_name_keys=[]`?

billbrod (commented):
Should probably add a doctest for at least the GLM (if not also regularizer and observation models?)

There are other notes scattered throughout

billbrod:
Most recent change adds label to the repr. This only affects basis right now (others don't have a label param). Questions about that:
- We don't handle `label=""` (which user would have to do themselves) in a special way, so you'd get back e.g., `(RaisedCosineLinearEval, width=2.0, ...)`. (With `label=None` and if the user doesn't set it directly, you'd get `RaisedCosineLinearEval(width=2.0, ...)`)
- I didn't add the label behavior to any doctests, because I didn't see it being used in any existing doctests. Should I?
- User can't set label for Additive and Multiplicative basis, so I don't handle it specifically there. But they do have a more informative label -- should I make use of it?
- With this, I don't think we're using exclude_keys arg anywhere (it's being used in for the GLM, but as I noted in my comments, I think that can be removed) -- should we keep it in case we need it later or remove it?

src/nemos/basis/_basis_mixin.py
```diff
@@ -644,3 +644,18 @@ def set_input_shape(self, *xi: int | tuple[int, ...] | NDArray) -> Basis:
             )._n_basis_input_,
         )
         return self
+
+    def __repr__(self, n=0):
```
sjvenditto, line 648:
```suggestion
    def __repr__(self, n=0):
         # number of rows in the terminal, reduced by 4 for multiline printing of a composite basis
         from pynapple.core.utils import _get_terminal_size
         _, rows = get_terminal_size()
         rows = rows // 4
```

src/nemos/basis/_basis_mixin.py
```diff
@@ -644,3 +644,18 @@ def set_input_shape(self, *xi: int | tuple[int, ...] | NDArray) -> Basis:
...
+        tab = "    "
+        try:
+            basis1 = self.basis1.__repr__(n=n)
+        except TypeError:
+            basis1 = self.basis1
+        try:
+            basis2 = self.basis2.__repr__(n=n)
+        except TypeError:
+            basis2 = self.basis2
+        rep = f"{self.__class__.__name__}(\n{n*tab}basis1={basis1},\n{n*tab}basis2={basis2},\n{(n-1)*tab})"
```
sjvenditto, line 660:
```suggestion
        if n < rows:
             rep = f"{self.__class__.__name__}(\n{n*tab}basis1={basis1},\n{n*tab}basis2={basis2},\n{(n-1)*tab})"
        elif n==rows:
             rep = f"{self.__class__.__name__}(\n{n*tab}...\n{(n-1)*tab})"
        else:
             rep = None
```

sjvenditto (commented):
I've added suggestions for the composite basis repr to have a maximum print length based on the terminal size. The line divisor might be too big (i.e. the print underfills the terminal), but here's an example of a line limited repr in my terminal:
image

src/nemos/utils.py
BalzaniEdoardo: actually, none of them should be initialized to a mutable

BalzaniEdoardo: I'll fix it

PR #289, opened 2025-01-10
fix split_by_feature bug

billbrod:
Small bugfix: in `split_by_feature` for the additive basis, jax tree map was sorting the dictionary that was being used to be alphabetical with respect to the labels of `basis1` and `basis2`. Thus, if those two basis objects had different `n_basis_input` values and were passed in using a different order than their alphabetical sorting, `split_by_feature` would fail. This fixes that by using an OrderedDict

It doesn't look like this was an issue for the `MultiplicativeBasis`, but I added that test anyway, let me know if you want me to remove it.

### DIFF
--- a/src/nemos/basis/_basis.py
+++ b/src/nemos/basis/_basis.py
@@ -694,7 +694,7 @@ class Basis(Base, abc.ABC, BasisTransformerMixin):
         # Apply the slicing using the custom leaf function
         out = jax.tree_util.tree_map(lambda sl: x[sl], index_dict, is_leaf=is_leaf)
 
-        # reshape the arrays to spilt by n_basis_input_
+        # reshape the arrays to match input shapes
         reshaped_out = dict()
         for items, bas in zip(out.items(), self):
             key, val = items
--- a/src/nemos/basis/_basis.py
+++ b/src/nemos/basis/_basis.py
@@ -1010,8 +1010,8 @@ class AdditiveBasis(CompositeBasisMixin, Basis):
         >>> split_coef = basis.split_by_feature(model.coef_, axis=0)
         >>> for feature, coef in split_coef.items():
         ...     print(f"{feature}: shape {coef.shape}")
-        feature_1: shape (1, 5)
-        feature_2: shape (1, 6)
+        feature_1: shape (5,)
+        feature_2: shape (6,)
 
         """
         return super().split_by_feature(x, axis=axis)
@@ -1358,7 +1358,7 @@ class MultiplicativeBasis(CompositeBasisMixin, Basis):
         >>> split_features = basis_mul.split_by_feature(X_multi, axis=1)
         >>> for feature, arr in split_features.items():
         ...     print(f"{feature}: shape {arr.shape}")
-        (one_input * two_inputs): shape (20, 2, 60)
+        (one_input * two_inputs): shape (20, 2, 30)
 
         """
         return super().split_by_feature(x, axis=axis)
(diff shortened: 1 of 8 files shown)


tests/test_basis.py
```diff
@@ -304,6 +304,20 @@ def test_expected_output_split_by_feature(basis_instance, super_class):
         np.testing.assert_array_equal(xx[~nans], x[~nans])
 
 
+@pytest.mark.parametrize("composite_op", ["add", "multiply"])
```
BalzaniEdoardo, line 307: This looks good, can you parametrize a few input shapes? Like (n,)  (n,1) (n, 2), (n,1,2)

BalzaniEdoardo (changes requested):
Great catch, if you add a couple of different input shapes in the test,  this is good to go for me!

BalzaniEdoardo:
## Changed in this PR:
- `split_by_feature`: the method splits the feature axis using the shape of the provided input. Example, if a basis with 5 elements processed an input of shape `(n_samples, 1, 2 ,3)`, the feature axis, which will be of length `1*2*3*5` will be reshaped to `(1, 2, 3, 5)`. 
- `__iter__`: method for iterating over the additive components, which usually corresponds to different task variables.
- `__len__`: returns the number of components.

src/nemos/basis/_basis.py
```diff
@@ -958,6 +962,7 @@ def split_by_feature(
             .. math::
                 (..., n_i, b_i, ...)
 
+            - ``n_i``: The number of inputs processed by the i-th basis component.
```
billbrod, line 965:
```suggestion
```
duplicate

src/nemos/basis/_basis.py
```diff
@@ -981,8 +986,8 @@ def split_by_feature(
         >>> split_features = basis.split_by_feature(X, axis=1)
         >>> for feature, arr in split_features.items():
         ...     print(f"{feature}: shape {arr.shape}")
-        feature_1: shape (20, 1, 5)
-        feature_2: shape (20, 1, 6)
+        feature_1: shape (20, 5)
+        feature_2: shape (20, 6)
         >>> # If one of the basis components accepts multiple inputs, the resulting dictionary will be nested:
```
billbrod, line 991:
```suggestion
        >>> # split_by_features can handle multiple inputs as well
```
I think this line is out of date, right?

billbrod (commented):
Can you write out the attribute renaming that you did? Otherwise this looks good to me!

billbrod:
To be explicit: this PR changes the behavior of `split_by_feature`. Previously, it would always return a 3d array.

Now, `basis.input_shape` is set when `compute_features` or `set_input_shape` is called on a Nd array `X`, storing `X.shape[1:]` (it can thus be empty, if `X` is 1d). Then, `basis.split_by_feature(X)` returns an array of shape `(n_samples, *basis.input_shape, basis.n_basis_funcs)`:

```python
n_samples = 100
basis = nmo.basis.RaisedCosineLinearEval(7)
input_shape = (4, 5, 10) # or some other tuple of ints
X = np.random.rand(n_samples, *input_shape)
split = basis.split_by_feature(b.compute_features(X))
split[basis.label].shape
>>> (100, 4, 5, 10, 7)
```

BalzaniEdoardo:
Additional changes to `_basis.py`:
- Renamed `_n_basis_input_` to '_input_shape_product`, a more descriptive name.
- Removed the property`n_basis_input_` since the attribute is used internally only.

PR #291, opened 2025-01-14
PopulationGLM sklearn clone

billbrod:
This PR adds a `__sklearn_clone__` method to  PopulationGLM, which removes the `feature_mask` attribute for reinitializing. This avoids a bug where if `PopulationGLM` was part of a pipeline with transformer basis, was fit (so that `feature_mask` was initialized) and then cross-validation was attempted over the number of basis funcs, errors would be raised:

```python
x = np.random.rand(100)
y = np.random.rand(2)
bas = CyclicBSplineEval(5)
bas.compute_features(x)
bas = TransformerBasis(bas)
pipe = Pipeline([("basis", bas), ("model", nmo.glm.PopulationGLM())])
pipe.fit(x, y)
cv = GridSearchCV(pipe, {"basis__n_basis_funcs": [2, 5]}, cv=3)
cv.fit(x, y)
```

### DIFF
--- a/src/nemos/glm.py
+++ b/src/nemos/glm.py
@@ -1618,8 +1618,7 @@ class PopulationGLM(GLM):
         )
 
     def __sklearn_clone__(self) -> GLM:
-        """Clone the PopulationGLM, dropping feature_mask
-        """
+        """Clone the PopulationGLM, dropping feature_mask"""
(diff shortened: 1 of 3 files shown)


BalzaniEdoardo (approved):
good to go for me

PR #297, opened 2025-01-21
set version dynamically

billbrod:
With this change, `setuptools_scm` will set the installed package version dynamically, based on the git tag. This is the same way I handle it for plenoptic.


BalzaniEdoardo (approved):
Great, thanks Billy!

PR #306, opened 2025-02-10
Label get param

BalzaniEdoardo:
This PR addresses the following

1. Add a `__getitem__` to `Basis`. Labels can be used to get the desired basis in the tree. 
2. Improved label logic: 
    - User provided labels in composite basis must be unique.
    - Default labels in composite bases (class name) is disambiguated by appending `_<number>`.
    - Labeling a basis using another basis type class name results in an error.
3. `get_params` and `set_params` override default behavior: instead of reflecting the full attribute structure, `basis1__basis2...__parameter_name`, the parameter key is set to `<label>_parameter_name`.

For a concrete example, refer to the new how-to note: `handling_composite_bases.md`

## Key Implementation: Setter of Basis

For the PR review, focus on the behavior of `@basis1.setter` and `@basis2.setter`. This property determines how  the `set_params` is finalized. In particular, in case one uses `set_params` to set a basis funciton, some extra attribute must be modfied:

- If one sets a basis with `set_params`, the new gets the `_parent` attribute from the current.
- Same thing for input shape
- And try to also grab the current label, if possible.


docs/conf.py
```diff
@@ -6,18 +6,22 @@
 # -- Project information -----------------------------------------------------
 # https://www.sphinx-doc.org/en/master/usage/configuration.html#project-information
 
-import nemos
 import sys, os
 from pathlib import Path
 
+from importlib.metadata import version
```
billbrod, line 12: I think this bit only shows up here because I messed up: it's from https://github.com/flatironinstitute/nemos/pull/308, which was merged into `main`, not `development`. We should merge `main` back into `development`, so this goes away.

docs/how_to_guide/handling_composite_bases.md
```diff
@@ -0,0 +1,156 @@
...
+```
+
+:::{admonition} Grid definition
+:class: info
+
+The parameter keys retrieved by `get_params` are the one needed to define a parameter grid when cross-validating your hyper-parameters with scikit-learn. Learn how to cross-validate basis parameters using [pipelines](https://scikit-learn.org/stable/modules/generated/sklearn.pipeline.Pipeline.html) with [this notebook](sklearn-how-to). 
+:::
+
+
+Note that if you don't provide a label, basis class name is used to construct the keys.
```
billbrod, line 138: From here to the end belongs in the second section, I think. it's not about sklearn

docs/how_to_guide/handling_composite_bases.md
```diff
@@ -0,0 +1,156 @@
...
+```
+
+This change is reflected on the composite basis.
+
+```{code-cell} ipython3
+# check that the input2 basis has now 8 basis funcs
+add
+```
+
+### Cross-validating Composite Basis Parameters
```
billbrod, line 116: Needs a different title: this section is actually about interacting with them in a sklearn-type way, using `set_params()`, so title should reflect that.

billbrod: and then there should be a comment saying something like "this is how sklearn interacts with basis objects, and so enables cross-validation, etc"

docs/how_to_guide/plot_02_glm_demo.md
```diff
@@ -457,7 +441,10 @@ if path.exists():
...
-
 ## References
 (ref-1)=
-[1] [Arribas, Diego, Yuan Zhao, and Il Memming Park. "Rescuing neural spike train models from bad MLE." Advances in Neural Information Processing Systems 33 (2020): 2293-2303.](https://arxiv.org/abs/2010.12362)
\ No newline at end of file
+[1] [Arribas, Diego, Yuan Zhao, and Il Memming Park. "Rescuing neural spike train models from bad MLE." Advances in Neural Information Processing Systems 33 (2020): 2293-2303.](https://arxiv.org/abs/2010.12362)
+
+```{code-cell} ipython3
+
+```
```
billbrod, line 450:
```suggestion
```

docs/how_to_guide/plot_03_population_glm.md
```diff
@@ -267,3 +258,7 @@ model_tree.fit(pytree_features, spikes)
 # print the coefficients
 print(model_tree.coef_)
 ```
+
+```{code-cell} ipython3
+
+```
```
billbrod, line 264:
```suggestion
```

docs/scripts/glm_predictors.py
```diff
@@ -97,7 +97,7 @@ def plot_custom_features():
 
     # Plot stacked signals
     plt.figure(figsize=(12, 4))
-    cmap = plt.cm.get_cmap("rainbow")
+    cmap = plt.colormaps["rainbow"]
```
billbrod, line 100: What's this difference?

docs/tutorials/plot_03_grid_cells.md
```diff
@@ -175,27 +167,24 @@ plt.tight_layout()
...
 The shape is (n_samples, n_basis). This means that for each time point "t", we evaluated the basis at the
 corresponding position. Let's plot 5 time steps.
 
-
 ```{code-cell} ipython3
 fig = plt.figure(figsize=(12, 4))
 gs = plt.GridSpec(2, 5)
 xt = np.arange(0, 1000, 200)
-cmap = plt.get_cmap("rainbow")
+cmap = plt.colormaps["rainbow"]
```
billbrod, line 187: same colormap question

pyproject.toml
```diff
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "nemos"
-version = "0.2.1"
+dynamic = ["version"]
```
billbrod, line 7: same point about https://github.com/flatironinstitute/nemos/pull/308

src/nemos/__init__.py
```diff
@@ -1,5 +1,6 @@
 #!/usr/bin/env python3
-__version__ = "0.2.1"
+from importlib.metadata import PackageNotFoundError as _PackageNotFoundError
```
billbrod, line 2: Same point about https://github.com/flatironinstitute/nemos/pull/308

src/nemos/_documentation_utils/plotting.py
```diff
@@ -204,7 +204,7 @@ def current_injection_plot(
     # define plotting parameters
     # colormap, color levels and transparency level
     # for the current injection epochs
-    cmap = plt.get_cmap("autumn")
+    cmap = plt.colormaps["autumn"]
```
billbrod, line 207: same question about colormaps throughout this file

src/nemos/base_class.py
```diff
@@ -69,7 +69,8 @@ def set_params(self, **params: Any):
         if not params:
             # Simple optimization to gain speed (inspect is slow)
             return self
-        valid_params = self.get_params(deep=True)
+        get_params = getattr(self, "__sklearn_get_params__", self.get_params)
```
billbrod, line 72: why are there two different functions for this?

src/nemos/basis/_transformer_basis.py
```diff
@@ -82,6 +82,11 @@ class TransformerBasis:
     )
 
     def __init__(self, basis: Basis):
+        if not hasattr(basis, "get_params") or not hasattr(basis, "compute_features"):
+            raise TypeError(
+                "Provide an object of type `Basis` to initialize a TransformerBasis. "
+                f"Object of type {type(basis)} provided instead!"
```
billbrod, line 88: this error message is confusing -- it's not actually the type that's the issue but the attributes. I'd say that directly (more helpful for someone trying to write their own), mentioning that all nemos basis objects satisfy this criterion

src/nemos/convolve.py
```diff
@@ -90,6 +90,8 @@ def _shift_time_axis_and_convolve(array: NDArray, eval_basis: NDArray, axis: int
     -----
     This function supports arrays of any dimensionality greater or equal than 1.
     """
+    # convert axis
+    axis = axis if axis >= 0 else array.ndim + axis
```
billbrod, line 94: is this part of this PR?

src/nemos/utils.py
```diff
@@ -156,6 +156,8 @@ def _pad_dimension(
         "acausal": ((pad_size) // 2, pad_size - (pad_size) // 2),
         "anti-causal": (0, pad_size),
     }
+    # convert negative axis in jax jit compilable way
+    axis = axis * (axis >= 0) + (array.ndim + axis) * (axis < 0)
```
billbrod, line 160: is this part of this PR?

src/nemos/utils.py
```diff
@@ -525,8 +527,9 @@ def format_repr(
 
     init_params = list(inspect.signature(obj.__init__).parameters.keys())
     disp_params = []
-
-    all_params = obj.get_params(deep=False)
+    # use special method for basis
+    get_params = getattr(obj, "__sklearn_get_params__", obj.get_params)
```
billbrod, line 531: same question: why are there two get_params funcs?

src/nemos/basis/_basis_mixin.py
```diff
@@ -73,15 +94,135 @@ def wrapper(self, *args, **kwargs):
...
+    # add a parent to the new basis
+    new._parent = getattr(current, "_parent", None)
+
+    # Carry-on input shape info if dimensions match
+    for attr in ("_input_shape_product", "_input_shape_"):
+        if (
+            getattr(new, attr, None) is None
+            and new._n_input_dimensionality == current._n_input_dimensionality
+        ):
+            setattr(new, attr, getattr(current, attr, None))
```
billbrod, line 115: does this belong in this PR?

src/nemos/basis/_basis.py
```diff
@@ -135,24 +186,144 @@ class Basis(Base, abc.ABC, BasisTransformerMixin):
...
+            out[key] = value
+        return out
+
+    def _get_params_and_key_map(self, deep=True) -> Tuple[dict, dict]:
+        """
+        From scikit-learn, get parameters by inspecting init.
+
+        Parameters
+        ----------
+        deep
```
billbrod, line 265: missing description

src/nemos/basis/_basis.py
```diff
@@ -570,13 +756,13 @@ def _get_feature_slicing(
         return split_dict, start_slice
 
     @staticmethod
-    def _generate_unique_key(existing_dict: dict, key: str) -> str:
+    def _generate_unique_key(existing_dict: dict | List | Tuple, key: str) -> str:
         """Generate a unique key if there is a conflict."""
         extra = 1
-        new_key = f"{key}-{extra}"
+        new_key = f"{key}_{extra}"
```
billbrod, line 762: Didn't we decide that `-` was better than `_`? because `__` gets used by sklearn and `_` is used in our attributes

src/nemos/basis/_basis_mixin.py
```diff
@@ -507,28 +657,151 @@ class CompositeBasisMixin:
...
+
+    @property
+    def basis1(self):
+        return self._basis1
+
+    @basis1.setter
+    def basis1(self, basis):
+        if not hasattr(basis, "get_params") or not hasattr(basis, "compute_features"):
+            raise TypeError(
+                f"`basis1` must be an object of type `Basis`. Type {type(basis)} provided instead."
```
billbrod, line 687: same point about type vs. attribute here

src/nemos/basis/_basis_mixin.py
```diff
@@ -507,28 +657,151 @@ class CompositeBasisMixin:
...
+
+    @property
+    def basis2(self):
+        return self._basis2
+
+    @basis2.setter
+    def basis2(self, basis):
+        if not hasattr(basis, "get_params") or not hasattr(basis, "compute_features"):
+            raise TypeError(
+                f"`basis2` must be an object of type `Basis`. Type {type(basis)} provided instead."
```
billbrod, line 705: same point about type vs. attribute here

src/nemos/basis/_basis.py
```diff
@@ -524,6 +697,19 @@ def __pow__(self, exponent: int) -> MultiplicativeBasis:
...
+    def __getitem__(self, index: str) -> Basis:
+
+        search = next(
+            (bas for lab, bas in generate_basis_label_pair(self) if lab == index), None
+        )
+
+        if search is None:
+            avail_index = ",".join(f"'{b}'" for b in self._generate_subtree_labels())
+            raise IndexError(
+                f"Basis label {index} not found. Available labels: {avail_index}"
```
billbrod, line 709: maybe have a different error message if `index` is an int? because then someone just assumed they can do index-based getting, whereas we only support label-based getting.

billbrod (changes requested):
I played around with this a bit locally, and I think it behaves as expected.

In addition to the comments:
- We need some linting to happen on the notebooks: we're getting a lot of changes that I think are just coming from differences in how editor vs. jupyter lab is saving them, which is distracting
- I want to draw particular attention to my comment in `basis/__init__.py` on `__all__`, something is weird there
- I don't understand what's going on with `sklearn_get_params` / `get_params` and with `_map_parameters` / `remap_parameters` -- can you explain? Basically, I think I want a description that's a bit more detailed than the high-level one in the PR now. 
- We only support `__getitem__` for labels, not `__getattr__`, which is fine, but do we want to raise a more specific error message if someone tries to grab an attribute using something we know is a label? if they use attributes, they need to do `basis1.basis2.etc` but getitem allows for labels. Probably worth explaining that in the new doc page?

docs/scripts/glm_predictors.py
BalzaniEdoardo: One is getting deprecated

docs/tutorials/plot_03_grid_cells.md
BalzaniEdoardo: responded above

docs/tutorials/plot_06_calcium_imaging.md
```diff
@@ -178,18 +168,16 @@ We can define a cyclic-BSpline for capturing the encoding of the heading angle,
...
+heading_basis = nmo.basis.CyclicBSplineEval(n_basis_funcs=12, label="heading")
+coupling_basis = nmo.basis.RaisedCosineLogConv(3, window_size=10, label="coupling")
 ```
 
-We need to make sure the design matrix will be full-rank by applying identifiability constraints to the Cyclic Bspline, and then combine the bases (the resturned object will be an [`AdditiveBasis`](nemos.basis._basis.AdditiveBasis) object).
-
+Let's combine both basis into a single additive element.
 
 ```{code-cell} ipython3
-heading_basis.identifiability_constraints = True
```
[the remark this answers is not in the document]
BalzaniEdoardo: this is an old attribute, now we moved the identifiability constraint logic outside the basis, so the flag does nothing at this point

src/nemos/base_class.py
BalzaniEdoardo: `set_params` needs to rely on the actual attribute structure. If one is calling a basis, then `get_params` would return the dictionary with keys the attributes of the basis, the `__sklearn_get_params__` returns the actual structure for the keys "basis1_basis2_...." etc.

BalzaniEdoardo:
From last version of the PR, many things got moved around after talking with @billbrod.

## Composite basis tree-structure traversing
Many operation on composite basis require updating child bases. So far, most of the machinery relied on the child basis methods and properties. This is included the label setter property implemented in this PR. 

With this revised PR, I moved all the label editing logic, as well as other functions that operate over the whole basis tree, in a dedicated module. This functions are now independent from the child bases internal machinery. This is independence is not tested here but will be the object of a dedicated PR in which I will add a mock basis class with a single method, the compute_features, and make sure that I can successfully add it to `nemos` basis and that the overall machinery keeps working.

## Get Params Logic
The overwritten `get_params` `set_params` logic has been moved to the `CompositeBasis`, which is the only basis that needs special machinery. Atomic basis rely on the standard sklearn machinery.

### Implementation details

1. `__sklearn_get_params__`: implements the standard sklearn parameter extraction recursion which relies on class signature. Needed for `Base.set_params` to work correctly, otherwise signature names and key structure of the parameter dict generated would mismatch.
2. `_get_params_and_key_map` recursion similar to `__sklearn_get_params__` but that re-labels the key of the parameter dictionary using `basis.label`, and additionally stores a `key_map` dictionary mapping the new keys to the standard structure based ones.
4. `_map_params`: this one calls `_get_params_and_key_map` and then drops the label from the key of the parameters of `self` (which is the top level bases), returns two dictionaries: the one storing the parameter and the one with the key mapping.
5. `get_params`: calls `_map_params` and return only the dictionaries with the parameters.
6. `remap_params`: decorator that wraps composite basis set_params. **pre-processing** It receives a parameter dictionary with either the keys based on attributes (if `set_params` is called internally by sklearn or `Base`) or the re-mapped  dictionary that uses labels, and maps it back to the standard sklearn attribute structure-based  dictionary. **processing:** calls the `nemos.Base`  `set_params` on the re-formatted dictionary. **post-processing:**
 re-assign labels.

BalzaniEdoardo (commented):
worked on the revisions! ready for a second round

billbrod (changes requested):
Some small changes throughout, but two bigger ones I wanted to mention:
- The new docs page should show that, when you have nested basis objects, you can use the label to directly grab the basis you want, regardless of how deeply nested (and this is what making sure the labels are unique gets us, so it should be placed right after we mentioned that).
- I thought I had made a comment in `basis/__init__.py` about this, but now I'm not seeing it, so maybe it didn't go through. There's something weird with that `__all__` behavior. Let's talk about it in person, because it's a bit hard to write up.

billbrod:
> I thought I had made a comment in `basis/__init__.py` about this, but now I'm not seeing it, so maybe it didn't go through. There's something weird with that `__all__` behavior. Let's talk about it in person, because it's a bit hard to write up.

I think I pushed a fix for this. It was smaller than I thought: basically `basis/__init__.py` should import the `__PUBLIC_BASES__` list directly and use that for its all/dir behavior. Then we have a test that checks whether the imported objects which are subclasses of `Basis` matches `__all__`. Finding out what was imported without using `__all__` was more complicated than I thought, I ended up using `nemos.basis.__dict__`. There might be a better way using `inspect`.

src/nemos/basis/_basis_mixin.py
```diff
@@ -1036,3 +996,138 @@ def update_default_label_id(basis1: Basis, basis2: Basis):
                 count_labels[cls_name] = 1 + current
 
         return
+
+    def __sklearn_get_params__(self, deep=True):
```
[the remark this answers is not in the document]
BalzaniEdoardo: we cannot really, because within this method we are calling recursively `__sklearn_get_params__` itself to get the nested parameters. If you call `Base.get_params`, this will work only for the current level, for any nesting it will call `get_params` of the child object, which in turns will return the dictionary with keys based on the labels instead of the one we need.

billbrod (approved):

BalzaniEdoardo:
Thanks @billbrod and @sjvenditto for the revisions

PR #307, opened 2025-02-12
Speed up recursions

BalzaniEdoardo:
In this PR:
- New accept integer for `__mul__`: `basis * 3` will be equivalent to `basis + basis+basis`
- Add an `__rmul__` enabling right multiplication`n * basis`
- Improved the efficiency of `__pow__` by bisect multiply (log2(n) multiplication, less deep copy)
- Add a mechanism that do not perform a deep-copy when cloning a bases

**Old clone performance**
```
bas = nmo.basis.BSplineEval(5)**500
%timeit bas.__sklearn_clone__()
2.47 s ± 107 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

**New clone performance**
```
bas = nmo.basis.BSplineEval(5)**500
%timeit bas.__sklearn_clone__()
10.9 ms ± 22 μs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```
**Old Power Performance**
```
bas = nmo.basis.BSplineEval(5)
%timeit bas ** 500
2.49 s ± 76.7 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

**New Power Performance**
```
bas = nmo.basis.BSplineEval(5)
%timeit bas ** 500
8.46 ms ± 79.3 μs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

### DIFF
--- a/src/nemos/basis/_basis.py
+++ b/src/nemos/basis/_basis.py
@@ -20,6 +20,23 @@ from ..validation import check_fraction_valid_samples
 from ._basis_mixin import BasisTransformerMixin, CompositeBasisMixin
 
 
+def _bisect_mul(base: Basis, mul: int):
+    """Speed up adding n basis of the same type"""
+    if mul == 1:
+        return base  # Base case
+    half = _bisect_mul(base, mul // 2)
+    sum = half + half  # Less deep copying
+    return sum + base if mul % 2 else sum
+
+
+def _bisect_power(base: Basis, exp: int):
+    if exp == 1:
+        return base  # Base case
+    half = _bisect_power(base, exp // 2)
+    squared = half * half  # Less deep copying
+    return squared * base if exp % 2 else squared
+
+
 def add_docstring(method_name, cls):
     """Prepend super-class docstrings."""
     attr = getattr(cls, method_name, None)
@@ -489,14 +506,15 @@ class Basis(Base, abc.ABC, BasisTransformerMixin):
             The resulting Basis object.
         """
         if isinstance(other, int):
-            def add(base, mul):
-                if mul == 1:
-                    return base  # Base case
-                half = add(base, mul // 2)
-                sum = half + half  # Less deep copying
-                return sum + base if mul % 2 else sum
-            return add(self, other)
-
+            if other <= 0:
+                raise ValueError(
+                    f"Basis multiplication error. Integer multiplicative factor must be positive, {other} provided instead."
+                )
+            return _bisect_mul(self, other)
+        if not isinstance(other, Basis):
+            raise TypeError(
+                "Basis multiplicative factor should be a Basis object or a positive integer!"
+            )
         return MultiplicativeBasis(self, other)
 
     def __pow__(self, exponent: int) -> MultiplicativeBasis:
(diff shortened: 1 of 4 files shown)


BalzaniEdoardo:
Thanks @billbrod and @sjvenditto  for the revisions!

src/nemos/basis/_basis.py
```diff
@@ -506,15 +536,20 @@ def __pow__(self, exponent: int) -> MultiplicativeBasis:
...
+            return deepcopy(self)
 
-        result = self
-        for _ in range(exponent - 1):
-            result = result * self
-        return result
+        mul = MultiplicativeBasis(self, self)
+        with mul._set_shallow_copy_temporarily(True):
+            for _ in range(2, exponent):
+                mul = MultiplicativeBasis(mul, deepcopy(self))
```
billbrod, line 551:
```suggestion
                mul *= deepcopy(self)
```
optional, but I think it's more readable and roughly the same speed (whether it was faster or slower fluctuated from test to test)

src/nemos/basis/_basis.py
```diff
@@ -480,9 +484,35 @@ def __mul__(self, other: Basis) -> MultiplicativeBasis:
...
             The resulting Basis object.
         """
+        if isinstance(other, int):
+            if other <= 0:
+                raise ValueError(
+                    "Basis multiplication error. Integer multiplicative factor must be positive, "
+                    f"{other} provided instead."
+                )
+            elif not all(
+                b._has_default_label for _, b in generate_basis_label_pair(self)
```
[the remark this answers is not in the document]
billbrod: Answer: because we need the labels to be unique, so user just needs to set them after calling `mul`.

src/nemos/basis/_basis.py
```diff
@@ -480,9 +484,35 @@ def __mul__(self, other: Basis) -> MultiplicativeBasis:
...
+            ):
+                raise ValueError(
+                    "Cannot multiply by an integer a basis including a user-defined labels."
```
billbrod, line 497:
```suggestion
                    "Cannot multiply by an integer a basis including a user-defined labels (because then they won't be unique). Set labels after multiplication."
```

src/nemos/basis/_basis_mixin.py
```diff
@@ -852,6 +860,16 @@ def _iterate_over_components(self):
             components2,
         )
 
+    @contextmanager
+    def _set_shallow_copy_temporarily(self, value):
+        """Context manger for setting the shallow copy flag in a thread safe way."""
+        old_value = self.__class__._shallow_copy
+        self.__class__._shallow_copy = value
```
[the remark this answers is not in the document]
billbrod: because this is "global for all composite basis classes", it affects the behavior of several different methods from the composite classes.

billbrod (changes requested):
Generally looks good! Just a couple of small changes:

- Need to update labels at the end of the mul/pow recursion (and add test).
```python
bas = nmo.basis.BSplineEval(5)
add = bas*3
add.basis1.basis1.label == add.basis2.label
```
- pow should have the same check about label as mul

billbrod:
While reviewing this PR, I looked into how to use not-quite-global variables, like the new shallow_copy flag. That flag affects the behavior of some functions of **all** composite classes, so it can't be instance level (i.e., `self._shallow_copy`). Right now, it's class level (`self.__class__._shallow_copy`), which I think is right. While looking into this, came across [this thread](https://discuss.python.org/t/context-manager-that-stores-a-temporary-context-for-globals/43365/4), which is discussing how stdlib's `warnings` handles filters, and they're set at the module-level, using `sys.modules["warnings"].filters`. So if we had flags that affected the behavior across the module (not just within a single class/mixin), then that would be the way to go.

BalzaniEdoardo:
I think I should have addressed everything, including adding missing tests! @billbrod

src/nemos/basis/_basis_mixin.py
```diff
@@ -852,6 +860,16 @@ def _iterate_over_components(self):
             components2,
         )
 
+    @contextmanager
+    def _set_shallow_copy(self, value):
+        """Context manger for setting the shallow copy flag in a thread safe way."""
```
gviejo, line 865:
```suggestion
        """Context manager for setting the shallow copy flag in a thread safe way."""
```

gviejo (approved):

PR #308, opened 2025-02-13
Dynamic version fix

billbrod:
While putting together the equivalent PR for pynapple, I noticed the [setuptools_scm docs](https://setuptools-scm.readthedocs.io/en/latest/usage/#as-cli-tool) (some issues rendering the docs here, search for "at runtime" on that page) suggest *not* writing version to a file anymore, but using `importlib.metadata` instead, which comes from [PEP 566](https://peps.python.org/pep-0566/).

That's what this fix does, as well as following [their instructions](https://setuptools-scm.readthedocs.io/en/latest/usage/#usage-from-sphinx) for sphinx

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -153,5 +153,3 @@ docs/data/
 # rst generated files
 docs/stubs
 
-# created automatically by setuptools.scm, don't track
-src/nemos/version.py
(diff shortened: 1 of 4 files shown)


BalzaniEdoardo (approved):
Good catch

PR #310, opened 2025-02-14
Update raised cosine width docs and error message

billbrod:
Realized that our error message for the Raised Cosine basis objects width argument was out-of-date. Tried to update docs and error message to be clearer.

@BalzaniEdoardo not sure about copying the `Raises ValueError` section to the init docstring: I don't remember if it's appropriate to put that section there or not.

### DIFF
--- a/src/nemos/basis/_raised_cosine_basis.py
+++ b/src/nemos/basis/_raised_cosine_basis.py
@@ -28,7 +28,8 @@ class RaisedCosineBasisLinear(Basis, AtomicBasisMixin, abc.ABC):
         The mode of operation. 'eval' for evaluation at sample points,
         'conv' for convolutional operation.
     width :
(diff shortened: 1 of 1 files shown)


billbrod:
Need to update this error to explain that the issue is about aliasing.

BalzaniEdoardo (approved):
Just fix the tests, if it passes, I'll merge. Thanks Billy

billbrod:
> Just fix the tests, if it passes, I'll merge. Thanks Billy

I was waiting to hear from @flyingfordanglia whether this error message is clear enough.

Also, the failing test looks like a temporary issue with htmlproofer (related to the nemos twitter account)

BalzaniEdoardo:
Ok, I'll merge this for now, if we get more feedback we can improve the error message even further.

PR #324, opened 2025-03-10
Updates current injection tutorial

billbrod:
This PR updates the current injection notebook to reflect the changes in our [January/February workshops](https://flatironinstitute.github.io/neurorse-workshops/workshops/jan-2025/branch/main/).

A lot of it is minor text changes, but this also adds a final section about using basis objects for extending the model. Does that belong here? Or should we move it to a sequel notebook?

EDIT: I also edited conf.py to set `nb_execution_mode = 'cache'` ([docs here](https://myst-nb.readthedocs.io/en/latest/computation/execute.html#notebook-execution-modes)), which should speed up local deevelopment.


src/nemos/_documentation_utils/plotting.py
```diff
@@ -238,9 +238,17 @@ def current_injection_plot(
...
+            lbls = [" (current history)", " (instantaneous only)"]
+        else:
+            lbls = [""]
+        for pred_fr, style, lbl in zip(predicted_firing_rate, ["-", "--"], lbls):
+            resp_ax.plot(
+                pred_fr,
+                linestyle=style,
+                color="tomato",
+                label=f"Predicted firing rate{lbl}",
+            )
```
BalzaniEdoardo, line 251: this zip doesn't really work. I assume `predicted_firing_rate` when one hits this is a (n_samples, 2) TsdFrame. If you zip that, you'll just iterate over the first two samples and you'll get an array `pred_fr` with two elements, not a pynapple Tsd. Transposition deosn't work either, because it kicks you out of pynapple.

src/nemos/_documentation_utils/plotting.py
```diff
@@ -273,7 +281,8 @@ def current_injection_plot(
         ax.plot(firing_rate.restrict(interval), color="k")
         ax.plot(spikes.restrict(interval).to_tsd([-1.5]), "|", color="k", ms=10)
         if predicted_firing_rate:
-            ax.plot(predicted_firing_rate.restrict(interval), color="tomato")
+            for pred_fr, style in zip(predicted_firing_rate, ["-", "--"]):
+                ax.plot(pred_fr.restrict(interval), linestyle=style, color="tomato")
```
BalzaniEdoardo, line 285: Same thing, this will for sure be an error because you are restricting a numpy array

BalzaniEdoardo (changes requested):
The docs are not building for an error in the plotting funcs. I fixed a couple of typos there already but there is still some leftovers that I don't want to misinterpret. Once the docs build I'll give you feedback on that too.

billbrod:
I think I fixed the issue, let me know what you think.

I also edited conf.py to set `nb_execution_mode = 'cache'` ([docs here](https://myst-nb.readthedocs.io/en/latest/computation/execute.html#notebook-execution-modes)), which should speed up local deevelopment. can remove if you'd prefer

BalzaniEdoardo (approved):
All good, thanks Billy

PR #330, opened 2025-03-25
Fix convergence tests not testing (Prox)SVRG

bagibence:
The tests were originally implemented for `GradientDescent` and `ProximalGradient`, then later parametrized to include `SVRG` and `ProxSVRG`, but the function bodies didn't actually use the parameters.
This fixes that, actually running the tests for `(Prox)SVRG`.


BalzaniEdoardo:
hei Bence, thanks for this, sarah has already a fox tor this in the pr on a
new observation model. she also refactors teste, check that out.

sorry about that, we should coordinate better

Issue #319, opened 2025-03-06
Set observation model at initialization by its string name


sjvenditto:
This will allow for easier initialization than calling the full path to the object

BalzaniEdoardo:
This is addressed in PR #335

PR #335, opened 2025-03-30
String obs model

BalzaniEdoardo:
This PR adds the possibility of defining an observation model from a string, in the same way one can initialize the regularizer with a string. The implementation mimics the regularizer builder.

Addresses #319

### DIFF
--- a/src/nemos/glm.py
+++ b/src/nemos/glm.py
@@ -160,14 +160,14 @@ class GLM(BaseRegressor):
     >>> # define a Gamma GLM providing a string
     >>> nmo.glm.GLM(observation_model="Gamma")
     GLM(
-        observation_model=GammaObservations(inverse_link_function=1/x),
+        observation_model=GammaObservations(inverse_link_function=one_over_x),
         regularizer=UnRegularized(),
         solver_name='GradientDescent'
     )
     >>> # or equivalently, passing the observation model object
     >>> nmo.glm.GLM(observation_model=nmo.observation_models.GammaObservations())
     GLM(
-        observation_model=GammaObservations(inverse_link_function=1/x),
+        observation_model=GammaObservations(inverse_link_function=one_over_x),
         regularizer=UnRegularized(),
         solver_name='GradientDescent'
     )
--- a/src/nemos/glm.py
+++ b/src/nemos/glm.py
@@ -14,6 +14,7 @@ from numpy.typing import ArrayLike
 
 from . import observation_models as obs
 from . import tree_utils, validation
+from ._observation_model_builder import instantiate_observation_model
 from .base_regressor import BaseRegressor
 from .exceptions import NotFittedError
 from .initialize_regressor import initialize_intercept_matching_mean_rate
@@ -167,7 +168,9 @@ class GLM(BaseRegressor):
 
     def __init__(
         self,
-        observation_model: obs.Observations = obs.PoissonObservations(),
+        # With python 3.11 Literal[*AVAILABLE_OBSERVATION_MODELS] will be allowed.
+        # Replace this manual list after dropping support for 3.10?
+        observation_model: obs.Observations | Literal["Poisson", "Gamma"] = "Poisson",
         regularizer: Union[str, Regularizer] = "UnRegularized",
         regularizer_strength: Optional[float] = None,
         solver_name: str = None,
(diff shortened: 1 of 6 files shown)


src/nemos/glm.py
```diff
@@ -167,7 +168,9 @@ class GLM(BaseRegressor):
 
     def __init__(
         self,
-        observation_model: obs.Observations = obs.PoissonObservations(),
+        # With python 3.11 Literal[*AVAILABLE_OBSERVATION_MODELS] will be allowed.
```
billbrod, line 171: I was about to ask about this. Should make this an issue so we don't forget.

billbrod (changes requested):
This looks good to me, just a couple small quality-of-life things:
- We're not intending for users to interact with `instantiate_observation_model`, right?
- If you pass a non-string / non-ObservationModel to GLM, you get a bit of a cryptic error message: `AttributeError: The provided object does not have the required inverse_link_function attribute`. This makes sense if the user was trying to pass their own custom observation model, but not if they accidentally passed an int or a list or something. Should probably make this a bit more informative.
- Unrelated to this PR, but I just noticed: the inverse_link_function of the Gamma is a lambda function, which means it's repr is uniformative. Can we def that function in the file so it has a proper name?
- Need to add at least one user-facing example for this, either in docs or in a doctest.

BalzaniEdoardo:
> This looks good to me, just a couple small quality-of-life things:
> 
>     * We're not intending for users to interact with `instantiate_observation_model`, right?
Yes, it is internal

>     * If you pass a non-string / non-ObservationModel to GLM, you get a bit of a cryptic error message: `AttributeError: The provided object does not have the required inverse_link_function attribute`. This makes sense if the user was trying to pass their own custom observation model, but not if they accidentally passed an int or a list or something. Should probably make this a bit more informative.
> 
I'll fix this after #329 since test will be refactored.

>     * Unrelated to this PR, but I just noticed: the inverse_link_function of the Gamma is a lambda function, which means it's repr is uniformative. Can we def that function in the file so it has a proper name?
Sarah set up a dict that will return a better repr for default links in her Bernoulli observation PR. I'll add repr for this link there. #329 

>     * Need to add at least one user-facing example for this, either in docs or in a doctest.
I added an example to GLM docstring

src/nemos/observation_models.py
```diff
@@ -890,6 +890,10 @@ def check_observation_model(observation_model):
     ...                     jnp.sum((y_true - y_true.mean()) ** 2))
     ...     def sample_generator(self, key, params, scale=1.):
     ...         return jax.random.bernoulli(key, params)
+    ...     def estimate_scale(self, y, predicted_rate, dof_resid):
+    ...         return 1
+    ...     def log_likelihood(self, params, y_true, aggregate_sample_scores=jnp.mean):
+    ...         return -self._negative_log_likelihood(params, y_true, aggregate_sample_scores)
```
billbrod, line 896: why these additions?

tests/test_observation_models.py
```diff
@@ -22,6 +22,55 @@ def gamma_observations():
...
+        ("Poisson", does_not_raise()),
+        ("Gamma", does_not_raise()),
+        (
+            "invalid",
+            pytest.raises(ValueError, match="Unknown observation model: invalid"),
+        ),
+    ],
+)
+@pytest.mark.parametrize("glm_class", [nmo.glm.GLM, nmo.glm.PopulationGLM])
+def test_glm_setter_observation_model(obs_model_string, glm_class, expectation):
```
billbrod, line 56: This test is checking that you can change observation model after the fact and the behavior is the same as if you pass the string at init?

billbrod (approved):
Just a couple comprehension questions and then this looks good to me

src/nemos/observation_models.py
BalzaniEdoardo: I started addressing the informative message comment before realizing that the test will be refactored in Sarah's pr and realized that to be a valid observation model we need these two methods also

tests/test_observation_models.py
BalzaniEdoardo: yes and yes, I'll add it at the level of the test.

BalzaniEdoardo:
**Edit:** I actually fixed the representation of the default inverse link function for the Gamma.
The solution I added is to create a class `NamedFunction` that wraps a given function and receives a `name`, which is a string that we can use in the repr.

billbrod (approved):

Issue #298, opened 2025-01-23
GLM metadata support?


billbrod:
Should we have better metadata support for the GLM objects? in particular, with predict, it feels weird to pass the glm object a pynapple object with columns and then... drop the columns. would be nice to grab columns if available (which would work with pynapple, pandas, etc.), and similarly for rate.

(columns here are neuron labels)

could also make them optional arguments to the function? but that starts to feel cluttered

BalzaniEdoardo:
What are you suggesting for this one? 

For a PopulationGLM the counts might have the information about the columns that we can store when fit is called.  In a regular GLM, the counts are usually a `Tsd`, does that have metadata associated with the neuron type @sjvenditto  and @gviejo ?

sjvenditto:
Yeah `Tsd` objects currently don't have metadata, but a `TsdFrame` of counts for a `PopulationGLM` could have metadata associated with each column

however, right now, when `count` is called on a `TsGroup`, metadata such as neuron type isn't conserved in the returned `TsdFrame`. although this may be a good feature to add

BalzaniEdoardo:
This is solved in #339, only for TsdFrame. I'll leave it open until the metadata is linked to `Tsd` as well

PR #339, opened 2025-04-07
Attach Metadata on `predict`

BalzaniEdoardo:
This tackles #298. Adds a decorator to`GLM.fit` which strips the metdata.

This PR must follow @sjvenditto #329 which refactors GLM tests.

## Additional machinery

- Private attribute `_metadata` for PopulationGLM.
- `strip_metadata` decorator factory: strip metdatata & columns from  a specified parameter if available as attributes.
- `cast_out` function in type casting receives an additional metadata argument.
- On `support_pynapple`, try to re-attach metadata if `self` has attribute `_metadata`.
- Generalized check for pynapple arrays that works with  `FeaturePytree`s.
- Generalized time info stripping that works with `FeaturePytree`s.
- Re-attach metadata in `cast_to_pynapple` when number of column of output matches metadata columns.


src/nemos/base_regressor.py
```diff
@@ -22,6 +23,28 @@
...
+        raise ValueError("Must specify either arg_num or kwarg_key.")
+
+    def decorator(func):
+        """Strip metadata if available."""
+
+        @wraps(func)
+        def wrapper(self, *args, **kwargs):
+            inp = args[arg_num] if arg_num is not None else kwargs[kwarg_key]
+            self._metadata = {
+                "metadata": inp.metadata if hasattr(inp, "metadata") else None,
```
sjvenditto, line 38:
```suggestion
                "metadata": inp._metadata if hasattr(inp, "_metadata") else None,
```

`inp.metadata` will always pull a pandas DataFrame -- this will be more relevant in the future when the underlying metadata is a dictionary. I would pull from `_metadata` instead which will ensure that the dictionary is pulled in the future

tests/test_glm.py
```diff
@@ -2626,6 +2629,49 @@ def test_feature_mask_compatibility_fit(
...
+        ],
+    )
+    def test_metadata_pynapple_fit(self, reg_setup, request):
+        X, y, model, true_params, firing_rate = request.getfixturevalue(reg_setup)
+        y = TsdFrame(
+            t=np.arange(y.shape[0]), d=y, metadata={"y": np.arange(y.shape[1])}
+        )
+        model.fit(X, y)
+        assert hasattr(model, "_metadata") and (model._metadata is not None)
+        assert np.all(y.metadata == model._metadata["metadata"])
```
sjvenditto, line 2646:
```suggestion
        assert np.all(y._metadata == model._metadata["metadata"])
```
if you make the change above, then this will need to get changed too

tests/test_glm.py
```diff
@@ -2626,6 +2629,49 @@ def test_feature_mask_compatibility_fit(
...
+        )
+
+        def convert_to_nap(arr):
+            return TsdFrame(t=y.t, d=getattr(arr, "d", arr))
+
+        X = jax.tree_util.tree_map(convert_to_nap, X)
+        model.fit(X, y)
+        rate = model.predict(X)
+        assert hasattr(rate, "metadata") and (rate.metadata is not None)
+        assert np.all(y.metadata == rate.metadata)
```
sjvenditto, line 2672:
```suggestion
        assert np.all(y._metadata == rate.metadata)
```

sjvenditto (changes requested):
Looking good! My main feedback is around how you pull the metadata from pynapple using the `metadata` attribute. This shouldn't change anything with the current metadata implementation, but in the future move towards dictionary-based metadata, the `metadata` attribute will still return a pandas DataFrame. So, in the future, if you want the metadata dictionary stored instead of a DataFrame, you'll need to pull it from the `_metadata` attribute. If you want it always stored as a DataFrame, it's fine as-is!

PR #342, opened 2025-04-28
Drop NaNs in `GLM.initialize_state`

bagibence:
Fitting a GLM to data containing NaNs with the `jaxopt.LBFGS` solver worked when using `GLM.fit`, but not when writing the training loop in Python and calling `GLM.update` repeatedly.
The problem was that this solver uses the data to initialize its state, and the state was populated with NaNs by `GLM.initialize_state`.
Dropping NaNs from the input data in `GLM.initialize_state` -- as is done in `GLM.update` and `GLM.run` -- solved this problem.

### DIFF
--- a/src/nemos/glm.py
+++ b/src/nemos/glm.py
@@ -982,6 +982,7 @@ class GLM(BaseRegressor):
 
         return init_params
 
+    @cast_to_jax
     def initialize_state(
         self,
         X: DESIGN_INPUT_TYPE,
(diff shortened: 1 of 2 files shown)


BalzaniEdoardo:
Great, thanks Bence, can you add a test checking the case that was failing for you? i.e., an initialization with nans, then this will be good to go.  There is a `test_glm_initialization.py` that you can edit

BalzaniEdoardo (changes requested):
If you add a test for the case that was failing (a vector with nans) this will be good to go

bagibence:
I ended up adding the test in `test_glm.py` and in that next to the update tests, because while the change is done in `GLM.initialize_state`, testing the state itself would have to be solver-specific, and the problem ultimately comes up in `GLM.update` giving NaN parameters after using the wrongly initialized state.

BalzaniEdoardo (approved):
Great, thanks Bence

PR #347, opened 2025-05-20
Move JAXopt code into NeMoS

### LINKED ISSUE #346: Move JAXopt code into NeMoS
To guard against new JAX versions breaking JAXopt and no way of fixing it, copy the needed algorithms into NeMoS so that they can be quickly fixed if needed.

Potentially only move the algorithms that NeMoS relies on instead of the whole jaxopt package.

bagibence:
- Added JAXopt code in `src/nemos/third_party` keeping the license
- Changed imports in NeMoS and JAXopt to reflect this location
- Added a log of changes to JAXopt's readme
- Adjusted the dependencies removing JAXopt from NeMoS's dependencies but adding JAXopt's dev dependencies
  - [x] Decide how of JAXopt's dev depencies are handled
- Adjusted `tox.ini` to run JAXopt's tests

I explored other locations to put the code (top-level, in `src`, in `src/nemos`, `src/nemos/_solvers`), and this turned out to be the best, e.g. allowing imports as `nemos.third_party.jaxopt`.

Some of JAXopt's tests in `isotonic_test.py` fail for me locally.

As NeMoS only requires a subset of the algorithms implemented in JAXopt, removing the rest would make sense before merging.

Fixes #346


BalzaniEdoardo (commented):
This looks good to me, but we should edit the readme of the package adding a statement acknowledging that we ported the jaxopt code while we are transitioning to optax/optimistix for our solvers. 

I would also link the readme to the jaxopt apache license.

BalzaniEdoardo (commented):
TODOs

- [x] Add a job to the CI that runs the jaxopt tests via tox
- [x] Move the test dependencies of jaxopt in a separate "jaxopt-tests" optional deps.
- [ ] Acknowledge the port of `jaxopt` in the readme and add a sentence stating that this is a temporary port while we complete the transition to `optax` and `optimistix`.

billbrod:
I agree with all Edoardo's points. I would also update CONTRIBUTING to mention this, linking to the readme description.

This practice of repackaging third-party libraries is known as "vendoring", and if you want to see how another library does this, [pip vendors many libraries](https://github.com/pypa/pip/tree/main/src/pip/_vendor). I'd check out that readme and [init](https://github.com/pypa/pip/blob/main/src/pip/_vendor/__init__.py), poke around and see how they handle attribution etc.

Fortunately for us, pip (like nemos) is [MIT-licensed](https://github.com/pypa/pip?tab=MIT-1-ov-file#readme) and vendors several Apache 2-licensed libraries (like jaxopt), e.g., [requests](https://github.com/pypa/pip/blob/main/src/pip/_vendor/requests/LICENSE). So that's an existence proof for license compatibility.

bagibence:
Added `tox_jaxopt` to the CI.

Added a readme in `src/nemos/third_party` and explained why JAXopt is vendored, and linked to it in CONTRIBUTING. Should it be mentioned in the main readme as well? It seemed a bit out of place there for me.

BalzaniEdoardo:
I removed the numba jitting in `src/nemos/third_party/jaxopt/_src/isotonic.py` which wasn't working. 

The CI of jaxopt wasn't requiring `numba`, so the bug wasn't detected, but nemos does install it as a pynapple dependecy. We do not use `isotonic.py` at all in our code so that doesn't make a difference.

pyproject.toml
```diff
@@ -23,13 +23,12 @@ classifiers = [
...
 dependencies = [
     "jax>=0.4.28",                     # Numerical computing library
-    "jaxopt>=0.6",                  # Optimization library built on JAX
     "numpy>1.20",                   # Numerical computing library
     "scipy>=1.10",                  # Scientific computing library
     "typing_extensions>=4.6",       # Typing extensions for Python
     "pynapple>=0.6.6",
 ]
-
+# jaxopt 0.8.5 is included in src/nemos/third_party/jaxopt
```
billbrod: this should probably live next to the `jaxopt_dev` optional dependency? otherwise we're likely to forget about this line

src/nemos/third_party/README.md
```diff
@@ -0,0 +1,13 @@
...
+\
+As JAXopt is no longer maintained, we are in the process of transitioning to [Optimistix](https://github.com/patrick-kidger/optimistix) and [Optax](https://github.com/google-deepmind/optax).
+\
+In order to be able to issue quick fixes in case new JAX versions break JAXopt's functionality,
+we temporarily vendor JAXopt's last release (0.8.5) until the transition is successful.
+
+## Changes made to `JAXopt`
+- Removed docs, examples, benchmarks
+- Changed import paths to import `jaxopt` from `nemos.third_party`
+- Removed Github workflows (`.github`)
```
billbrod, line 25: add the bit about numba @BalzaniEdoardo ?

billbrod: I would also add a line like "Our approach is inspired by how [pip](https://github.com/pypa/pip/tree/main/src/pip/_vendor) vendors third-party libraries", just so we preserve that bit of info.

CONTRIBUTING.md
```diff
@@ -282,3 +282,6 @@ properly documented as outlined below.
 
 > [!NOTE]
 > All internal links to NeMoS documentation pages **must be relative**. Using absolute links can lead to broken references whenever the documentation structure is altered. Any presence of absolute links to documentation pages will cause the continuous integration checks to fail. Please ensure all links follow the relative format before submitting your PR.
+
+### Modifying vendored libraries
+All changes made to vendored libraries should be documented in the [changelog](./src/nemos/third_party/README.md).
```
billbrod, line 287: I'd have a general purpose statement here along the lines of, "We may occasionally temporarily vendor third-party libraries. That is, we include license-compatible libraries within the nemos source code when needed, e.g., when a library is being deprecated. In such cases, this is only temporary: ultimately, we will remove the dependency, but vendoring the code gives us a bit more stability while doing so. And note that this is only done in extreme circumstances." 

Just something to explain why and when we would do such a thing.

billbrod (approved):
This looks like a reasonable way to handle it to me. I just noted a couple of small extra pieces of info that I think would be useful. It's probably also useful to write in an issue / the project where the pieces that will be deleted live, so that we don't forget to delete something.

CONTRIBUTING.md
bagibence: Makes sense. I added a version of this in `third_party/README.md`. Let me know how you like it.

src/nemos/third_party/README.md
bagibence: Added a line about removing numba.
Added the "Our approach ..." sentence after the longer description of the why and when of vendoring.

pyproject.toml
bagibence: Moved it!

Issue #345, opened 2025-05-18
GPU memory allocation when using compute_features


arturoptophys:
**Describe the bug**
When running the code example below. I get 
`jaxlib.xla_extension.XlaRuntimeError: RESOURCE_EXHAUSTED: Out of memory while trying to allocate 3199993600 bytes.`

`2025-05-18 21:59:55.086788: W external/xla/xla/tsl/framework/bfc_allocator.cc:310] Allocator (GPU_0_bfc) ran out of memory trying to allocate 3.37GiB with freed_by_count=0. The caller indicates that this is not a failure, but this may mean that there could be performance gains if more memory were available.
2025-05-18 21:59:55.405578: W external/xla/xla/tsl/framework/bfc_allocator.cc:310] Allocator (GPU_0_bfc) ran out of memory trying to allocate 3.37GiB with freed_by_count=0. The caller indicates that this is not a failure, but this may mean that there could be performance gains if more memory were available.
2025-05-18 21:59:55.963174: W external/xla/xla/tsl/framework/bfc_allocator.cc:310] Allocator (GPU_0_bfc) ran out of memory trying to allocate 4.49GiB with freed_by_count=0. The caller indicates that this is not a failure, but this may mean that there could be performance gains if more memory were available.
2025-05-18 21:59:57.340116: W external/xla/xla/tsl/framework/bfc_allocator.cc:310] Allocator (GPU_0_bfc) ran out of memory trying to allocate 4.49GiB with freed_by_count=0. The caller indicates that this is not a failure, but this may mean that there could be performance gains if more memory were available.
2025-05-18 21:59:57.340144: W external/xla/xla/tsl/framework/bfc_allocator.cc:310] Allocator (GPU_0_bfc) ran out of memory trying to allocate 4.49GiB with freed_by_count=0. The caller indicates that this is not a failure, but this may mean that there could be performance gains if more memory were available.
2025-05-18 21:59:57.915230: W external/xla/xla/tsl/framework/bfc_allocator.cc:310] Allocator (GPU_0_bfc) ran out of memory trying to allocate 4.49GiB with freed_by_count=0. The caller indicates that this is not a failure, but this may mean that there could be performance gains if more memory were available.
[...]

PR #348, opened 2025-05-21
Batch convolution

BalzaniEdoardo:
This PR addresses issues in #345. In particular, it introduces the possibility of batching the convolution over input channels and filters (basis functions usually).

## Background

NeMoS efficiently applies convolution in a vectorized way. By default, tries to vectorize over all dimensions:

- The input channels (for example, the number of neurons in a spike count array).
- The number of filters (number of basis function usually).

This behavior is equivalent to the original implementation. This vectorization process allocates a large amount of memory (num channels x num filters). For example, convolving the counts of 200 neurons and 8 basis functions, would require vectorizing over (200 x 8) dimensions, causing the GPU to run out of memory even if the input and output of the convolution easily fit in memory.

## What's New

In this PR, I introduced three optional parameters: `batch_size_samples`, `batch_size_channels` and `batch_size_basis`. This are integer values representing the batch size over which applying the vectorized convolution. In particular, when specified, the convolution will vectorize over (`batch_size_channels` x `batch_size_basis`) dimensions and apply the convolution in sequential fashion over chunks of the array of length `batch_size_samples`. The most memory conservative approach would be to set those parameters to 1, defaulting to applying one convolution at the time.

## Example
Here is a script demonstrating the feature. 

```python
import os

# JAX memory management flags
os.environ["XLA_PYTHON_CLIENT_PREALLOCATE"] = "false"
# Optional: Uncomment to make JAX deallocate GPU memory aggressively
os.environ["XLA_PYTHON_CLIENT_ALLOCATOR"] = "platform"

import time
import subprocess
import nemos as nmo
import jax
import numpy as np

def print_gpu_mem(tag=""):
    output = subprocess.check_output([
        'nvidia-smi', '--query-gpu=memory.used,memory.total',
        '--format=csv,nounits,noheader'
    ])
    used, total = map(int, output.decode().split(','))
    print(f"[{tag}] GPU memory usage: {used} MiB / {total} MiB")

print("JAX version:", jax.__version__)
print("JAX device:", jax.devices()[0])
print("JAX device type:", jax.devices()[0].device_kind)
print("sys.platform:", os.sys.platform)
print("Nemos version:", nmo.__version__)
print("NumPy version:", np.__version__)

window_size = 20

basis = nmo.basis.RaisedCosineLogConv(
    n_basis_funcs=8, window_size=window_size, label="count_history",
    conv_kwargs={"batch_size_samples": 125_000,"batch_size_channels":1, "batch_size_basis": 1}
)

spike_matrix = np.random.randint(0, 2, size=(500_000, 200))

print_gpu_mem("Before compute_features")
start = time.time()
conv_spk = basis.compute_features(spike_matrix)
elapsed = time.time() - start
print_gpu_mem("After compute_features")

print(f"conv_spk shape: {conv_spk.shape}")
print(f"conv_spk nbytes: {conv_spk.nbytes / 1024**2:.2f} MiB")
print(f"Computation took {elapsed:.2f} seconds")


```

Which on my machine outputs,

```
(venv) [ebalzani@ccnlin052 generalized-linear-models]$ python _scripts/gpu_mem_allocation.py 
JAX version: 0.5.0
JAX device: cuda:0
JAX device type: Quadro RTX 6000
sys.platform: linux
Nemos version: 0.2.3.dev190
NumPy version: 2.1.3
[Before compute_features] GPU memory usage: 1463 MiB / 24576 MiB
[After compute_features] GPU memory usage: 4527 MiB / 24576 MiB
conv_spk shape: (500000, 1600)
conv_spk nbytes: 3051.76 MiB
Computation took 2.60 seconds
```


BalzaniEdoardo:
@arturoptophys you can check this branch out and let me know how it goes.

arturoptophys:
Yes, works now !
However, 
`os.environ["XLA_PYTHON_CLIENT_ALLOCATOR"] = "platform"`
is still required

with setting it:
```
JAX version: 0.6.0
JAX device: cuda:0
JAX device type: NVIDIA RTX 4000 Ada Generation Laptop GPU
sys.platform: linux
Nemos version: 0.2.3
NumPy version: 1.26.4

[Before compute_features] GPU memory usage: 2285 MiB / 12282 MiB
2025-05-21 22:03:01.496869: W external/xla/xla/hlo/transforms/simplifiers/hlo_rematerialization.cc:3021] Can't reduce memory use below 5.63GiB (6044409844 bytes) by rematerialization; only reduced to 5.98GiB (6419757132 bytes), down from 5.98GiB (6419757132 bytes) originally
[After compute_features] GPU memory usage: 5349 MiB / 12282 MiB
conv_spk shape: (500000, 1600)
conv_spk nbytes: 3051.76 MiB
Computation took 1.79 seconds
```

without:
2025-05-21 22:05:41.966492: W external/xla/xla/hlo/transforms/simplifiers/hlo_rematerialization.cc:3021] Can't reduce memory use below 5.08GiB (5452891411 bytes) by rematerialization; only reduced to 5.98GiB (6419757132 bytes), down from 5.98GiB (6419757132 bytes) originally
2025-05-21 22:05:42.161530: W external/xla/xla/hlo/transforms/simplifiers/hlo_rematerialization.cc:3021] Can't reduce memory use below 2.60GiB (2792892015 bytes) by rematerialization; only reduced to 2.98GiB (3200000000 bytes), down from 2.98GiB (3200000000 bytes) originally
2025-05-21 22:05:42.215929: W external/xla/xla/hlo/transforms/simplifiers/hlo_rematerialization.cc:3021] Can't reduce memory use below 2.60GiB (2790882583 bytes) by rematerialization; only reduced to 2.98GiB (3200493599 bytes), down from 2.98GiB (3200493599 bytes) originally
2025-05-21 22:05:52.238461: W external/xla/xla/tsl/framework/bfc_allocator.cc:501] Allocator (GPU_0_bfc) ran out of memory trying to allocate 2.98GiB (rounded to 3199993600)requested by op 
2025-05-21 22:05:52.238531: W external/xla/xla/tsl/framework/bfc_allocator.cc:512] *****_________________________________***********************************___________________________
E0521 22:05:52.238543   23357 pjrt_stream_executor_client.cc:2839] Execution of replica 0 failed: RESOURCE_EXHAUSTED: Out of memory while trying to allocate 3199993600 bytes. [tf-allocator-allocation-error='']
Traceback (most recent call last):

BalzaniEdoardo:
Ok, I will make it bullet proof by batching over time as well (so that in principle it would be possible to parallelise on the other axes and have an efficient convolution on the gpu). Good that it works already but i'll keep working on ut

BalzaniEdoardo:
@arturoptophys try it now, I added the possibility of scan through time as well. I updated my comment to reflect the new parameter names. Scanning over time will be slightly slower (or quite slower for small batch sizes) but safer. Let me know if this work without the flags for deallocating etc.

docs/background/plot_03_1D_convolution.md
```diff
@@ -86,7 +86,8 @@ _, w = basis_obj.evaluate_on_grid(ws)
 
 plt.plot(w)
 
-spk_conv = nmo.convolve.tensor_convolve(spk, w)
+# convolve specifying the batch size for time, channels and basis
+spk_conv = nmo.convolve.tensor_convolve(spk, w, n_samples, 1, w.shape[1])
```
sjvenditto: Should we change this example to use numpy or scipy convolve instead? If defaults aren't being supplied for batch sizes in `tensor_convolve`, we might not want it to be a user-facing function

sjvenditto (commented):
Everything looks good! The user-side setting of batch sizes is clean, and the documentation is clear on what axes the batch sizes act on. I only have a comment on the use of `tensor_convolve` in the docs, since it doesn't supply default batch sizes.

docs/background/plot_03_1D_convolution.md
billbrod: or should it just have defaults?

docs/how_to_guide/convolve_large_arrays.md
```diff
@@ -0,0 +1,62 @@
...
+
+<figure markdown>
+<!-- note that the src here has an extra ../ compared to other images, necessary when specifying path directly in html -->
+<img src="../_static/convolve_batching_scheme.svg" style="width: 100%", alt="Batched dimensions scheme."/>
+<figcaption>Schematic of the batched dimensions.</figcaption>
+</figure>
+
+## Example
+
+```{code-cell} ipython3
```
billbrod, line 43: might be useful to show memory/time performance with a couple different choices for `batch_size_dict`

docs/how_to_guide/convolve_large_arrays.md
```diff
@@ -0,0 +1,62 @@
...
+
+time_series = np.random.randn(n_samples, n_channels)
+
+# define a basis in conv mode sepecifying the batch sizes
+basis = nmo.basis.RaisedCosineLogConv(n_basis, window_size, conv_kwargs=batch_size_dict)
+
+# performe the convolution as usual
+out = basis.compute_features(time_series)
+
+# note that the channels can be resahaped and nothing would change
```
billbrod: I don't understand the point that's being made here

src/nemos/convolve.py
```diff
@@ -6,24 +6,160 @@
...
+    other_array: NDArray,
+    binary_func: Callable[[NDArray, NDArray], NDArray],
+    batch_size: int,
+    axis: int = 1,
+    out_axis: int = 1,
+    pad_final_batch: bool = True,
+):
     """
-    Apply a convolution on the given array with the evaluation basis and reshapes the result.
+    Apply a binary function in memory-efficient batches along a specified axis.
```
billbrod, line 49: what's a binary function?

src/nemos/convolve.py
```diff
@@ -207,8 +433,12 @@ def create_convolutional_predictor(
     predictor_causality: Literal["causal", "acausal", "anti-causal"] = "causal",
     shift: Optional[bool] = None,
     axis: int = 0,
+    batch_size_samples: Optional[int] = None,
+    batch_size_channels: Optional[int] = None,
+    batch_size_basis: Optional[int] = None,
 ):
-    """Create a convolutional predictor by convolving a basis matrix with a time series.
```
billbrod, line 211: A lot has changed here -- is the behavior the same? It looks like some exceptions are removed -- or would they get raised in a called function?

billbrod (changes requested):
In general, this looks good, couple of small comments.
- It doesn't look like `tensor_convolve` or `batch_binary_func` are in the tests, am I just missing them? I feel like it would be worth testing that the time/memory usage changes with the different batch arguments, but that the output remains the same.

docs/background/plot_03_1D_convolution.md
BalzaniEdoardo: Maybe I should use another function here and make `tensor_convolve` private. The reason I am not providing defaults is that it's a `jit` compiled function and adding an if statement increases the compilation time

BalzaniEdoardo: made it private and simplify the docs by removing session on "valid" conv

docs/how_to_guide/convolve_large_arrays.md
BalzaniEdoardo: I did want to show that it works on ND arrays not only 2D

BalzaniEdoardo: I'll rephrase

src/nemos/convolve.py
BalzaniEdoardo: it's a function that receives two arguments and outputs one, $f \\; : A \times B \longrightarrow C$
```python
def add(a, b):
    return a+b
```

src/nemos/convolve.py
BalzaniEdoardo: they are raised in the called functions. the input/output behavior is unchanged. The new parameter changes how the compute is performed, not the input/output

BalzaniEdoardo: There are some extra raises if the batch parameters are not positive integers or None

docs/how_to_guide/convolve_large_arrays.md
BalzaniEdoardo: I tried but:
- The memory allocation during vectorization is only a problem on the GPU. 
- The CPU memory allocation is the same with or without the batching.
- On the docs built on rtd iI can only use the CPU so I won't be able to see effects on the memory footprint. 

Instead of showing the benefits of the approach, I added a note describing that the batching doesn't impact the memory footprint on the CPU.

I also linked to the relevant issue for more details in the initial paragraph.

PR #351, opened 2025-06-02
Allowed regularizer to be set to None. Setting the regularizer to Non…

camila-maura:
None is now equivalent to setting it to "UnRegularized"

### DIFF
--- a/src/nemos/base_regressor.py
+++ b/src/nemos/base_regressor.py
@@ -186,10 +186,10 @@ class BaseRegressor(Base, abc.ABC):
         return self._regularizer
 
     @regularizer.setter
-    def regularizer(self, regularizer: Union[str, Regularizer]):
+    def regularizer(self, regularizer: Union[str, None, Regularizer]):
         """Setter for the regularizer attribute."""
         # instantiate regularizer if str
-        if isinstance(regularizer, str):
+        if isinstance(regularizer, str) or isinstance(regularizer, type(None)):
             self._regularizer = create_regularizer(name=regularizer)
         elif isinstance(regularizer, Regularizer):
             self._regularizer = regularizer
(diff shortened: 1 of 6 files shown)


src/nemos/base_regressor.py
```diff
@@ -186,10 +186,10 @@ def regularizer(self) -> Union[None, Regularizer]:
         return self._regularizer
 
     @regularizer.setter
-    def regularizer(self, regularizer: Union[str, Regularizer]):
+    def regularizer(self, regularizer: Union[str, None, Regularizer]):
```
BalzaniEdoardo: In the setter I would prefer only `str` or `Regularizer`.
 In the `__init__` of `BaseRegressor` you can add
 
 ```python
self.regularizer = "UnRegularized" if regularizer is None else regularizer
```

File: src/nemos/base_regressor.py [hunk 2]
```diff
@@ -186,10 +186,10 @@ def regularizer(self) -> Union[None, Regularizer]:
...
         """Setter for the regularizer attribute."""
         # instantiate regularizer if str
-        if isinstance(regularizer, str):
+        if isinstance(regularizer, str) or isinstance(regularizer, type(None)):
```
BalzaniEdoardo: this should not be needed if we won't allow None, otherwise use `or (regularizer is None)`

src/nemos/glm.py
```diff
@@ -185,7 +185,7 @@ def __init__(
         # With python 3.11 Literal[*AVAILABLE_OBSERVATION_MODELS] will be allowed.
         # Replace this manual list after dropping support for 3.10?
         observation_model: obs.Observations | Literal["Poisson", "Gamma"] = "Poisson",
-        regularizer: Union[str, Regularizer] = "UnRegularized",
+        regularizer: Union[str, None, Regularizer] = "UnRegularized",
```
BalzaniEdoardo: This one should be 

```suggestion
        regularizer: Optional[Union[str, Regularizer]] = None,
```

src/nemos/base_regressor.py
BalzaniEdoardo:
```suggestion
    def regularizer(self, regularizer: Union[str, Regularizer]):
```

BalzaniEdoardo (changes requested):
Good job working through the class inheritance and the tests. I left a couple of todos.

BalzaniEdoardo:
Great job and thanks @camila-maura

PR #352, opened 2025-06-02
save_params function implementation

pcrespo2017:
implementation of a public method to save the parameters of a base regressor object


src/nemos/base_regressor.py
```diff
@@ -684,3 +685,56 @@ def _optimize_solver_params(self, X: DESIGN_INPUT_TYPE, y: jnp.ndarray) -> dict:
     def _get_optimal_solver_params_config(self):
         """Return the functions for computing default step and batch size for the solver."""
         pass
+
+    def save_params(self, filename: str = None) -> None:
```
BalzaniEdoardo: have the path be required (instead of assigning defaults). Accept both str, or pathlib.Path

src/nemos/base_regressor.py
```diff
@@ -684,3 +685,56 @@ def _optimize_solver_params(self, X: DESIGN_INPUT_TYPE, y: jnp.ndarray) -> dict:
...
+
+        model_params = self.get_params()
+
+        # filter out non-native types
+        native_types = (int, float, str, list, dict, tuple, set, bool, type(None))
+
+        saved_params = {}
+
+        # save all native types and numpy objects
+        for key, value in model_params.items():
```
BalzaniEdoardo: You can make this loop a recursion
```python
import numpy as np
import warnings
import nemos as nmo


def unpack_params(params_dict):
    native_types = (int, float, str, list, dict, tuple, set, bool, type(None))
    out = dict()
    for key, value in params_dict.items():
        if hasattr(value, "get_params") and hasattr(value, "set_params"):
            cls = value.__class__
            cls_name = f"{cls.__module__}.{cls.__qualname__}"
            out[cls_name] = unpack_params(value.get_params(deep=False))
        else:
            if (
                isinstance(value, native_types)
                or isinstance(value, np.ndarray)
                or np.isscalar(value)
            ):
                out[key] = value
            else:
                out[key] = str(value)
                warnings.warn(
                    UserWarning(
                        f"Cannot save parameter {key} of type {type(value)}. saved as a string instead."
                    )
                )
    return out

model = nmo.glm.GLM()

unpack_params(model.get_params(deep=False))
```

src/nemos/base_regressor.py
```diff
@@ -684,3 +685,56 @@ def _optimize_solver_params(self, X: DESIGN_INPUT_TYPE, y: jnp.ndarray) -> dict:
     def _get_optimal_solver_params_config(self):
         """Return the functions for computing default step and batch size for the solver."""
         pass
+
+    def save_params(self, filename: str = None) -> None:
+        """
+
+        Save parameters of the model to a npz file.
```
billbrod:
```suggestion
        """
        Save parameters of the model to a npz file.
```
To match nemos conventions

src/nemos/base_regressor.py
```diff
@@ -684,3 +685,56 @@ def _optimize_solver_params(self, X: DESIGN_INPUT_TYPE, y: jnp.ndarray) -> dict:
...
+
+        Parameters
+        ----------
+        filename : str, optional
```
billbrod:
```suggestion
        filename
```
when using type annotations in function signature, don't need them in docstring.

src/nemos/base_regressor.py
billbrod:
```suggestion
    def save_params(self, filename: str = None)
```
in nemos, we don't annotate if we return nothing

src/nemos/base_regressor.py
```diff
@@ -684,3 +685,56 @@ def _optimize_solver_params(self, X: DESIGN_INPUT_TYPE, y: jnp.ndarray) -> dict:
...
+                or np.isscalar(value)
+            ):
+                saved_params[key] = value
+            else:
+                saved_params[key] = str(value)
+                warnings.warn(
+                    UserWarning(
+                        f"Cannot save parameter {key} of type {type(value)}. saved as a string instead."
+                    )
+                )
```
billbrod: I don't think we should warn here -- this is expected behavior and should be user-invisible (i.e., they can save and load using our objects without a problem). I could see using a warning if they're e.g., using a custom link function. That would require checking whether the object comes from within nemos, which I believe you can do with `inspect` (Edoardo's the inspect master, so he'll know)

src/nemos/base_regressor.py
```diff
@@ -684,3 +685,56 @@ def _optimize_solver_params(self, X: DESIGN_INPUT_TYPE, y: jnp.ndarray) -> dict:
...
+
+        Parameters
+        ----------
+        filename : str, optional
+            Name of the npz file to save the model parameters to. If not provided,
+            default is "model_params.npz".
+
+        """
+
+        filename = "model_params.npz" if filename is None else f"{filename}.npz"
```
billbrod: what happens if the user passes a filename with a different extension?

billbrod (changes requested):
Thanks Pablo, some additional notes:
- For getting the name of parameters, use something like [what I do in plenoptic](https://github.com/plenoptic-org/plenoptic/blob/eba78eb8034231501d90123cb45f342f148ba6bd/src/plenoptic/synthesize/synthesis.py#L21). Calling `str` gives you the `__repr__` (meant for users to inspect while using the code), whereas we want to save the info that tells us how to initialize the object.
- I think we want to save `coef_` and `intercept_` even if they're `None`, since that's what they're initialized as.
- Note that their names are `coef_` and `intercept_`! With the trailing underscore, which is a [scikit-learn convention](https://scikit-learn.org/dev/developers/develop.html#estimated-attributes) that signifies "this value is fit".
- There's a general question about how to handle coef and intercept: they won't necessarily be the fit attributes for other models. We could iterate through all attributes, finding those that end in underscores, but I think the better way to do it is to have this be an abstract method which accepts an argument like `save_attrs`, and the concrete subclasses (e.g., `GLM`) should call the superclass save, specifying these attributes (here, `coef_` and `intercept_`). We'll need to write a subclass version of `save` anyway, to specify the user-facing docstring.
- My inclination is that we shouldn't be checking attribute types to see if we can save them, that feels pretty brittle. Instead, since we'll be writing this separately as sub- and super-class methods, I think any attributes that can be saved should be passed to `save_attrs`, along with `coef_` and `intercept_`, and the others should be passed in a separate argument, maybe `save_name_attrs`.
  - @BalzaniEdoardo, in plenoptic, I save the input/output behavior of attributes that I can't save, along with their name. for example, I save the output of the model when passed an image. This allows me to validate them on load, which I do because I require the object to be initialized before loading (and you might want to resume synthesis after saving/loading). Is that the way we want to do it for nemos? It might be overkill, in which case we can just save the object name.
  - You can look through `save` in `synthesis.py` and `metamer.py` to see how this works. 

In addition the changes Edoardo and I suggested, I'd write a load method, with some examples, which will both help you make sure your design decisions make sense, and help us to review how everything fits together.

src/nemos/base_regressor.py
pcrespo2017: done

src/nemos/base_regressor.py
pcrespo2017: done

src/nemos/base_regressor.py
pcrespo2017: applied

src/nemos/base_regressor.py
pcrespo2017: Removed. we still need to think how to address the link function saving and loading

src/nemos/base_regressor.py
pcrespo2017: we leave this to numpy savez. if other extension is used, .npz is appended at the end 

https://numpy.org/doc/2.2/reference/generated/numpy.savez.html

src/nemos/base_regressor.py
```diff
@@ -45,6 +56,75 @@ def wrapper(self, *args, **kwargs):
...
+    """
+
+    attributes_as_string = ["inverse_link_function"]
+    out = dict()
+    for key, value in params_dict.items():
+        # if the parameter is an objet with get_params/set_params,
+        # extract its class name and parameters
+        if hasattr(value, "get_params") and hasattr(value, "set_params"):
+            cls = value.__class__
+            cls_name = f"{cls.__module__}.{cls.__qualname__}"
```
billbrod: I think you just want to use `get_name` here, to avoid duplicated code.

I haven't seen`cls.__qualname__` before, why use that instead of `cls.__name__`?

src/nemos/base_regressor.py
```diff
@@ -684,3 +764,132 @@ def _optimize_solver_params(self, X: DESIGN_INPUT_TYPE, y: jnp.ndarray) -> dict:
...
         pass
+
+    @abstractmethod
+    def save_params(self, filename: Union[str, Path]):
+        """
+        Abstract method that subclasses must implement.
+        """
+        pass
+
+    def _save_params_base(
```
billbrod: is there a reason why this isn't just `save_params`? (which can still be abstract)

src/nemos/base_regressor.py
```diff
@@ -684,3 +764,132 @@ def _optimize_solver_params(self, X: DESIGN_INPUT_TYPE, y: jnp.ndarray) -> dict:
...
+        self,
+        filename: Union[str, Path],
+        save_attrs: dict,
+    ):
+        """
+        Save model parameters and specified attributes to a .npz file.
+        This is a private method intended to be used by subclasses to implement.
```
billbrod, line 707: need a newline above this

src/nemos/base_regressor.py
```diff
@@ -684,3 +764,132 @@ def _optimize_solver_params(self, X: DESIGN_INPUT_TYPE, y: jnp.ndarray) -> dict:
...
+            Path to the saved .npz file.
+
+        Examples
+        --------
+        >>> import nemos as nmo
+        >>> # Create a GLM model with some parameters
+        >>> model = nmo.glm.GLM(regularizer='Lasso', regularizer_strength=100)
+        >>> # Inspect initial parameters
+        >>> print(model.get_params())
+        >>> # Load parameters from previously saved file
```
billbrod: example should show saving and loading behavior here.

billbrod: basically, just save and load the same model

src/nemos/base_regressor.py
```diff
@@ -684,3 +764,132 @@ def _optimize_solver_params(self, X: DESIGN_INPUT_TYPE, y: jnp.ndarray) -> dict:
...
+            "jax_version": importlib.metadata.version("jax"),
+            "nemos_version": importlib.metadata.version("nemos"),
+        }
+
+        try:
+            np.savez(filename, params=save_attrs)
+        except Exception as e:
+            raise RuntimeError(f"Saving failed for '{filename}': {e}")
+
+    def load_params(self, filename: Union[str, Path]):
```
billbrod: @BalzaniEdoardo we need to decide -- do we want `load_params` to be an instance method, like here, or a class method, so they don't need to instantiate the model beforehand?

if it's an instance method, I think we want to do something to make sure that the saved and instantiated objects have the same attributes where both are set (regularizer, link function, etc.), so that all we're changing is the coef and intercept

in plenoptic, it's an instance method, because they can be arbitrary and I didn't want to worry about saving/loading models. But for nemos, where most of the non-savable objects / functions are standard and either part of nemos or jax, I think a class method makes sense.

What do you think?

billbrod: @pcrespo2017 was there a particular reason you made load an instance method?

src/nemos/base_regressor.py
```diff
@@ -684,3 +764,132 @@ def _optimize_solver_params(self, X: DESIGN_INPUT_TYPE, y: jnp.ndarray) -> dict:
...
+
+        save_attrs["save_metadata"] = {
+            "jax_version": importlib.metadata.version("jax"),
+            "nemos_version": importlib.metadata.version("nemos"),
+        }
+
+        try:
+            np.savez(filename, params=save_attrs)
+        except Exception as e:
+            raise RuntimeError(f"Saving failed for '{filename}': {e}")
```
billbrod: when would this happen?

src/nemos/base_regressor.py
```diff
@@ -684,3 +764,132 @@ def _optimize_solver_params(self, X: DESIGN_INPUT_TYPE, y: jnp.ndarray) -> dict:
...
+        >>> # Load parameters from previously saved file
+        >>> model.load_params("model_params.npz")
+        >>> # Updated parameters reflect the saved configuration
+        >>> print(model.get_params())
+        """
+        filename = Path(filename)
+        if not filename.exists():
+            raise FileNotFoundError(f"File not found: {filename}")
+
+        data = np.load(filename, allow_pickle=True)
```
billbrod: we want to avoid `allow_pickle=True`, because that's the thing that allows for arbitrary code to be executed.

See my overall comment for more.

src/nemos/glm.py
```diff
@@ -1158,6 +1159,44 @@ def __sklearn_clone__(self) -> GLM:
...
+        ...     observation_model="Gamma",
+        ...     solver_kwargs={"stepsize": 0.1, "acceleration": False},
+        ...     solver_name="BFGS",
+        ... )
+
+        >>> # Fit the model to add coefficients and intercept attributes
+        >>> model = model.fit(X, spike_counts)
+
+        >>> # Save model parameters to a file
+        >>> model.save_params("model_params")
```
billbrod:
```suggestion
        >>> model.save_params("model_params.npz")
```
while not required, I think we want to encourage users to add the file extension

src/nemos/base_regressor.py
billbrod: (then the subclass just calls `super().save_params(filename, save_attrs)`)

src/nemos/base_regressor.py
```diff
@@ -684,3 +764,132 @@ def _optimize_solver_params(self, X: DESIGN_INPUT_TYPE, y: jnp.ndarray) -> dict:
...
+        saved_attrs = data["params"].item()
+
+        # Extract model parameters
+        valid_params = self.get_params()
+        model_params = {k: v for k, v in saved_attrs.items() if k in valid_params}
```
billbrod: see my comment above about instance vs. class method. regardless of what we decide, I don't like this behavior because it will silently ignore any keys in `saved_attrs` that aren't in `self.get_params()`. If that's the case, then the user probably messed up and we should raise an error (e.g., they tried to call `GLM.load` when they saved a `PopulationGLM`)

billbrod: oh actually, I see that you are raising an exception in the try/except at the end of this function. still, this is a bit hard to parse, I'll write a description in my overall comment that I think is a better structure for the function.

src/nemos/base_regressor.py
```diff
@@ -4,24 +4,35 @@
...
 
 from . import solvers, utils, validation
+from ._observation_model_builder import instantiate_observation_model
 from ._regularizer_builder import AVAILABLE_REGULARIZERS, create_regularizer
 from .base_class import Base
 from .regularizer import Regularizer, UnRegularized
 from .typing import DESIGN_INPUT_TYPE, SolverInit, SolverRun, SolverUpdate
 
+# move somewhere else?
+OBSERVER_NAME_MAP = {
```
billbrod: yes, it would probably be cleaner to just make `instantiate_observation_model` also able to accept `"PoissonObservations"`, etc. (and it should just remove the `Observations` part of the string, rather than containing this explicit map, since it's so straightforward)

src/nemos/base_regressor.py
```diff
@@ -684,3 +764,132 @@ def _optimize_solver_params(self, X: DESIGN_INPUT_TYPE, y: jnp.ndarray) -> dict:
...
+        save_attrs
+            Dictionary containing the attributes specific to the subclass model.
+        """
+
+        # extract model parameters
+        model_params = self.get_params(deep=False)
+        save_params = unpack_params(model_params)
+
+        # append the model parameters to the save_attrs
+        save_attrs.update(save_params)
```
billbrod: the names here are a bit confusing -- let's keep `model_params` as `model_params` throughout. 

For now, `save_attrs` only consists of attributes that are changed during the fitting process. I *think* that will always be the case. In which case, I'd prefer something like `fit_attrs` (and updating the docstring description accordingly). That gives us a fairly clear demarcation -- `self.get_params` should return every attribute that does not change by fitting, and we only need to specify those that do.

src/nemos/glm.py
```diff
@@ -1158,6 +1159,44 @@ def __sklearn_clone__(self) -> GLM:
         klass = self.__class__(**params)
         return klass
 
+    def save_params(self, filename: Union[str, Path]):
+        """
+        Save GLM model parameters to a .npz file.
```
billbrod, line 1180: need to document `filename`

src/nemos/base_regressor.py
```diff
@@ -45,6 +56,75 @@ def wrapper(self, *args, **kwargs):
...
+    params_dict
+        Dictionary of parameters, possibly containing objects.
+
+    Returns
+    -------
+    dict
+        Serializable dictionary with class names and nested parameters.
+    """
+
+    attributes_as_string = ["inverse_link_function"]
```
billbrod: This should be an argument to the function -- it's something that may change with different subclasses.

billbrod:
Ah okay, after talking with Edoardo about `np.savez`, the problem is that it doesn't like nested dictionaries. So I think the way is to write a small method that flattens and unflattens the dictionary, converting nested dictionaries into `__`, like in sklearn. Then you end up with a single flat dictionary and instead of saving it like `np.savez(filename, params=save_attrs)`, you'd call `np.savez(filename, **save_attrs)`.

billbrod:
Also, I chatted with Edoardo and he agreed that the load method should be a class method, and you should change `__init__` to handle the strings we save

src/nemos/base_regressor.py
pcrespo2017: For our cases I think they behave similar:

"For top-level functions and classes, the __qualname__ attribute is equal to the __name__ attribute. For nested classes, methods, and nested functions, the __qualname__ attribute contains a dotted path leading to the object from the module top-level" - https://peps.python.org/pep-3155/

I agree that we can just use get_name

src/nemos/base_regressor.py
pcrespo2017: > (then the subclass just calls `super().save_params(filename, save_attrs)`)

Don't they need to have the exact same signature? How should this look like?

src/nemos/base_regressor.py
pcrespo2017: How? I think I copied Nemos style

src/nemos/base_regressor.py
pcrespo2017: I wanted to be consistent with the saving behavior, but I agree a class method might be better! will write it

src/nemos/base_regressor.py
BalzaniEdoardo: Agree

src/nemos/base_regressor.py
```diff
@@ -45,6 +56,75 @@ def wrapper(self, *args, **kwargs):
     return decorator
 
 
+def get_name(x: object) -> str:
+    """
+    Get the name of an object ``x``, for saving/loading purposes.
+
+    Parameters
+    ----------
+    x
```
BalzaniEdoardo:
```suggestion
    x:
```

src/nemos/base_regressor.py
```diff
@@ -45,6 +56,75 @@ def wrapper(self, *args, **kwargs):
...
+        A python object or function.
+
+    Returns
+    -------
+    name
+        The name of that object, with full module path (e.g.,
```
BalzaniEdoardo:
```suggestion
        The name of the object, with full module path (e.g.,
```

src/nemos/base_regressor.py
```diff
@@ -45,6 +56,75 @@ def wrapper(self, *args, **kwargs):
...
+        ``nemos.observation_models.PoissonObservations``).
+    """
+    if x is None:
+        return None
+    try:
```
BalzaniEdoardo: Can modify this to an if statement, `if hasattr(x, "__class__")`, `elif hasattr(x, "__module")`, and add an else raising a TypeError saying somthing like `f"Cannot retrieve name of variable {x} of type {type(x)}."` Not very different from what you have but I like the informative error message in case we have to debug when something goes wrong.

src/nemos/base_regressor.py
BalzaniEdoardo: yes, you can implement the common saving logic here

src/nemos/base_regressor.py
```diff
@@ -684,3 +764,132 @@ def _optimize_solver_params(self, X: DESIGN_INPUT_TYPE, y: jnp.ndarray) -> dict:
...
+        """
+        pass
+
+    def _save_params_base(
+        self,
+        filename: Union[str, Path],
+        save_attrs: dict,
+    ):
+        """
+        Save model parameters and specified attributes to a .npz file.
```
BalzaniEdoardo, line 705:
```suggestion
        Save model parameters and specified attributes to a .npz file.
        
```

BalzaniEdoardo: like this

src/nemos/base_regressor.py
BalzaniEdoardo: I prefer the class method idea. What about having a `nemos.load_model` function?  We save the model class too, so that we can load directly any saved file without invoking the class method directly.

In this save function we could just retrieve the class method, and calling it.

src/nemos/glm.py
BalzaniEdoardo: since we are using a general `npz` extension, but not all `npz` contain data we can load. If the load fails, we can go over the the `npz` keys, and list the one that we do not recognize. If we are unpacking objects recursively (from more nested to top level), we can keep a list of invalid keys, and then print that.

BalzaniEdoardo (commented):
Generally I agree with Billy's comments, but left a couple of suggestions

src/nemos/base_regressor.py
billbrod: Are you saying have `nemos.load_model` in addition to having class methods (e.g., `nemos.glm.GLM.load`)? I have a slight preference towards just having the class method, because I think a top level `load_model` adds complexity without a clear benefit -- given that we won't have many models, I don't think it's too much to ask users to know how to load their specific model.

src/nemos/glm.py
billbrod: Agreed. Plenoptic has a helper function for this, which might be helpful: https://github.com/plenoptic-org/plenoptic/blob/main/src/plenoptic/tools/io.py

I wasn't thinking about "someone tries to load a file saved with a different library", which is basically what Edoardo is saying, so you would probably want extra info that says something like "these keys look like they belong to this class, these don't" or the like.

src/nemos/base_regressor.py
BalzaniEdoardo: I prefer to have it already because as soon as we will have a second model I would want the `load_model` function.

BalzaniEdoardo: What we'll do here is:
- have a private class method for loading implementing the class-specific loading logic.
- have a single `nemos.load_model` function that calls the appropriate private class method

src/nemos/_observation_model_builder.py
```diff
@@ -25,6 +25,10 @@ def instantiate_observation_model(observation_model: str):
     ValueError
         If the `observation_model` provided does not match to any available observation models.
     """
+    if "." in observation_model:
+        observation_model = observation_model.split(".")[-1].removesuffix(
+            "Observations"
+        )
```
billbrod: We might as well support `PoissonObservations` as well as `nemos.observation_model.PoissonObservations`. So just have the `split` and `removesuffix` act separately

src/nemos/__init__.py
```diff
@@ -19,6 +19,7 @@
     type_casting,
     utils,
 )
+from .io import load_model
```
billbrod:
```suggestion
from .io.load_model import load_model
```

so that we have `nemos.load_model` (not `nemos.load_model.load_model`)

src/nemos/glm.py
```diff
@@ -1158,6 +1159,112 @@ def __sklearn_clone__(self) -> GLM:
...
+
+        >>> # Save the model parameters to a file
+        >>> model.save_params("model_params.npz")
+
+        >>> # Load the model from the saved file
+        >>> model = nmo.load_model("model_params.npz")
+
+        >>> # Print the parameters of the loaded model
+        >>> # This should match the original model parameters
+        >>> print(model.get_params())
```
billbrod: should test this directly.

src/nemos/glm.py
```diff
@@ -1158,6 +1159,112 @@ def __sklearn_clone__(self) -> GLM:
...
+        model = cls(**parameter_config)
+
+        # Set inverse_link_function if available
+        try:
+            if hasattr(model.observation_model, "inverse_link_function"):
+                model.observation_model.inverse_link_function = str(
+                    saved_attrs["inverse_link_function"]
+                )
+        except KeyError:
+            pass
```
billbrod: this section is confusing to me -- let's talk through this more, I feel like it can be simplified.

src/nemos/io/load_model.py
```diff
@@ -0,0 +1,79 @@
...
+
+def load_model(filename: Union[str, Path], mapping_dict: dict = None):
+    """
+    Load a previously saved nemos model from a .npz file.
+
+    Parameters
+    ----------
+    filename :
+        Path to the saved .npz file.
+
```
billbrod:
```suggestion
```

src/nemos/io/load_model.py
```diff
@@ -0,0 +1,79 @@
...
+    mapping_dict :
+        Optional dictionary to map costume attribute names to their actual objects.
```
billbrod: what is "costume attribute"?

src/nemos/io/load_model.py
```diff
@@ -0,0 +1,79 @@
...
+
+    >>> # Save the model parameters to a file
+    >>> model.save_params("model_params.npz")
+
+    >>> # Load the model from the saved file
+    >>> model = nmo.load_model("model_params.npz")
+
+    >>> # Print the parameters of the loaded model
+    >>> # This should match the original model parameters
+    >>> print(model.get_params())
```
billbrod: same point about testing this

src/nemos/io/load_model.py
```diff
@@ -0,0 +1,79 @@
...
+    """
+
+    # load the model from a .npz file
+    filename = Path(filename)
+    if not filename.exists():
+        raise FileNotFoundError(f"File not found: {filename}")
```
billbrod: what happens if we don't check this -- what does `np.load` do if the file doesn't exist?

src/nemos/glm.py
```diff
@@ -1158,6 +1159,112 @@ def __sklearn_clone__(self) -> GLM:
...
+
+        fit_attrs["coef_"] = self.coef_
+        fit_attrs["intercept_"] = self.intercept_
+
+        string_attrs = ["inverse_link_function"]
+
+        super().save_params(filename, fit_attrs, string_attrs)
+
+    @classmethod
+    def _load_from_dict(
```
billbrod: I think this is a nice solution, but we should document somewhere (for ourselves), that this is how `load_model` works, so we know that we need to write this class method for every new model we write

src/nemos/utils.py
```diff
@@ -601,3 +601,147 @@ def _get_terminal_size():
...
+
+def flatten_dict(d: dict, parent_key: str = "") -> dict:
+    """
+    Flatten a nested dictionary into a single-level dictionary with keys representing the hierarchy.
+
+    Parameters
+    ----------
+    d :
+        The dictionary to flatten.
+
```
billbrod:
```suggestion
```

src/nemos/utils.py
```diff
@@ -601,3 +601,147 @@ def _get_terminal_size():
...
+    parent_key :
+        This key starts blank, but recursively it will be filled with the parent key,
+        which is used to create the hierarchy in the flattened dictionary.
+
+    Returns
+    -------
+    dict :
+        A flattened dictionary where the hierarchy is represented by concatenated keys (using __ as a separator).
+    -----------
```
billbrod:
```suggestion
```

src/nemos/utils.py
```diff
@@ -601,3 +601,147 @@ def _get_terminal_size():
...
+                v = np.nan
+            elif not isinstance(v, (str, int, float, bool)):
+                try:
+                    v = np.array(v)
+                except Exception as e:
+                    warnings.warn(
+                        f"Could not convert value '{v}' to numpy array: {e}, "
+                        "setting it to NaN."
+                    )
+                    v = np.nan
```
billbrod: when does this happen?

src/nemos/utils.py
```diff
@@ -601,3 +601,147 @@ def _get_terminal_size():
...
+    Convert a parameter dictionary into serializable format.
+
+    For objects with `get_params`/`set_params`, extracts the class name and
+    parameters. Some attributes are converted to strings to facilitate saving and loading.
+
+    Parameters
+    ----------
+    params_dict :
+        Dictionary of parameters, possibly containing objects.
+
```
billbrod:
```suggestion
```

src/nemos/glm.py
```diff
@@ -1158,6 +1159,112 @@ def __sklearn_clone__(self) -> GLM:
...
+        """
+        Load a GLM model from a dictionary of saved attributes.
+
+        This private method is designed to be called internally by the function load_model.
+
+        Parameters
+        ----------
+        saved_attrs :
+            Dictionary containing the saved attributes of the model.
+
```
billbrod:
```suggestion
```

billbrod (changes requested):
This is looking good! See my following comments, about some bigger picture issues to fix, otherwise most of my comments are nitpicking.

- For the regularizer and observation_model builder, right now, you do `name.split(".")[-1]`. That means that we treat `made_up_module.Unregularized` and `nemos.regularizer.Unregularized` the same. I think we probably don't want that, yes? You could instead do `name = name.split("nemos.regularizer.")[-1]` (and analogously for observation model; in fact you probably want to check that it starts with that string). Then if the name starts with something else, we'll just say "don't know how to handle this".
- The following fails, I think because of issues with the jax function names (link function has name `jax._src.numpy.ufuncs.exp`):
    ```python    
    import nemos as nmo
    glm = nmo.glm.GLM()
    glm.save_params("test.npz")
    nmo.load_model.load_model("test.npz")      
    ```

For how to handle custom constructors, my inclination would be to handle it in the top-level `load_model` function. Before calling `_load_from_dict`, check whether any of the values from `saved_attrs` are keys in `mapping_dict` and, if so, replace with the corresponding object, like:
```python
for k, v in saved_attrs.items():
    if v in mapping_dict:
        saved_attrs[k] = mapping_dict[v]
```

(I don't think the above will actually work because I'm not sure you can edit a dictionary while iterating through it, but hopefully that gets my point across).

Then you'll pass that through to GLM, and the GLM `__init__` method already knows how to handle objects being passed to it.

I think there's a question about whether we only do the above for strings that don't start with `nemos` or `jax` -- that's weird behavior, but maybe that's on the user if they want to do that.

@BalzaniEdoardo what do you think of all the above?

src/nemos/io/load_model.py
pcrespo2017: oh typo, it's custom

src/nemos/io/load_model.py
pcrespo2017: right, it already outputs: FileNotFoundError: File not found: whatever_name

BalzaniEdoardo:
This is looking good. I agree with @billbrod that the load classmethods can be cleaned up a little, and I also agree that the provided mapping should have the priority over the loaded params. We could print or warn about it if we want to be open about what is happeining. If you find a map that is overriding one of our standard mapping, we warn: "hey you are providing a custom mapping, we are using your mapping instead of the standard `key`: `map`"

src/nemos/_observation_model_builder.py
```diff
@@ -25,12 +30,26 @@ def instantiate_observation_model(observation_model: str):
     ValueError
         If the `observation_model` provided does not match to any available observation models.
     """
```
billbrod, line 31: add a line for the ValueError thrown if link function is unknown

src/nemos/_observation_model_builder.py
```diff
@@ -7,14 +7,19 @@
 AVAILABLE_OBSERVATION_MODELS = ["Bernoulli", "Gamma", "Poisson"]
 
 
-def instantiate_observation_model(observation_model: str):
+def instantiate_observation_model(
+    observation_model: str, inverse_link_function: str = None
```
billbrod: callables are also valid for inverse_link_function, so include that in annotation

src/nemos/_observation_model_builder.py
```diff
@@ -7,14 +7,19 @@
...
     """
     Create an observation model from a given name.
 
     Parameters
     ----------
     observation_model:
         The observation model as a string.
+    inverse_link_function:
+        The inverse link function to use for the observation model.
+        If not provided, the default inverse link function of the observation model will be used.
```
billbrod: Should say that this can be either one of the expected strings or any callable. given that there are so many possible strings (and this function is only used internally), I'd just say that the allowed strings are the keys of `nemos.observation_models.LINK_NAME_TO_FUNC`

src/nemos/_observation_model_builder.py
```diff
@@ -7,14 +7,19 @@
 AVAILABLE_OBSERVATION_MODELS = ["Bernoulli", "Gamma", "Poisson"]
 
 
-def instantiate_observation_model(observation_model: str):
+def instantiate_observation_model(
```
billbrod: okay, so the other way to do this would be to have this function just take `observation_model: str` and `**kwargs`, passing all of kwargs to the initialization method (since we're already not doing any processing of them).

That makes this docstring a bit less informative, but I think that's fine -- otherwise, we'll have to revise this whenever we add an observation model with new kwargs.

@BalzaniEdoardo what do you think?

billbrod: (if we do that, ignore my comments about `inverse_link_function` elsewhere on this function

src/nemos/glm.py
```diff
@@ -1159,6 +1160,72 @@ def __sklearn_clone__(self) -> GLM:
...
+        This method allows to reuse the model parameters. The saved parameters can be loaded back
+        into a GLM instance using the `load_params` function.
+
+        Parameters
+        ----------
+        filename :
+            The name of the file where the model parameters will be saved. The file will be saved in `.npz` format.
+
+        Examples
```
billbrod, line 1190: get rid of the empty lines here (it makes the code show up in separate blocks in the [built docs](https://nemos--352.org.readthedocs.build/en/352/generated/glm/nemos.glm.GLM.save_params.html#nemos.glm.GLM.save_params))

src/nemos/glm.py
```diff
@@ -1159,6 +1160,72 @@ def __sklearn_clone__(self) -> GLM:
...
+        >>> # Save the model parameters to a file
+        >>> model.save_params("model_params.npz")
+
+        >>> # Initialize a default GLM
+        >>> model = nmo.glm.GLM()
+
+        >>> # Load the model from the saved file
+        >>> model = nmo.load_model("model_params.npz")
+
+        >>> # Print the parameters of the loaded model
```
billbrod:
```suggestion
        >>> # Model has the same parameters before and after load
```

src/nemos/glm.py
```diff
@@ -1159,6 +1160,72 @@ def __sklearn_clone__(self) -> GLM:
...
+        >>> solver_args = {"stepsize": 0.1, "maxiter": 1000, "tol": 1e-6}
+        >>> model = nmo.glm.GLM(
+        ...     regularizer="Ridge",
+        ...     regularizer_strength=0.1,
+        ...     observation_model="Gamma",
+        ...     solver_name="BFGS",
+        ...     solver_kwargs=solver_args,
+        ... )
+
+        >>> # Print the model parameters
```
billbrod:
```suggestion
```
comment doesn't add anything

src/nemos/io/io.py
```diff
@@ -0,0 +1,219 @@
...
+        Path to the saved .npz file.
+    mapping_dict :
+        Optional dictionary to map custom attribute names to their actual objects.
+
+    Returns
+    -------
+    model :
+        An instance of the model class with the loaded parameters.
+
+    Examples
```
billbrod, line 56: same point about comments and empty newlines as GLM.save_params

src/nemos/io/io.py
```diff
@@ -0,0 +1,219 @@
...
+    for key, value in fit_params.items():
+        if hasattr(model, key):
+            setattr(model, key, value)
+        else:
+            warnings.warn(
+                f"Ignoring unrecognized attribute '{key}' during model loading.{check_str}"
+            )
+
+
+def examine_saved_model(file_path: Union[str, Path]):
```
billbrod: this function also needs to be publicly-accessible. though I'm not sure where -- @BalzaniEdoardo do you want it at `nemos.io.examine_saved_model`? Or `nemos.examine_saved_model`? or somewhere else?

billbrod: (currently it is at `nemos.io.io.examine_saved_model`, but that doesn't feel right)

src/nemos/io/io.py
```diff
@@ -0,0 +1,219 @@
...
+    """Apply mapping dictionary to replace values if keys match."""
+    if not mapping_dict:
+        return params
+
+    for k, v in params.items():
+        try:
+            if v in mapping_dict:
+                try:
+                    params[k] = mapping_dict[v]
+                except Exception as e:
```
billbrod: when can this happen?

src/nemos/io/io.py
```diff
@@ -0,0 +1,219 @@
...
+                    warnings.warn(
+                        f"Failed to replace '{v}' for key '{k}': {e}. "
+                        "Ensure the mapping dictionary is correctly formatted."
+                    )
+        except TypeError:
```
billbrod: when can this happen?

src/nemos/io/io.py
```diff
@@ -0,0 +1,219 @@
...
+            f"`{get_name(examine_saved_model)}('{filename}')` to inspect the saved object."
+        )
+
+    # Set the rest of the parameters as attributes if recognized
+    _set_fit_params(model, fit_params, filename)
+
+    return model
+
+
+def _apply_custom_map(params: dict, mapping_dict: dict) -> dict:
```
billbrod: I'm pretty sure this function is basically the same as `params.update(mapping_dict)`, no? I guess it doesn't add any new keys.

But if someone passes us a dict with new keys for mapping_dict, I think we want to raise an exception. so this should be two steps: check if there are any keys in mapping_dict not in params (and, if so, raise an informative ValueError), then call update.

src/nemos/io/io.py
```diff
@@ -0,0 +1,219 @@
...
+    """Set remaining model attributes, warn if unrecognized."""
+    check_str = (
+        f"\nIf this is confusing, try calling "
+        f"`{get_name(examine_saved_model)}('{filename}')` to inspect the saved object."
+    )
+    for key, value in fit_params.items():
+        if hasattr(model, key):
+            setattr(model, key, value)
+        else:
+            warnings.warn(
```
billbrod: I think we should raise a ValueError here -- not sure when we could end up here that's not a mistake

src/nemos/io/io.py
```diff
@@ -0,0 +1,219 @@
...
+    save_metadata         : <class 'dict'> with length 2 : {'jax_version': '0.4.38', 'nemos_version': '0.2.4.dev59'}
+    model_class           : <class 'str'> : nemos.glm.GLM
+    """
+    file_path = Path(file_path)
+    data = np.load(file_path, allow_pickle=True)
+    data = unflatten_dict(data)
+
+    pad_len = max(len(k) for k in data.keys()) + 2
+
+    for key in data:
```
billbrod: see the plenoptic version of this, but I think this should be prettier -- split it into metadata, model class, initialization params, fit params (in that order). and if the type is a python basic type, I think we don't need to show the type, just print in a way that makes it obvious (wrapped in double quotes, etc).

for a user right now, I think it's confusing to see that `model_class` is a string -- they didn't mean it to be one!

billbrod: the thing to keep in mind is that this is user-facing. we want to make it easy for them to catch simple errors. the most common of which are going to be: jax/nemos version change or wrong model_class. so that info should be forefronted. we want to make it (hopefully) clear to users how they can solve this problem themselves

src/nemos/utils.py
```diff
@@ -601,3 +601,141 @@ def _get_terminal_size():
 def one_over_x(x: NDArray):
     """Implement 1/x."""
     return jnp.power(x, -1)
+
+
+def flatten_dict(d: dict, parent_key: str = "") -> dict:
```
billbrod: should all these new utils be public? My inclination is no -- @BalzaniEdoardo what do you think? (in which case, define __all__ and __dir__ at the top, excluding them)

BalzaniEdoardo: no, I think it's too complicated for our users

src/nemos/io/io.py
BalzaniEdoardo: I would surface it to the same level as the load (first level). I would also choose another name:

`preview_npz_contents`, `inspect_npz`

this can be used to inspect any npz without loading, it is more generic than inspecting the model  parameter npz.

src/nemos/_observation_model_builder.py
BalzaniEdoardo: Let's keep it more general. For the observation model we have so far, there is only the inverse link function, but that's not a given for future implementation. For the negative binomial I am working on, we need a dispersion parameter too for example (a single scalar)

BalzaniEdoardo (commented):
thanks for reviewing @billbrod, let me know if i responded to all the questions

pcrespo2017:
> Thanks Pablo! Most of my comments are about how users will interact with this.
> 
> One bigger question:
> 
> * Right now, `nemos.io` shows up after someone imports nemos. To avoid this, we would need to define `__all__` and `__dir__` in `nemos/__init__.py` -- @BalzaniEdoardo do you think we should do that?

@billbrod @BalzaniEdoardo why don't we want that? should we also hide something else?

billbrod:
> > Thanks Pablo! Most of my comments are about how users will interact with this.
> > One bigger question:
> > 
> > * Right now, `nemos.io` shows up after someone imports nemos. To avoid this, we would need to define `__all__` and `__dir__` in `nemos/__init__.py` -- @BalzaniEdoardo do you think we should do that?
> 
> @billbrod @BalzaniEdoardo why don't we want that? should we also hide something else?

Right now, it's redundant: they can access the function at both `nemos.load_model` and `nemos.io.io.load_model`. That's confusing, so instead I think we want `nemos.load_model` and `nemos.inspect_npz`, as Edoardo said in his comment. (The other stuff that lives within `io.py` are all hidden methods that users shouldn't interact with)

pcrespo2017:
@BalzaniEdoardo @billbrod tox test error seems to be related to the latest scipy version : https://github.com/statsmodels/statsmodels/issues/9584

.github/workflows/connect.yml
```diff
@@ -26,7 +26,6 @@ jobs:
       - name: Install dependencies     # Step to install dependencies
         run: |
           python -m pip install --upgrade pip   # Upgrade pip
-          python -m pip install .                # Install the project
```
billbrod, line 29: I think this line is a relic of Edoardo trying to figure out the issue with flake8, right? should remove this change

src/nemos/io/io.py
```diff
@@ -0,0 +1,228 @@
...
+    regularizer: Lasso()
+    regularizer_strength: 0.1
+    solver_kwargs: {'stepsize': 0.1, 'maxiter': 1000, 'tol': 1e-06}
+    solver_name: ProxSVRG
+    >>> # Test the custom link function does what we expect
+    >>> x = jnp.array([5.0])
+    >>> expected_output = jnp.array([25.0])
+    >>> output = model.observation_model.inverse_link_function(x)
+    >>> print(f"Custom link function output: {output}, Expected: {expected_output}")
+    Custom link function output: [25.], Expected: [25.]
```
billbrod:
```suggestion
    >>> model.observation_model.inverse_link_function(x)
    [25.]
```

my suggestion here does the same thing (testing that the custom link function has the expected output) and is the more standard way to do so in doctests

tests/test_glm.py
```diff
@@ -1599,6 +1603,360 @@ def test_reg_set_params_reg_str_only(
...
+        [
+            "PoissonObservations",
+        ],
+    )
+    @pytest.mark.parametrize(
+        "solver_name",
+        [
+            "ProxSVRG",
+        ],
+    )
```
billbrod, line 1758: these don't need to be parametrized, since there's only one value.

tests/test_glm.py
```diff
@@ -1599,6 +1603,360 @@ def test_reg_set_params_reg_str_only(
...
+
+        # Save
+        save_path = tmp_path / "test_model.npz"
+        model.save_params(save_path)
+
+        # Load
+        loaded_model = nmo.load_model(save_path, mapping_dict=mapping_dict)
+
+        # Validate custom link function in mapping_dict
+        test_value = test_value
```
billbrod: ?

tests/test_glm.py
```diff
@@ -1599,6 +1603,360 @@ def test_reg_set_params_reg_str_only(
...
+    )
+    @pytest.mark.parametrize(
+        "test_value",
+        [
+            jnp.array([25.0]),
+            jnp.array([7]),
+            jnp.array([0.0]),
+        ],
+    )
+    def test_save_load_with_custom_link_function(
```
billbrod: I think should test that `initialize_intercept_matching_mean_rate` works with the squaring for a variety of different values in a different test, which doesn't include saving / loading. Just that if you get a bunch of random floats, that it returns the sqrt

billbrod: and then this test should just check that the loaded model has the right parameters (same as the model before saving), like with the other save/load tests

tests/test_observation_models.py
```diff
@@ -70,20 +74,55 @@ def test_glm_setter_observation_model(obs_model_string, glm_class, expectation):
...
+    ):
         assert isinstance(
             model.observation_model, nmo.observation_models.BernoulliObservations
         )
 
 
+@pytest.mark.parametrize(
+    "link_func_string, expectation",
+    [
+        ("jax.numpy.exp", does_not_raise()),
```
billbrod: really, should check all the link functions that we expect to work. that is, every value in `LINK_NAME_TO_FUNC`. test that the code runs (as here) and that the instantiated link function is the expected one.

tox.ini
```diff
@@ -49,6 +49,8 @@ commands=
     ruff check --fix src/nemos
 
 [testenv:check]
+# helps debugging
+commands_pre = python -c "import sys; print(sys.path)"
```
billbrod: is this also leftover from Edoardo's debuggin? I think it can be removed

billbrod (changes requested):
Still a handful of suggestions. The main thing is that in the example for `load_model` that shows how to use `mapping_dict`, the example still shows user to over-write the regularizer / observation model they used in the saved model. Shouldn't show that: we want the saved and loaded model to be the same, and the example should just show how to use a custom link function with the Poisson observation model. so the only thing in the mapping_dict should be `PoissonObservations(custom_link_function)`

tox.ini
pcrespo2017: should I merge development then?

tests/test_glm.py
pcrespo2017: alright

src/nemos/io/io.py
```diff
@@ -0,0 +1,222 @@
...
+    ...     print(f"{key}: {value}")
+    observation_model__inverse_link_function: <function custom_link_function at ...>
+    observation_model: PoissonObservations(inverse_link_function=custom_link_function)
+    regularizer: Ridge()
+    regularizer_strength: 0.1
+    solver_kwargs: {'stepsize': 0.1, 'maxiter': 1000, 'tol': 1e-06}
+    solver_name: BFGS
+    >>> # Test the custom link function does what we expect
+    >>> x = jnp.array([5.0])
+    >>> print(model.observation_model.inverse_link_function(x))
```
billbrod:
```suggestion
    >>> model.observation_model.inverse_link_function(x)
```
Not necessary

tests/test_observation_models.py
billbrod: I should've been clearer here -- you should be able to import `LINK_NAME_TO_FUNC` at the top of file and just iterate through its keys. That way we don't have to worry about keeping this up to date

tox.ini
billbrod: Yep!

src/nemos/glm.py
```diff
@@ -1159,6 +1160,76 @@ def __sklearn_clone__(self) -> GLM:
...
+        regularizer_strength: 0.1
+        solver_kwargs: {'stepsize': 0.01, 'acceleration': False}
+        solver_name: GradientDescent
+        """
+
+        # initialize saving dictionary
+        fit_attrs = {}
+
+        fit_attrs["coef_"] = self.coef_
+        fit_attrs["intercept_"] = self.intercept_
```
BalzaniEdoardo: we are currently storing the following outputs in glm:
```python
        # initialize to None fit output
        self.intercept_ = None
        self.coef_ = None
        self.solver_state_ = None
        self.scale_ = None
        self.dof_resid_ = None
 ```

File: src/nemos/io/io.py [hunk 62]
```diff
@@ -0,0 +1,223 @@
...
+
+    # load the model from a .npz file
+    filename = Path(filename)
+    data = np.load(filename, allow_pickle=False)
+
+    # Unflatten the dictionary to restore the original structure
+    saved_params = _unflatten_dict(data)
+
+    # "save_metadata" is used to store versions of Nemos and Jax, not needed for loading
+    saved_params.pop("save_metadata")
```
BalzaniEdoardo, line 159: Should we warn/print something if there is a mismatch in the versions? This should not be a problem unless us or JAX make a breaking change.

src/nemos/io/io.py
```diff
@@ -0,0 +1,223 @@
...
+
+def _split_model_params(params: dict, model_class) -> tuple:
+    """Split parameters into config and fit parameters."""
+    model_param_names = model_class._get_param_names()
+    config_params = {k: v for k, v in params.items() if k in model_param_names}
+    fit_params = {k: v for k, v in params.items() if k not in model_param_names}
+    return config_params, fit_params
+
+
+def _set_fit_params(model, fit_params: dict, filename: Path):
```
BalzaniEdoardo: If you were to type the model, should be BaseRegressor, correct? we could type it if you want. To avoid circular imports you can do

```python
from typing import TYPE_CHECKING

if TYPE_CHECKING:
    from .base_regressor import BaseRegressor
    
# ... rest of utils.py
```

src/nemos/observation_models.py
```diff
@@ -17,6 +18,51 @@ def __dir__():
...
+    "jax.scipy.special.expit": jax.scipy.special.expit,
+    "jax._src.scipy.special.expit": jax.scipy.special.expit,
+    "jax.lax.logistic": jax.lax.logistic,
+    "jax._src.lax.lax.logistic": jax.lax.logistic,
+    "jax.scipy.stats.norm.cdf": jax.scipy.stats.norm.cdf,
+    "jax._src.scipy.stats.norm.cdf": jax.scipy.stats.norm.cdf,
+}
+
+
+def create_link_function(link_name: str):
```
BalzaniEdoardo: maybe name it `link_function_from_string`?

src/nemos/utils.py
```diff
@@ -573,3 +587,142 @@ def _get_terminal_size():
 def one_over_x(x: NDArray):
     """Implement 1/x."""
     return jnp.power(x, -1)
+
+
+def _flatten_dict(d: dict, parent_key: str = "") -> dict:
```
BalzaniEdoardo: can you rename the dictionary as `nested_dict` instead of `d`?

src/nemos/utils.py
```diff
@@ -573,3 +587,142 @@ def _get_terminal_size():
...
+        if isinstance(v, dict):
+            items.extend(_flatten_dict(v, new_key).items())
+        else:
+            # None values and non-standard types are converted to numpy
+            if v is None:
+                v = np.nan
+            elif not isinstance(v, (str, int, float, bool)):
+                v = np.array(v)
+            items.append((new_key, v))
+    return dict(items)
```
BalzaniEdoardo, line 626: nice, very clean

src/nemos/utils.py
```diff
@@ -573,3 +587,142 @@ def _get_terminal_size():
...
+
+
+def _unflatten_dict(d: dict) -> dict:
+    """
+    Unflatten a dictionary with keys representing hierarchy into a nested dictionary.
+
+    Parameters
+    ----------
+    d :
```
BalzaniEdoardo: same here, maybe call it `"flat_dict`

src/nemos/utils.py
```diff
@@ -573,3 +587,142 @@ def _get_terminal_size():
...
+        The dictionary to unflatten.
+
+    Returns
+    -------
+    dict :
+        A nested dictionary with the original hierarchy restored.
+    """
+
+    sep = "__"
+    result = {}
```
BalzaniEdoardo: call it `nested_dict`

src/nemos/utils.py
```diff
@@ -573,3 +587,142 @@ def _get_terminal_size():
...
+    Parameters
+    ----------
+    params_dict :
+        Dictionary of parameters, possibly containing objects.
+    string_attrs :
+        List of attributes that should be converted to strings (e.g., `inverse_link_function`).
+
+    Returns
+    -------
+    dict :
```
BalzaniEdoardo, line 727:
```suggestion
    out :
```

BalzaniEdoardo (changes requested):
I reviewed the implementation, it is looking very clean, I like it a lot. I left a couple of comments here and there but I feel this is almost done.

Once we are happy with the implementation I'll give a round of reviewing the tests.

src/nemos/glm.py
BalzaniEdoardo: For what it concerns the `solver_state_`, currently it is a named tuple, but soon we will transition to optimistix, and the state will become a dataclass unfortunately. So I wonder if we should just ignore that? what do you guys think @pcrespo2017 @billbrod ?

BalzaniEdoardo: Added a `_get_fit_state` method in `BaseRegressor` that uses the naming convention to get the states and used that dictionary at save time.

src/nemos/io/io.py
BalzaniEdoardo: I ended up adding then removing the warning. We should document how to debug issues related to nemos version compatibility

docs/how_to_guide/save_and_load.md
```diff
@@ -0,0 +1,103 @@
...
+  display_name: Python 3
+  language: python
+  name: python3
+---
+
+# Saving and Loading
+
+## Saving and Loading a Model
+
+In nemos, you can save a model by calling the `save_params` method, which writes `npz` file, a NumPy specific binary.
```
billbrod: link to numpy docs about npz

docs/how_to_guide/save_and_load.md
```diff
@@ -0,0 +1,103 @@
...
+    regularizer="Ridge",
+    solver_name="GradientDescent"
+)
+
+# save
+model.save_params("ridge_glm_params.npz")
+model
+```
+
+The output file is loaded without pickling, because NeMoS does not store objects directly. Instead, it stores only a string representing the object’s class and its parameters. This approach is both more robust and safer than pickling the objects themselves.
```
billbrod: I think this should be an admonition? or mentioned later on -- if users are only using built-in nemos objects, they don't need to worry about this

billbrod: I'm also not sure about it being "more robust" -- depends on your definition of robust

docs/how_to_guide/save_and_load.md
```diff
@@ -0,0 +1,103 @@
...
+
+```{code-cell}
+
+loaded_model = nmo.load_model("ridge_glm_params.npz")
```
billbrod: should show that the loaded model is the same as the saved.

also probably show this works both before and after fitting?

docs/how_to_guide/save_and_load.md
```diff
@@ -0,0 +1,103 @@
...
+loaded_model
+```
+
+## Inspecting the `npz`
+You can check the content of the stored `npz` by calling the `nemos.inspect_npz` function.
```
billbrod: should probably mention this is meant largely to help debug, if something goes wrong

docs/how_to_guide/save_and_load.md
```diff
@@ -0,0 +1,103 @@
...
+
+```{code-cell}
+
+nmo.inspect_npz("ridge_glm_params.npz")
+```
+
+
+## Save and Load Custom Objects
+
+Advanced users may want to specify custom models and still be able to save and load. For example, one could try a different inverse link function (non-linearity) or a custom  `Regularizer`.
```
billbrod, line 90: we don't have an example of this in the docs right now, do we? else, should link there

docs/how_to_guide/save_and_load.md
billbrod: actually, I think remove this line entirely, the description in the section on save/load custom objects is good enough I think

docs/how_to_guide/save_and_load.md
```diff
@@ -0,0 +1,103 @@
...
+        self.new_param = new_param
+
+obs_model = nmo.observation_models.PoissonObservations(custom_link)
+model = nmo.glm.GLM(regularizer=CustomRegularizer(10))
+model.save_params("custom_regularizer_params.npz")
+
+nmo.inspect_npz("custom_regularizer_params.npz")
+```
+
+As you can see, the regularizer class is stored as a string, `"{object_class.__module__}.{object_class.__name__}"`. This means that trying to load this model directly will result in an error, because NeMoS doesn’t pickle objects and therefore doesn’t know how to recreate the `CustomRegularizer` automatically.
```
billbrod, line 107: add a small admonition pointing out this is for safety, can link to [here](https://news.ycombinator.com/item?id=41901475) if you want an example of safety issues, and I'm sure we can find a link that describes issues with changing versions as well.

docs/how_to_guide/save_and_load.md
```diff
@@ -0,0 +1,103 @@
...
+
+```{code-cell}
+:tags: [raises-exception]
+
+loaded_model = nmo.load_model("custom_regularizer_params.npz")
+```
+
+You can circumvent this error, by providing a mapping to the class and to the callable.
```
billbrod:
```suggestion
You can tell nemos how to load the custom objects by providing a mapping between the saved string and to the callable.
```
it's not really an error

docs/how_to_guide/save_and_load.md
```diff
@@ -0,0 +1,103 @@
...
+mapping = {
+    "regularizer": CustomRegularizer,
+    "observation_model__inverse_link_function": custom_link
+}
+loaded_model = nmo.load_model("custom_regularizer_params.npz", mapping_dict=mapping)
+loaded_model
+```
+::: {tip}
+
+Always keep your custom classes and functions in importable modules if you plan to load saved models later. This makes it easier to share models with collaborators or run them on different machines.
```
billbrod: I worry this is a bit cryptic for most users, so I'd remove it

docs/how_to_guide/save_and_load.md
```diff
@@ -0,0 +1,103 @@
...
+:::
+
+:::{admonition} Allowed Mappings
+:class: warning
+
+Mapping is allowed **only** for callables (functions) and classes, because these cannot be stored directly without pickling.
+Other values (like numbers, strings, or arrays) are always stored directly in the `.npz` and cannot be remapped.
+
+⚠️ When mapping a custom class, you must provide the class **type**, not an instance (e.g., `mapping = {"regularizer": CustomRegularizer}` is valid, but `CustomRegularizer()` is not).
```
billbrod: really what you're passing is the initialization function. not sure if that distinction matters here

src/nemos/glm.py
```diff
@@ -1171,6 +1172,76 @@ def __sklearn_clone__(self) -> GLM:
...
+        >>> for key, value in model.get_params().items():
+        ...     print(f"{key}: {value}")
+        observation_model__inverse_link_function: <function one_over_x at ...>
+        observation_model: GammaObservations(inverse_link_function=one_over_x)
+        regularizer: Ridge()
+        regularizer_strength: 0.1
+        solver_kwargs: {'stepsize': 0.1, 'maxiter': 1000, 'tol': 1e-06}
+        solver_name: BFGS
+        >>> # If you want to load the model with a custom mapping for callables and custom classes,
+        >>> # you can use the mapping_dict parameter.
```
billbrod: this is not what we want -- it shows how to overwrite the saved parameters. we instead want a separate block (so insert some blank newlines), that shows how to save/load a custom link function.

src/nemos/io/io.py
billbrod, line 1: This file is not included in the API docs, which it should be.

src/nemos/io/io.py
```diff
@@ -0,0 +1,442 @@
...
+    >>> # Model has the same parameters before and after load
+    >>> for key, value in model.get_params().items():
+    ...     print(f"{key}: {value}")
+    observation_model__inverse_link_function: <function one_over_x at ...>
+    observation_model: GammaObservations(inverse_link_function=one_over_x)
+    regularizer: Ridge()
+    regularizer_strength: 0.1
+    solver_kwargs: {'stepsize': 0.1, 'maxiter': 1000, 'tol': 1e-06}
+    solver_name: BFGS
+    >>> # If you want to load the model with a custom mapping for callables and custom classes,
```
billbrod: same point about overwriting saved parameters

src/nemos/io/io.py
```diff
@@ -0,0 +1,442 @@
...
+        return instantiate_observation_model(class_name, **kwargs)
+    else:
+        # Hit when loading a custom class without mapping
+        raise ValueError(
+            f"The class '{class_basename}' is not a native NeMoS class.\n"
+            "To load custom observation or regularization classes, please provide a mapping:\n\n"
+            "  - For observation models:\n"
+            "      nemos.load_model(save_path, mapping_dict={'observation_model': YourCustomObservations})\n"
+            "  - For regularization:\n"
+            "      nemos.load_model(save_path, mapping_dict={'regularization': YourCustomRegularizer})"
```
billbrod: can we be more specific in this error message? we know the name of the key and the value that they should pass, right? so can't we tell them exactly what how to call load_model?

tox.ini
billbrod:
```suggestion
```
right? or is this useful to keep around?

docs/how_to_guide/save_and_load.md
BalzaniEdoardo: ok, I removed the text and joined the two blocks of code

docs/how_to_guide/save_and_load.md
BalzaniEdoardo: I don't think we have an example of custom regularizer or link function

docs/how_to_guide/save_and_load.md
BalzaniEdoardo: I think this is formally correct but too specific. I rephrased this but still saying they need to provide the class

src/nemos/io/io.py
BalzaniEdoardo: done

docs/how_to_guide/save_and_load.md
```diff
@@ -0,0 +1,145 @@
...
+For a real-world example, see [this discussion](https://news.ycombinator.com/item?id=41901475).
+:::
+
+```{code-cell}
+:tags: [raises-exception]
+
+loaded_model = nmo.load_model("custom_regularizer_params.npz")
+```
+
+You can tell nemos how to load the custom objects by providing a mapping between the saved string and to the callable.
```
billbrod:
```suggestion
As the error explains, you can tell nemos how to load the custom objects by providing a mapping between the saved string and to the callable.
```

docs/how_to_guide/save_and_load.md
```diff
@@ -0,0 +1,145 @@
...
+loaded_model
+```
+
+:::{admonition} Allowed Mappings
+:class: warning
+
+Mapping is allowed **only** for callables (functions) and classes, because these cannot be stored directly without pickling.
+Other values (like numbers, strings, or arrays) are always stored directly in the `.npz` and cannot be remapped.
+
+⚠️ hen mapping a custom class, you must pass the **class itself** (e.g., `mapping = {"regularizer": CustomRegularizer}`), not an instance (`CustomRegularizer()`).
```
billbrod:
```suggestion
⚠️ When mapping a custom class, you must pass the **class itself** (e.g., `mapping = {"regularizer": CustomRegularizer}`), not an instance (`CustomRegularizer()`).
```

billbrod: I'm not sure if we want a hazard sign inside an admonition that already uses it, feels redundant

PR #356, opened 2025-06-23
add elastic net regularizer

sjvenditto:
# ElasticNet regularizer
This PR adds the elastic net regularization scheme by adding a new regularizer class, `nmo.regularizer.ElasticNet`. 

A key difference between this regularizer and existing regularizers is that in now requires a tuple of two values passed to `regularizer_strength`, the first for the *regularizer strength* and the second for the *regularizer ratio*, which is a value between 0 and 1 that defines the balance between $L_1$ (value of $1$) and $L_2$ (value of $0$) regularization. The default is $0.5$.
```
nmo.glm.GLM(regularizer="ElasticNet", regularizer_strength=(1.0,0.5))
```
A single float is still accepted for setting the regularizer strength only, which takes the default value for the ratio.
## Notable changes
- The validation of `regularizer_strength` was moved to be a `Regularizer` class method, `_validate_regularizer_strength`. This method handles type checking and setting the default value. Every class uses the default method (checking for a float and setting `regularizer_strength=1.0` as default) except:
	- `UnRegularized`, which now forces the regularization strength to always be `None`
	- `ElasticNet`, which adds additional checks for one or two values for regularizer strength and ratio, setting the default to `(1.0, 0.5)`
- Switching between any regularizer will reset the regularizer strength to its default if it is not specified 
- Following default behavior of scikit-learn and statsmodels, NeMoS no longer throws a warning when the user doesn't set `regularizer_strength`
	- A warning is still passed when trying to set `regularizer_strength` with the regularizer set to `UnRegularized`
	- A new warning is thrown when only a single value is passed to `regularizer_strength` for `ElasticNet` to highlight the change of syntax. As stated above, a single value is still accepted as a valid input

### DIFF
--- a/src/nemos/regularizer.py
+++ b/src/nemos/regularizer.py
@@ -339,8 +339,8 @@ class ElasticNet(Regularizer):
     The Elasitc Net penalty [3]_ [4]_ is defined as:
 
     .. math::
-        P(\beta) = \alpha ((1 - \lambda) \frac{1}{2} ||\beta||_{\ell_2}^2 +
-        \lambda ||\beta||_{\ell_1}
+        P(\beta) = \alpha \left((1 - \lambda) \frac{1}{2} ||\beta||_{\ell_2}^2 +
+        \lambda ||\beta||_{\ell_1} \right)
 
     where :math:`\alpha` is the regularizer strength, and :math:`\lambda` is the regularizer ratio.
     The regularizer ratio controls the balance between L1 (Lasso) and L2 (Ridge)


docs/quickstart.md
```diff
@@ -297,6 +297,14 @@ glm = nmo.glm.GLM(regularizer="Ridge", regularizer_strength=0.1)
 
 ```
 
+For [ElasticNet](nemos.regularizer.ElasticNet) regularization, you can also specify a regularizer ratio, specifies the balance between $L_1$ and $L_2$ regularization, where a ratio of 1 is equivalent to $L_1$ (Lasso) and a value of 0 is equivalent to $L_2$ (Ridge). This can be set by passing a tuple of values to `regularizer_strength`, where the first value is the strength and the second is the ratio.
```
BalzaniEdoardo:
```suggestion
For [ElasticNet](nemos.regularizer.ElasticNet) regularization, you can also set a regularizer ratio, specifying the balance between $L_1$ and $L_2$ regularization, where a ratio of 1 is equivalent to $L_1$ (Lasso) and a value of 0 is equivalent to $L_2$ (Ridge). This can be set by passing a tuple of values to `regularizer_strength`, where the first value is the strength and the second is the ratio.
```

src/nemos/proximal_operator.py
```diff
@@ -184,3 +184,47 @@ def fun(u, v):
...
+    r"""Proximal operator for the elastic net.
+
+    .. math::
+
+      \underset{y}{\text{argmin}} ~ \frac{1}{2} ||x - y||_2^2
+      + \text{scaling} \cdot \text{hyperparams[0]} \cdot g(y)
+
+    where :math:`g(y) = ||y||_1 + \text{hyperparams[1]} \cdot 0.5 \cdot ||y||_2^2`.
+
+    Args:
```
BalzaniEdoardo: Can you change this to numpydoc format?

src/nemos/regularizer.py
```diff
@@ -307,6 +332,175 @@ def _penalized_loss(params, X, y):
...
+            # since we do not allow array regularization assume we pass a tuple
+            regularizer_strength, regularizer_ratio = netreg
+            regularizer_strength /= bs.shape[0]
+            lam = regularizer_strength * regularizer_ratio  # hyperparams[0]
+            gam = (1 - regularizer_ratio) / regularizer_ratio  # hyperparams[1]
+            # if Ws is a pytree, netreg needs to be a pytree with the same
+            # structure
+            lam = jax.tree_util.tree_map(lambda x: lam * jnp.ones_like(x), Ws)
+            gam = jax.tree_util.tree_map(lambda x: gam * jnp.ones_like(x), Ws)
+            return jaxopt.prox.prox_elastic_net(Ws, (lam, gam), scaling=scaling), bs
```
BalzaniEdoardo: are we still using this or `nemos.proximal_operator.prox_elastic_net`?

src/nemos/regularizer.py
```diff
@@ -307,6 +332,175 @@ def _penalized_loss(params, X, y):
...
+    @staticmethod
+    def _penalization(
+        params: Tuple[DESIGN_INPUT_TYPE, jnp.ndarray],
+        net_regularization: Tuple[float, float],
+    ) -> jnp.ndarray:
+        r"""
+        Compute the Elastic Net penalization for given parameters. The elasitc net penalty is defined as:
+
+        .. math::
+            P(\beta) = \alpha * ((1 - \lambda) * \frac{1}{2} ||\beta||_{\ell_2}^2 +
```
BalzaniEdoardo: should * be replaced with `\cdot`?

src/nemos/regularizer.py
```diff
@@ -307,6 +332,175 @@ def _penalized_loss(params, X, y):
...
+        :math:`\lambda = 1` is equivalent to Lasso regularization.
+
+        Parameters
+        ----------
+        params :
+            Model parameters for which to compute the penalization.
+
+        Returns
+        -------
+        float
```
BalzaniEdoardo:
```suggestion
        :
```

tests/test_regularizer.py
```diff
@@ -230,15 +238,14 @@ def test_regularizer_strength_none(self):
...
-            model.regularizer = "Lasso"
+        # changing to ridge, lasso, or grouplasso should set to 1.0
+        model.regularizer = "Lasso"
 
         assert model.regularizer_strength == 1.0
 
         with pytest.warns(UserWarning):
             model.regularizer = regularizer
-        assert model.regularizer_strength == 1.0
+        assert model.regularizer_strength == None
```
BalzaniEdoardo, line 320: when is this set now?

tests/test_regularizer.py
```diff
@@ -964,6 +968,357 @@ def test_solver_combination(self, solver_name, poissonGLM_model_instantiation):
...
+        alpha_sm = np.ones(X.shape[1] + 1) * reg_strength
+        alpha_sm[0] = 0
+
+        # pure lasso = elastic net with L1 weight = 1
+        res_sm = glm_sm.fit_regularized(
+            method="elastic_net",
+            alpha=alpha_sm,
+            L1_wt=reg_ratio,
+            cnvrg_tol=10**-12,
+            zero_tol=1e-1000,
```
BalzaniEdoardo, line 1339: lol

tests/test_regularizer.py
```diff
@@ -964,6 +968,357 @@ def test_solver_combination(self, solver_name, poissonGLM_model_instantiation):
...
+            maxiter=10000,
+        )
+        # assert weights are the same
+        assert np.allclose(res.x, glm_res)
+        assert np.allclose(res.x, res_sm.params)
+        assert np.allclose(glm_res, res_sm.params)
+
+    def test_lasso_pytree(self, poissonGLM_model_instantiation_pytree):
```
BalzaniEdoardo: Are these supposed to test elastic net?

tests/test_regularizer.py
```diff
@@ -964,6 +968,357 @@ def test_solver_combination(self, solver_name, poissonGLM_model_instantiation):
...
+        """Check pytree X can be fit."""
+        X, y, model, true_params, firing_rate = poissonGLM_model_instantiation_pytree
+        model.set_params(regularizer=nmo.regularizer.Lasso(), regularizer_strength=1.0)
+        model.solver_name = "ProximalGradient"
+        model.fit(X, y)
+
+    @pytest.mark.parametrize("solver_name", ["ProximalGradient", "ProxSVRG"])
+    @pytest.mark.parametrize("reg_str", [0.001, 0.01, 0.1, 1, 10])
+    def test_lasso_pytree_match(
```
BalzaniEdoardo: is this Lasso?

tests/test_regularizer.py
sjvenditto: now with the class specific validation of regularizer strength, I made the `UnRegularized` class always return `None` for strength regardless of what is passed. this way we can skip type checking as nothing will ever be saved. we can change this behavior if you want it to match the old handling, which still stored the value if it passed type checks

src/nemos/regularizer.py
sjvenditto: we can use the nemos version

src/nemos/regularizer.py
sjvenditto: I just removed them completely, it was a relic from when I had regularizer_strength in place of $\alpha$ and regularizer_ratio in place of $\lambda$, which I replaced to make the equation not as long in the docs

tests/test_regularizer.py
sjvenditto: yes, fixed

tests/test_regularizer.py
sjvenditto: fixed

src/nemos/proximal_operator.py
sjvenditto: fixed

Issue #359, opened 2025-06-25
Jointly learn dispersion and GLM coefficients for NB distributiuon


BalzaniEdoardo:
As in `glm.nb` of R, we could set up an iterative procedure for fitting both parameters. The procedure simply alternates between updating the scale by fixing the coefficients and fixing the scale and updating the coefficients.

The C implementation of the algorithm can be found here:
https://github.com/gersteinlab/nb-fit/blob/master/glm.nb.cpp#L697

Should be pretty straight forward, but we should consider implementation details:

- Create a dedicated object for that or not? (instead of the standard GLM class). Probably yes, since the fit is more computationally inefficient (and equivalent to fitting multiple GLMs).
- Think about how general this class should be. For non canonical exp-family distribution there will be a similar entanglement of the scale and the mean of the distribution, requiring a joint estimation procedure. We could predispose the class to be ready to accommodate other distribution types (setting up a general mechanism for an alternate optimization).

PR #360, opened 2025-06-25
Negative binomial

BalzaniEdoardo:
## Negative Binomial Observation Model
Add a negative binomial observation model.

- **NegativeBinomialObservations**: new class implementing the negative binomial observations.

### Key feature
The observation model initialization has an extra parameter, `scale`, set by default at `1.0`, this is an arbitrary value but matches what `statsmodels` do for `NegativeBinomial`.  This parameter corresponds to `1/r` where `r` is the number of successes of the Negative Binomial. We used the same parametrization as [statsmodels](https://www.statsmodels.org/dev/generated/statsmodels.genmod.families.family.NegativeBinomial.html), where their `alpha` parameter corresponds to `scale`. 

 
**Why do we need an extra parameter at intialization? (And we do not need it for linear regression and `Gamma` GLM.)**
The scale parameter of a `NegativeBinomial` GLM interacts with the point estimate of the coefficients (i.e. different values of the scale result in different values for the optimal GLM coefficients).  The `scale` regulates how much over-dispersion we are assuming, and therefore is a key parameter for this observation model and must be surfaced. On the other hand, the scale of a `Gamma` GLM (and the noise variance in a linear regression) do not affect the estimate of the coefficients, and can be left unspecified at fit time. (This is why we do not require a scale parameter at `GammaObservations` initialization). 

### Learning jointly the `scale` and the GLM coefficient
In a subsequent PR we could implement a joint optimization procedure. See issue #359.

### DIFF
--- a/docs/quickstart.md
+++ b/docs/quickstart.md
@@ -295,7 +295,6 @@ To set up a Negative Binomial GLM, provide this `observation_model` at initializ
 import nemos as nmo
 
 # set up a Negative Binomial GLM for modeling counts (with scale = 1)
-# note: you should tune the scale parameter (for example, through cross-validation)
 glm = nmo.glm.GLM(observation_model=nmo.observation_models.NegativeBinomialObservations(scale=1))
 
 ```
--- a/docs/quickstart.md
+++ b/docs/quickstart.md
@@ -287,7 +287,8 @@ glm = nmo.glm.GLM(observation_model=nmo.observation_models.BernoulliObservations
 
 A [Negative Binomial observation model](nemos.observation_models.NegativeBinomialObservations) is appropriate for describing a counting process (such as neural spike counts) that exhibits supra-Poisson variability — that is, it is over-dispersed relative to a Poisson model (equivalently, the variance of the counts is larger than the mean).
 
-To set up a Negative Binomial GLM, provide this `observation_model` at initialization. You can regulate the degree of overdispersion by setting the `scale` parameter: the larger the `scale`, the more dispersed the Negative Binomial will be compared to a Poisson model.
+To set up a Negative Binomial GLM, provide this `observation_model` at initialization. The degree of overdispersion can be controlled via the `scale` parameter: larger values of `scale` result in greater variability compared to a Poisson model, while as `scale` approaches zero, the distribution converges to a Poisson. An appropriate value for `scale` depends on the distribution of the observed spike counts and should be tuned, for example through cross-validation.
+
 
 ```{code-cell} ipython3
 
--- a/docs/quickstart.md
+++ b/docs/quickstart.md
@@ -287,7 +287,7 @@ glm = nmo.glm.GLM(observation_model=nmo.observation_models.BernoulliObservations
 
 A [Negative Binomial observation model](nemos.observation_models.NegativeBinomialObservations) is appropriate for describing a counting process (such as neural spike counts) that exhibits supra-Poisson variability — that is, it is over-dispersed relative to a Poisson model (equivalently, the variance of the counts is larger than the mean).
 
-To set up a Negative Binomial GLM, provide this `observation_model` at initialization. The degree of overdispersion can be controlled via the `scale` parameter: larger values of `scale` result in greater variability compared to a Poisson model, while as `scale` approaches zero, the distribution converges to a Poisson. An appropriate value for `scale` depends on the distribution of the observed spike counts and should be tuned, for example through cross-validation.
+To set up a Negative Binomial GLM, provide this `observation_model` at initialization. The degree of overdispersion can be controlled via the `scale` parameter: larger values of `scale` result in greater variability compared to a Poisson model, while as `scale` approaches zero, the distribution converges to a Poisson. An appropriate value for `scale` depends on the distribution of the observed spike counts and should be tuned, for example through [cross-validation](how_to_guide/plot_06_sklearn_pipeline_cv_demo).
 
 
 ```{code-cell} ipython3
--- a/docs/quickstart.md
+++ b/docs/quickstart.md
@@ -287,7 +287,7 @@ glm = nmo.glm.GLM(observation_model=nmo.observation_models.BernoulliObservations
 
 A [Negative Binomial observation model](nemos.observation_models.NegativeBinomialObservations) is appropriate for describing a counting process (such as neural spike counts) that exhibits supra-Poisson variability — that is, it is over-dispersed relative to a Poisson model (equivalently, the variance of the counts is larger than the mean).
 
-To set up a Negative Binomial GLM, provide this `observation_model` at initialization. The degree of overdispersion can be controlled via the `scale` parameter: larger values of `scale` result in greater variability compared to a Poisson model, while as `scale` approaches zero, the distribution converges to a Poisson. An appropriate value for `scale` depends on the distribution of the observed spike counts and should be tuned, for example through [cross-validation](how_to_guide/plot_06_sklearn_pipeline_cv_demo).
+To set up a Negative Binomial GLM, provide this `observation_model` at initialization. The degree of overdispersion can be controlled via the `scale` parameter: larger values of `scale` result in greater variability compared to a Poisson model, while as `scale` approaches zero, the distribution converges to a Poisson. An appropriate value for `scale` depends on the distribution of the observed spike counts and should be tuned, for example through [cross-validation](sklearn-how-to).
 
 
 ```{code-cell} ipython3
--- a/docs/quickstart.md
+++ b/docs/quickstart.md
@@ -245,7 +245,10 @@ X_multi.shape
 
 For additional information on one-dimensional convolutions, see [here](convolution_background).
 
-## **Continuous Observations**
+
+## **Other Observation Models**
+
+### **Continuous Observations**
 
 
 By default, NeMoS' GLM uses [Poisson observations](nemos.observation_models.PoissonObservations), which are a natural choice for spike counts. However, the package also supports a [Gamma](nemos.observation_models.GammaObservations) GLM, which is more appropriate for modeling continuous, non-negative observations such as calcium transients.
@@ -265,7 +268,7 @@ glm = nmo.glm.GLM(observation_model=nmo.observation_models.GammaObservations())
 
 Take a look at our [tutorial](tutorial-calcium-imaging) for a detailed example.
 
-## **Binary Observations**
+### **Binary Observations**
 
 
 NeMoS additionally supports a [Bernoulli observation model](nemos.observation_models.BernoulliObservations), which is useful for modeling binary observations, such as a binary choices or singular spikes. To use this observation model, set the `observation_model` argument during model initialization.
@@ -279,6 +282,24 @@ import nemos as nmo
 glm = nmo.glm.GLM(observation_model=nmo.observation_models.BernoulliObservations())
 
 ```
+
+### **Counts with Supra-Poisson Variability**
+
+A [Negative Binomial observation model](nemos.observation_models.NegativeBinomialObservations) is appropriate for describing a counting process (such as neural spike counts) that exhibits supra-Poisson variability — that is, it is over-dispersed relative to a Poisson model (equivalently, the variance of the counts is larger than the mean).
+
+To set up a Negative Binomial GLM, provide this `observation_model` at initialization. You can regulate the degree of overdispersion by setting the `scale` parameter: the larger the `scale`, the more dispersed the Negative Binomial will be compared to a Poisson model.
+
+```{code-cell} ipython3
+
+import nemos as nmo
+
+# set up a Negative Binomial GLM for modeling counts (with scale = 1)
+# note: you should tune the scale parameter (for example, through cross-validation)
+glm = nmo.glm.GLM(observation_model=nmo.observation_models.NegativeBinomialObservations(scale=1))
+
+```
+
+
 
 ## **Regularization**
 
--- a/src/nemos/observation_models.py
+++ b/src/nemos/observation_models.py
@@ -1382,7 +1382,7 @@ class NegativeBinomialObservations(Observations):
         Parameters
         ----------
         observations:
-            Observed count data. Shape (n_time_bins,) or (n_time_bins, n_observations).
+            Observed count data. Shape ``(n_time_bins,)`` or ``(n_time_bins, n_observations).``
         predicted_rate:
             Predicted mean count of the Negative Binomial distribution. Shape matches `observations`.
         scale:
@@ -1391,7 +1391,7 @@ class NegativeBinomialObservations(Observations):
         Returns
         -------
         :
-            The residual deviance of the model. Shape matches `observations`.
+            The residual deviance of the model. Shape matches ``observations``.
 
         Notes
         -----
--- a/src/nemos/observation_models.py
+++ b/src/nemos/observation_models.py
@@ -1407,10 +1407,11 @@ class NegativeBinomialObservations(Observations):
         dof_resid: Union[float, jnp.ndarray],
     ) -> Union[float, jnp.ndarray]:
         r"""
-        Assign 1 to the scale parameter of the Bernoulli model.
+        Return the scale parameter of the distribution.
 
-        For the Binomial exponential family distribution (to which the Bernoulli belongs), the scale parameter
-        :math:`\phi` is always 1.
+        The `scale` parameter of this distribution is set at initialization and
+        affect the likelihood landscape. This implies that the `scale`
+        parameter cannot be estimated post-hoc without re-fitting a model.
 
         Parameters
         ----------
@@ -1422,5 +1423,4 @@ class NegativeBinomialObservations(Observations):
         dof_resid :
             The DOF of the residuals.
         """
-        resid = y - predicted_rate
-        return ((resid**2 / predicted_rate - 1) / predicted_rate).sum() / dof_resid
+        return self.scale
--- a/src/nemos/observation_models.py
+++ b/src/nemos/observation_models.py
@@ -1174,7 +1174,7 @@ class NegativeBinomialObservations(Observations):
     This model represents a Negative Binomial distribution [3]_ commonly used to model overdispersed
     count data [4]_ [5]_ (i.e., data where the variance exceeds the mean), which cannot be captured by a
     standard Poisson model. The distribution is parameterized by the predicted mean rate
-    (:math:`\mu`) and a fixed dispersion parameter (:math:`\phi`) or "scale" of the model.
+    (:math:`\mu`) and a fixed dispersion parameter (:math:`\phi`) or `scale` of the model.
 
     **Important:** the scale parameter must be estimated from the data for accurately capturing the
     dispersion. In the context of NeMoS GLM, estimation can be achieved by cross-validating the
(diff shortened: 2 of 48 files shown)


src/nemos/observation_models.py
```diff
@@ -1165,3 +1165,271 @@ def estimate_scale(
...
+    Poisson distribution. This makes the model flexible for handling both equidispersed
+    (Poisson-like) and overdispersed data.
+
+    Parameters
+    ----------
+    inverse_link_function :
+        Inverse of the link function used to map linear predictors to the predicted mean rate.
+        For count data, this is typically `jax.numpy.exp` or `jax.nn.softplus`.
+    scale :
+        The dispersion parameter φ. Lower values correspond to lower overdispersion, and as
```
billbrod: we should probably be consistent on whether to use unicode in these docstrings vs. sphinx's `:math:` directive

src/nemos/observation_models.py
```diff
@@ -1165,3 +1165,271 @@ def estimate_scale(
             The DOF of the residuals.
         """
         return jnp.ones_like(jnp.atleast_1d(y[0]))
+
+
+class NegativeBinomialObservations(Observations):
+    r"""
+    A Negative Binomial model for overdispersed count data using mean-dispersion parameterization.
```
billbrod, line 1172: Since this is a less common observation model, should we add some links / references? both to general math about negative binomial and also to papers showing how it's used in neuroscience?

src/nemos/observation_models.py
```diff
@@ -1165,3 +1165,271 @@ def estimate_scale(
...
+
+        Parameters
+        ----------
+        y :
+            Observed count data. Shape ``(n_time_bins,)`` or ``(n_time_bins, n_observations)``.
+        predicted_rate :
+            The predicted mean of the Negative Binomial distribution. Shape
+            ``(n_time_bins,)`` or ``(n_time_bins, n_observations)``.
+        aggregate_sample_scores :
+            Function that aggregates the log-likelihood across samples (e.g., jnp.mean or jnp.sum).
```
billbrod: do you want to either cross-ref jax.numpy or put `jnp.mean`/`jnp.sum` in monospace?

src/nemos/observation_models.py
```diff
@@ -1165,3 +1165,271 @@ def estimate_scale(
...
+        observations: jnp.ndarray,
+        predicted_rate: jnp.ndarray,
+        scale: Union[float, jnp.ndarray, None] = None,
+    ) -> jnp.ndarray:
+        r"""Compute the residual deviance for a Negative Binomial model.
+
+        Parameters
+        ----------
+        observations:
+            Observed count data. Shape (n_time_bins,) or (n_time_bins, n_observations).
```
billbrod:
```suggestion
            Observed count data. Shape ``(n_time_bins,)`` or ``(n_time_bins, n_observations).``
```

src/nemos/observation_models.py
```diff
@@ -1165,3 +1165,271 @@ def estimate_scale(
...
+        predicted_rate:
+            Predicted mean count of the Negative Binomial distribution. Shape matches `observations`.
+        scale:
+            Dispersion parameter of the distribution.
+
+        Returns
+        -------
+        :
+            The residual deviance of the model. Shape matches `observations`.
```
billbrod:
```suggestion
            The residual deviance of the model. Shape matches ``observations``.
```

src/nemos/observation_models.py
```diff
@@ -1165,3 +1165,271 @@ def estimate_scale(
...
+        .. math::
+            \begin{aligned}
+                D(y_{tn}, \hat{y}_{tn}) &= 2 \left( \text{LL}\left(y_{tn} | y_{tn}\right) - \text{LL}\left(y_{tn}
+                  | \hat{y}_{tn}\right)\right) \\\
+                &= 2 \left[ y_{tn} \log\left(\frac{y_{tn}}{\hat{y}_{tn}}\right) + (1 - y_{tn}) \log\left(\frac{1
+                  - y_{tn}}{1 - \hat{y}_{tn}}\right) \right]
+            \end{aligned}
+
+        where :math:`y` is the observed data, :math:`\hat{y}` is the predicted data, and :math:`\text{LL}` is
+        the model log-likelihood. Lower values of deviance indicate a better fit.
```
billbrod: "Lower values of deviance indicate a better fit." should probably be in the main summary, not down here

src/nemos/observation_models.py
```diff
@@ -1165,3 +1165,271 @@ def estimate_scale(
...
+        y: jnp.ndarray,
+        predicted_rate: jnp.ndarray,
+        dof_resid: Union[float, jnp.ndarray],
+    ) -> Union[float, jnp.ndarray]:
+        r"""
+        Return the scale parameter of the distribution.
+
+        The `scale` parameter of the Negative Binomial distribution is set
+        at initialization and affect the likelihood landscape. This implies
+        that the `scale` parameter cannot be estimated post-hoc without
```
billbrod:
```suggestion
        The ``scale`` parameter of the Negative Binomial distribution is set
        at initialization and affect the likelihood landscape. This implies
        that the ``scale`` parameter cannot be estimated post-hoc without
```

tests/conftest.py
```diff
@@ -923,3 +923,132 @@ def population_bernoulliGLM_model_instantiation_pytree(
...
+    r = 1 / model.observation_model.scale
+    spikes = np.random.poisson(np.random.gamma(shape=r, size=rate.shape) * (r / rate))
+    return X, spikes, model, (w_true, b_true), rate
+
+
+@pytest.fixture
+def negativeBinomialGLM_model_instantiation_pytree(
+    negativeBinomialGLM_model_instantiation,
+):
+    """Set up a Negative Binomial GLM for testing purposes.
```
billbrod: this docstrings hould make it clear how it differs from `negativeBinomialGLM_model_instantiation` (I assume it's returning a pytree vs. an array?) since right now it's identical

tests/conftest.py
```diff
@@ -923,3 +923,132 @@ def population_bernoulliGLM_model_instantiation_pytree(
...
+    # make sure that at least one entry is non-zero
+    spikes = spikes.at[-1].set(1)
+    return X, spikes, model, (w_true, b_true), rate
+
+
+@pytest.fixture
+def population_negativeBinomialGLM_model_instantiation_pytree(
+    population_negativeBinomialGLM_model_instantiation,
+):
+    """Set up a population Negative Binomial GLM for testing purposes.
```
billbrod, line 1027: this docstrings hould make it clear how it differs from population_negativeBinomialGLM_model_instantiation (I assume it's returning a pytree vs. an array?) since right now it's identical

tests/test_glm.py
```diff
@@ -14,9 +15,26 @@
...
 
+GLM_COMMON_PARAMS_NAMES = {
+    "observation_model__inverse_link_function",
+    "observation_model",
+    "regularizer",
+    "regularizer_strength",
+    "solver_kwargs",
+    "solver_name",
+}
+OBSERVATION_MODEL_EXTRA_PARAMS_NAMES = {
```
billbrod, line 30: if you use `.get(obs_name, {})` where you call this, could remove the empty lines here (and will be easier to update going forward)

tests/test_glm.py
```diff
@@ -1797,49 +1850,66 @@ def test_get_params(self, request, glm_type, model_instantiation):
...
-                return [
-                    model.feature_mask,
-                    model.observation_model.inverse_link_function,
-                    model.observation_model,
-                    model.regularizer,
-                    model.regularizer_strength,
-                    model.solver_kwargs,
-                    model.solver_name,
-                ]
+                if isinstance(model.observation_model, NegativeBinomialObservations):
```
billbrod, line 1868: why isn't this able to use expected_keys defined above? (along with `getattr`)

tests/test_glm.py
```diff
@@ -1797,49 +1850,66 @@ def test_get_params(self, request, glm_type, model_instantiation):
...
-                    model.solver_kwargs,
-                    model.solver_name,
+            expected_keys = GLM_COMMON_PARAMS_NAMES.union(
+                OBSERVATION_MODEL_EXTRA_PARAMS_NAMES[
+                    model.observation_model.__class__.__name__
                 ]
+            )
 
-        _, _, model, _, _ = request.getfixturevalue(glm_type + model_instantiation)
+            def get_expected_values(model):
```
billbrod, line 1879: same point about expected_keys

tests/test_glm.py
```diff
@@ -2142,6 +2225,7 @@ def test_simulate_pynapple(
         )
         model.coef_ = true_params[0]
         model.intercept_ = true_params[1]
+        model.scale_ = model.observation_model.scale
```
billbrod: what's going on here?

tests/test_observation_models.py
```diff
@@ -1001,3 +1012,324 @@ def test_repr_out(self, link_func, link_func_name):
...
+            jax.nn.softplus,
+            np.exp,
+            lambda x: x,
+            sm.families.links.Log(),
+        ],
+    )
+    def test_initialization_link_is_jax(
+        self, link_function, negative_binomial_observations
+    ):
+        raise_exception = (
```
billbrod: why not do this in the parametrize?

tests/test_observation_models.py
```diff
@@ -1001,3 +1012,324 @@ def test_repr_out(self, link_func, link_func_name):
...
+            sts.norm.cdf,
+            np.exp,
+            lambda x: x,
+            sm.families.links.Log(),
+        ],
+    )
+    def test_initialization_link_is_jax_set_params(
+        self, link_function, negative_binomial_observations
+    ):
+        raise_exception = (
```
billbrod: same question about putting this in the parametrize

tests/test_observation_models.py
```diff
@@ -1001,3 +1012,324 @@ def test_repr_out(self, link_func, link_func_name):
...
+    ):
+        observation_model = negative_binomial_observations()
+        raise_exception = not callable(link_function)
+        if raise_exception:
+            with pytest.raises(
+                TypeError,
+                match="The `inverse_link_function` function must be a Callable",
+            ):
+                observation_model.set_params(inverse_link_function=link_function)
+        else:
```
billbrod: use `expectation = does_not_raise` and then get rid of this if/else

tests/test_observation_models.py
```diff
@@ -1001,3 +1012,324 @@ def test_repr_out(self, link_func, link_func_name):
...
+    ):
+        raise_exception = not isinstance(link_function(1.0), (jnp.ndarray, float))
+        observation_model = negative_binomial_observations()
+        if raise_exception:
+            with pytest.raises(
+                TypeError,
+                match="The `inverse_link_function` must handle scalar inputs correctly",
+            ):
+                observation_model.set_params(inverse_link_function=link_function)
+        else:
```
billbrod: same point about `does_not_raise()` and getting rid of if/else

tests/test_observation_models.py
```diff
@@ -1001,3 +1012,324 @@ def test_repr_out(self, link_func, link_func_name):
...
+    def test_scale_setter(
+        self, scale, expectation, negativeBinomialGLM_model_instantiation
+    ):
+        _, _, model, _, _ = negativeBinomialGLM_model_instantiation
+        with expectation:
+            model.observation_model.scale = scale
+
+    def test_scale_getter(self, negativeBinomialGLM_model_instantiation):
+        _, _, model, _, _ = negativeBinomialGLM_model_instantiation
+        assert model.observation_model.scale == 1  # or your default value
```
billbrod:
```suggestion
        assert model.observation_model.scale == 1  
```
doesn't add anything

billbrod (changes requested):
- This observation model doesn't show up in the [API docs](https://nemos--360.org.readthedocs.build/en/360/api_reference.html#the-nemos-observation-models-module), does some change need to be made there?
- Can we have an example of how to use this somewhere? Could be nice to show how the Poisson behaves with (simulated) over-dispersed data and show that NegativeBinomial does a better job. 
  - And also make it clear that scale is currently a hyperparameter that must be selected using cross-validation?

src/nemos/observation_models.py
BalzaniEdoardo: yes, it's a leftover, I prefer the sphinx math

src/nemos/observation_models.py
BalzaniEdoardo: I am adding intersphinx in a separate pr, so for now I'll monospace.

src/nemos/observation_models.py
BalzaniEdoardo: added! good call

tests/conftest.py
BalzaniEdoardo: Fixed the description and changed the type

tests/conftest.py
BalzaniEdoardo: yes, you are right. I changed the description

tests/test_glm.py
BalzaniEdoardo: How do I deal with nested params? i cannot use `getattr` unless I write a recursion, if i use `get_param`, it is circular.

tests/test_glm.py
BalzaniEdoardo: same as above

tests/test_glm.py
BalzaniEdoardo: not needed, removed.

src/nemos/observation_models.py
```diff
@@ -1165,3 +1165,292 @@ def estimate_scale(
...
+    def estimate_scale(
+        self,
+        y: jnp.ndarray,
+        predicted_rate: jnp.ndarray,
+        dof_resid: Union[float, jnp.ndarray],
+    ) -> Union[float, jnp.ndarray]:
+        r"""
+        Return the scale parameter of the distribution.
+
+        The ``scale`` parameter of the Negative Binomial distribution is set
```
sjvenditto, line 1433: I would be more explicit saying that the input parameters aren't being used in this function, and that it is parameterized for consistency with other observation models. (The reasoning is already there, but just to make it very obvious)

sjvenditto (approved):
Looks great!

src/nemos/observation_models.py
BalzaniEdoardo: I added a comment, let me know if it addresses your concern @sjvenditto

docs/quickstart.md
```diff
@@ -280,6 +283,24 @@ glm = nmo.glm.GLM(observation_model=nmo.observation_models.BernoulliObservations
 
 ```
 
+### **Counts with Supra-Poisson Variability**
+
+A [Negative Binomial observation model](nemos.observation_models.NegativeBinomialObservations) is appropriate for describing a counting process (such as neural spike counts) that exhibits supra-Poisson variability — that is, it is over-dispersed relative to a Poisson model (equivalently, the variance of the counts is larger than the mean).
+
+To set up a Negative Binomial GLM, provide this `observation_model` at initialization. You can regulate the degree of overdispersion by setting the `scale` parameter: the larger the `scale`, the more dispersed the Negative Binomial will be compared to a Poisson model.
```
billbrod: should mention that this model approaches the Poisson as scale goes to 0.

docs/quickstart.md
```diff
@@ -280,6 +283,24 @@ glm = nmo.glm.GLM(observation_model=nmo.observation_models.BernoulliObservations
...
+
+```{code-cell} ipython3
+
+import nemos as nmo
+
+# set up a Negative Binomial GLM for modeling counts (with scale = 1)
+# note: you should tune the scale parameter (for example, through cross-validation)
```
billbrod: this shouldn't be in a comment, but above or in an admonition (and link to something about cross-validation)

src/nemos/observation_models.py
```diff
@@ -1165,3 +1165,296 @@ def estimate_scale(
...
+
+
+class NegativeBinomialObservations(Observations):
+    r"""
+    A Negative Binomial model for overdispersed count data using mean-dispersion parameterization.
+
+    This model represents a Negative Binomial distribution [3]_ commonly used to model overdispersed
+    count data [4]_ [5]_ (i.e., data where the variance exceeds the mean), which cannot be captured by a
+    standard Poisson model. The distribution is parameterized by the predicted mean rate
+    (:math:`\mu`) and a fixed dispersion parameter (:math:`\phi`) or "scale" of the model.
```
billbrod:
```suggestion
    (:math:`\mu`) and a fixed dispersion parameter (:math:`\phi`) or `scale` of the model.
```

since this is the parameter they're setting

src/nemos/observation_models.py
```diff
@@ -1165,3 +1165,296 @@ def estimate_scale(
...
+
+    **Important:** the scale parameter must be estimated from the data for accurately capturing the
+    dispersion. In the context of NeMoS GLM, estimation can be achieved by cross-validating the
+    ``scale`` parameter. One may use scikit-learn
+    `GridSearchCV <https://scikit-learn.org/stable/modules/generated/sklearn.model_selection.GridSearchCV.html>`_
```
billbrod: is this what we should link, rather than our example of how to do CV?

tests/test_glm.py
billbrod: ah okay, I missed the difficulty of the nested parameters. 

Still, to make this simpler: all models have the values in the else block, and then just have the `if` statement add the `scale`. And could have that check whether observation model has a scale or not (rather than specifically whether it's Negative Binomial)?

tests/test_glm.py
billbrod: and same point about simplification

billbrod (changes requested):
Just some small comments. If you don't want to include a page in the docs about how to use this observation model in this PR, can we create an issue for that?

src/nemos/observation_models.py
```diff
@@ -1165,3 +1165,296 @@ def estimate_scale(
...
+
+    This model represents a Negative Binomial distribution [3]_ commonly used to model overdispersed
+    count data [4]_ [5]_ (i.e., data where the variance exceeds the mean), which cannot be captured by a
+    standard Poisson model. The distribution is parameterized by the predicted mean rate
+    (:math:`\mu`) and a fixed dispersion parameter (:math:`\phi`) or `scale` of the model.
+
+    **Important:** the scale parameter must be estimated from the data for accurately capturing the
+    dispersion. In the context of NeMoS GLM, estimation can be achieved by cross-validating the
+    ``scale`` parameter. One may use scikit-learn
+    `GridSearchCV <../../how_to_guide/plot_06_sklearn_pipeline_cv_demo.html>`_
```
billbrod: This link works, but would be more future proof to use a proper sphinx link (which I think you can do from docstring), by linking to the explicit anchor `sklearn-how-to`

billbrod (approved):
Some very small comments and then this looks good! I don't need to review it again.
- I noticed that the `ObservationModel` base class "See also" section does not include `NegativeBinomial`. Should do one of: add it, find a way to automatically generate that section, or remove it entirely.

src/nemos/observation_models.py
BalzaniEdoardo: I see if I can figure that out

BalzaniEdoardo: done

PR #368, opened 2025-07-09
Sklearn main dep

BalzaniEdoardo:
# Inheriting `BaseEstimator` from scikit-learn

Recently, scikit-learn added a new [API for tagging](https://scikit-learn.org/stable/developers/develop.html#estimator-tags) the estimators. This tags are data-classes containing meta-information about the estimators and are used by scikit-learn for testing and validation purposes.

Re-implementing the dataclasses and keeping them up-to-date with sklearn seems too brittle and therefore I think it is worthwhile inheriting `BaseEstimator` directly.

## Changing `Basis` inheritance order

As part of this PR, I changed the order of inheritance of `Basis` because I need to prioritize our own overwritten methods over that of `BaseEstimator`. This methods include:
- `get_params` and `set_params`: basis get and set params using labels instead of the nested attribute structure. 
- `__sklearn_clone__`: we are re-appending some attributes that are set after initialization (like the input shape) and would not be correctly set with `self.__class__(self.get_params())`, which is what the default clone does.

### DIFF
--- a/docs/developers_notes/02-base_regressor.md
+++ b/docs/developers_notes/02-base_regressor.md
@@ -59,7 +59,7 @@ When devising a new model subclass based on the `BaseRegressor` abstract class, 
 - **Must** realize the abstract methods, see above.
 - **Should not** overwrite the `get_params` and `set_params` methods, inherited from `Base`.
 - **May** introduce auxiliary methods for added utility.
-- **May** add re-implement the `__sklearn_tags__` method adding metadata that is relevant to the specific estimator implemented. See the [`scikit-learn` Tags](https://scikit-learn.org/stable/modules/generated/sklearn.utils.Tags.html#sklearn.utils.Tags) for the available tagging options.
+- **May** re-implement the `__sklearn_tags__` method to add metadata that is relevant to the specific estimator implemented. See the [`scikit-learn` documentation](https://scikit-learn.org/stable/modules/generated/sklearn.utils.Tags.html#sklearn.utils.Tags) for the available tagging options.
 
 :::{admonition} Tags
 :class: note


billbrod (changes requested):
Just minor comments, largely to make sure we know how this affects future models we'll write.
- Can we add something to the developer notes to point out that new models will need to define `__sklearn_tags__`? And any guidance we can provide there / link to sklearn docs
- That's the only change, right? They should still inherit only our classes?
- Can you write an explanation in this PR's opening comment about why you changed the inheritance order for the basis objects?

BalzaniEdoardo:
> Just minor comments, largely to make sure we know how this affects future models we'll write.
> 
>     * Can we add something to the developer notes to point out that new models will need to define `__sklearn_tags__`? And any guidance we can provide there / link to sklearn docs

Let me add to the developer note some info. I think we do not need the tags for anything we use (including pipelining and cv), but I added some tags that made sense for regression models. Sklearn has a [page](https://scikit-learn.org/stable/developers/develop.html) for developers of estimators, but there is only a small paragraph on Tags. From that I deduced that it is used to automate checks and internal testing of sklearn.

>     * That's the only change, right? They should still inherit only our classes?

Yes. With the change we get a new public method, like `get_metadata_routing`, that returns a dictionary with the kwargs that estimator methods accept. 

```
In [5]: mdl.get_metadata_routing()
Out[5]: {'fit': {'init_params': None}, 'score': {'score_type': None, 'aggregate_sample_scores': None}}
```

The routing can be used for stuff like changing the score type in a cross-validation etc. but is an advanced usage I don't feel like we should advertise (and must be enabled explicitly according to sklearn docs, so it is quite advanced and not as stable as the rest maybe?) see [here](https://scikit-learn.org/stable/metadata_routing.html). 

>     * Can you write an explanation in this PR's opening comment about why you changed the inheritance order for the basis objects?

Added to the description of the PR above.

docs/developers_notes/02-base_regressor.md
```diff
@@ -59,7 +59,16 @@ When devising a new model subclass based on the `BaseRegressor` abstract class,
 - **Must** realize the abstract methods, see above.
 - **Should not** overwrite the `get_params` and `set_params` methods, inherited from `Base`.
 - **May** introduce auxiliary methods for added utility.
+- **May** add re-implement the `__sklearn_tags__` method adding metadata that is relevant to the specific estimator implemented. See the [`scikit-learn` Tags](https://scikit-learn.org/stable/modules/generated/sklearn.utils.Tags.html#sklearn.utils.Tags) for the available tagging options.
```
billbrod:
```suggestion
- **May** re-implement the `__sklearn_tags__` method to add metadata that is relevant to the specific estimator implemented. See the [`scikit-learn` documentation](https://scikit-learn.org/stable/modules/generated/sklearn.utils.Tags.html#sklearn.utils.Tags) for the available tagging options.
```
I think I'd add [this page](https://scikit-learn.org/stable/developers/develop.html#estimator-tags) here as well (not just in the admonition).

billbrod: maybe worth giving an example here? right now, it's not clear to me when I'd write additional tags.

src/nemos/base_class.py
```diff
@@ -26,6 +26,22 @@ class Base(BaseEstimator):
     Additionally, it has methods for selecting target devices and sending arrays to them.
     """
 
+    def get_params(self, deep=True) -> dict:
+        """
+        From scikit-learn, get parameters by inspecting init.
+
+        Parameters
+        ----------
+        deep
```
billbrod: missing explanation

PR #372, opened 2025-07-17
Fix param naming

BalzaniEdoardo:
Fixed parameter inconsistencies and added checks to tox.


.github/workflows/ci.yml
```diff
@@ -76,10 +76,10 @@ jobs:
         uses: actions/checkout@v4
 
       - name: Make .sh executable
-        run: chmod +x bash_scripts/prevent_absolute_links_to_docs.sh
+        run: chmod +x scripts/prevent_absolute_links_to_docs.sh
```
billbrod: don't know why I didn't see this before, but if you make the script executable locally and then commit it, you don't have to do this.

.github/workflows/ci.yml
```diff
@@ -91,10 +91,10 @@ jobs:
         run: npm install -g markdown-link-check
 
       - name: Make .sh executable
-        run: chmod +x bash_scripts/check_markdown_links.sh
+        run: chmod +x scripts/check_markdown_links.sh
```
billbrod: same point about chmod

.github/workflows/ci.yml
```diff
@@ -106,10 +106,10 @@ jobs:
...
```
billbrod: same point about chmod

scripts/check_parameter_naming.py
```diff
@@ -0,0 +1,155 @@
...
+    Parameters
+    ----------
+    package : module
+        The root package to analyze (e.g., `pynapple`).
+    root_name : str, optional
+        The dotted name of the root package. If not provided, it is inferred from
+        `package.__name__`.
+    similarity_cutoff : float, optional
+        Similarity threshold between 0 and 1 used to group parameters based on
+        lexical similarity (default is 0.8).
```
billbrod:
```suggestion
        lexical similarity.
```

scripts/check_parameter_naming.py
```diff
@@ -0,0 +1,155 @@
+import difflib
+import inspect
+import logging
+import sys
+import types
+import itertools
+
+import nemos as nmo
+
+VALID_PAIRS = [
```
billbrod, line 23: add a comment explaining what this is

scripts/check_parameter_naming.py
```diff
@@ -0,0 +1,155 @@
...
+
+
+def collect_similar_parameter_names(package, root_name=None, similarity_cutoff=0.8):
+    """
+    Recursively collect and group similar parameter names from functions and methods.
+
+    This function traverses the given package and its submodules, extracting parameter
+    names from all user-defined functions and methods. Parameter names that are
+    lexically similar (based on `difflib.get_close_matches`) are grouped together.
+    This can be used to detect inconsistent naming conventions across a codebase.
```
billbrod: describe how VALID_PAIRS is used. maybe make that an optional argument that defaults to the global here?

scripts/check_parameter_naming.py
```diff
@@ -0,0 +1,155 @@
...
+import sys
+import types
+import itertools
+
+import nemos as nmo
+
+VALID_PAIRS = [
+    {"conv_time_series", "time_series"},
+    {"inhib_a", "inhib_b"},
+    {"excit_a", "excit_b"},
```
billbrod, line 26: why is this formatted as a `list[str]` instead of a `dict[str, list[str]]`? that seems more intuitive to me (assuming I understand what this is doing)

scripts/check_parameter_naming.py
```diff
@@ -0,0 +1,155 @@
...
+                process_function(obj, path_prefix)
+
+        elif inspect.isclass(obj):
+            if getattr(obj, "__module__", "").startswith(root_name):
+                for name, member in inspect.getmembers(obj):
+                    if name.startswith("_"):
+                        continue
+                    walk(member, f"{path_prefix}.{name}")
+
+        elif isinstance(obj, types.ModuleType):
```
billbrod: what's ModuleType?

scripts/check_parameter_naming.py
```diff
@@ -0,0 +1,155 @@
...
+            if not getattr(obj, "__name__", "").startswith(root_name):
+                return  # external module, skip
+            for name, member in inspect.getmembers(obj):
+                if name.startswith("_"):
+                    continue
+                walk(member, f"{path_prefix}.{name}")
+
+    walk(package, package.__name__)
```
billbrod: why isn't this `root_name`?

scripts/check_parameter_naming.py
```diff
@@ -0,0 +1,155 @@
...
+    parser.add_argument(
+        "--threshold",
+        "-t",
+        type=float,
+        default=0.8,
+        help="Similarity threshold (between 0 and 1) for grouping parameter names (default: 0.9)",
+    )
+    args = parser.parse_args()
+
+    logger = logging.getLogger("check_parameter_naming")
```
billbrod, line 258: out of curiosity, why use logger instead of just print?

src/nemos/tree_utils.py
```diff
@@ -95,13 +95,13 @@ def get_valid_multitree(*tree: Any) -> jnp.ndarray:
...
     """
-    return reduce(jnp.logical_and, map(_get_valid_tree, tree))
+    return reduce(jnp.logical_and, map(_get_valid_tree, pytree))
 
 
 def pytree_map_and_reduce(
     map_fn: Callable,
     reduce_fn: Callable,
-    *pytrees: Any,
+    *pytree: Any,
```
billbrod: is this correct? I feel like it being plural is meant to mark that it can be more than one

scripts/check_parameter_naming.py
```diff
@@ -0,0 +1,155 @@
+import difflib
+import inspect
+import logging
+import sys
+import types
+import itertools
+
+import nemos as nmo
```
billbrod: if you want this script to be more general (not sure that's necessary), then shouldn't import this here, but let the user specify the module as an arg, and only then import it.

billbrod (changes requested):
This looks very useful! I don't understand the logic of the core code though, can you explain that a bit more? maybe some comments in the file. I also tried to trigger this manually by opening up `basis/basis.py`, and on line 565, changing `n_basis_funcs` to `n_basis_func`, and that didn't cause the script to fail -- am I misunderstanding what this should be catching?

- Can also add this to pre-commit-config, if you'd like. I can do that, if you'd like.

scripts/check_parameter_naming.py
BalzaniEdoardo: I added a long description

scripts/check_parameter_naming.py
BalzaniEdoardo: check if obj is a module, for example

isinstance(np, types.ModuleType)
Out[4]: True

scripts/check_parameter_naming.py
BalzaniEdoardo: it should be

scripts/check_parameter_naming.py
BalzaniEdoardo: In the github CI the prints are suppressed, logs are shown

scripts/check_parameter_naming.py
BalzaniEdoardo: good idea

src/nemos/tree_utils.py
BalzaniEdoardo: I'll revert that and add the pair to VALID_PAIRS

scripts/check_parameter_naming.py
BalzaniEdoardo: It is convenient for the checking if a pair is actually a violation (i.e. two strings are similar but should not be). All I need to do is `{matched_name, parameter_name} in valid_pairs`.


Note that I am assuming the key matching is transitive (and symmetric), i.e. key_a matches key_b and key_c, then key_b matches key_c. I am not sure this is true, and most likely it is not (because one can keep adding a letter to a string and eventually end up with dissimilar outputs "hello" -> "helloa" -> "helloab" ... ) but that makes the code simpler.

BalzaniEdoardo: I could make it more general by checking over all the unique keys in the matching dict

BalzaniEdoardo: actually, I'll do that

billbrod: I missed the fact that set equality doesn't care about ordering, that makes a lot of sense.

scripts/check_parameter_naming.py
```diff
@@ -0,0 +1,240 @@
...
+    package : module
+        The root package to analyze (e.g., pynapple).
+    root_name : str, optional
+        The dotted name of the root package. If not provided, it is inferred from
+        package.__name__.
+    similarity_cutoff : float, optional
+        Similarity threshold between 0 and 1 used to group parameters based on
+        lexical similarity.
+    valid_pairs :
+        Pairs of similar strings that are allowed as distinct parameter names.
```
billbrod: describe what happens when `valid_pairs is None`

scripts/check_parameter_naming.py
```diff
@@ -0,0 +1,240 @@
...
+    names from all user-defined functions and methods. Parameter names that are
+    lexically similar (based on difflib.get_close_matches) are grouped together.
+    This can be used to detect inconsistent naming conventions across a codebase.
+
+    Parameters
+    ----------
+    package : module
+        The root package to analyze (e.g., pynapple).
+    root_name : str, optional
+        The dotted name of the root package. If not provided, it is inferred from
```
billbrod:
```suggestion
        The dotted name of the root package. If ``None``, it is inferred from
```

scripts/check_parameter_naming.py
```diff
@@ -0,0 +1,240 @@
...
+            is_in_category = any(match in v["unique_names"] for match in matches)
+            if is_in_category:
+                v["info"].append((current_parameter, current_path))
+                v["unique_names"].add(current_parameter)
+
+
+def collect_similar_parameter_names(
+    package,
+    root_name: Optional[str] = None,
+    similarity_cutoff=0.8,
```
billbrod: type annotation?

scripts/check_parameter_naming.py
```diff
@@ -0,0 +1,240 @@
...
+    if valid_pairs is None:
+        valid_pairs = VALID_PAIRS
+
+    results = {}
+    visited_ids = set()
+    # set of all unique parameter names
+    unique_param_names = set()
+
+    def process_function(func, path):
+        if "jaxopt." in path:
```
billbrod: in path? or `path.startswith`? 

I'm assuming we would never do something as terrible as including a `nemos.jaxopt` module (that has nothing to do with `third_party/jaxopt`) but you never know...

scripts/check_parameter_naming.py
billbrod: I've never used that before, that's handy

scripts/check_parameter_naming.py
```diff
@@ -0,0 +1,240 @@
...
+            return
+        try:
+            sig = inspect.signature(func)
+            param_names = list(sig.parameters)
+            for par in param_names:
+                if par in results:
+                    results[par]["unique_names"].add(par)
+                    results[par]["info"].append((par, path))
+                    continue  # exact name already exists store
```
billbrod:
```suggestion
                    # exact name already exists store
                    results[par]["unique_names"].add(par)
                    results[par]["info"].append((par, path))
                    continue
```
the comment is about this whole if branch, right?

scripts/check_parameter_naming.py
```diff
@@ -0,0 +1,240 @@
...
+    {"func", "funcs"},
+    {"l_smooth", "l_smooth_max"},
+    {"attr_name", "var_name"},
+    # jaxopt solvers use fun as kwarg => our SVRG must use fun too
+    # it is common for decorators to have a "func" argument, therefore I'll allow both
+    {"func", "fun"},
+]
+
+
+def handle_matches(
```
billbrod, line 74: docstring?

scripts/check_parameter_naming.py
```diff
@@ -0,0 +1,240 @@
...
+                    results[par]["unique_names"].add(par)
+                    results[par]["info"].append((par, path))
+                    continue  # exact name already exists store
+
+                # match with all unique parameters
+                match = difflib.get_close_matches(
+                    par, unique_param_names, n=100, cutoff=similarity_cutoff
+                )
+                # add to result dictionary
+                handle_matches(par, path, match, results, valid_pairs)
```
billbrod: how is results updated here? shouldn't this return results as well?

scripts/check_parameter_naming.py
```diff
@@ -0,0 +1,240 @@
...
+
+
+def handle_matches(
+    current_parameter: str,
+    current_path: str,
+    matches: List[str],
+    results: Dict,
+    valid_pairs: List[set[str]],
+):
+    # a parameter name is valid if no matches or all matches in valid pairs
```
billbrod, line 115: I'm confused as to how the following line checks this -- it looks like it only checks the first part?

scripts/check_parameter_naming.py
```diff
@@ -0,0 +1,240 @@
...
+    Returns
+    -------
+    dict
+        A dictionary mapping canonical parameter names to a list of tuples.
+        Each tuple contains:
+            - The actual parameter name
+            - The fully qualified function or method path where it appears
+
+        Example
+        -------
```
billbrod: this not an example?

billbrod: and incorrectly indented

scripts/check_parameter_naming.py
```diff
@@ -0,0 +1,240 @@
...
+                # match with all unique parameters
+                match = difflib.get_close_matches(
+                    par, unique_param_names, n=100, cutoff=similarity_cutoff
+                )
+                # add to result dictionary
+                handle_matches(par, path, match, results, valid_pairs)
+                # add to unique params
+                unique_param_names.add(par)
+        except Exception:
+            pass  # some built-ins or extension modules may not support signature()
```
billbrod:
```suggestion
            # some built-ins or extension modules may not support signature()
            pass
```

scripts/check_parameter_naming.py
BalzaniEdoardo: jaxopt is in nemos.third_party.jaxopt, so I used third_party instead

PR #382, opened 2025-07-24
Log likelihood

BalzaniEdoardo:
Add a log-likelihood compute function. This method is implemented in the base `Observations` class, simply as `exp(log-likelihood)`, however, it can be re-implemented wherever there is a smarter way to compute the likelihood, as in the `BernoulliObservations` case.

In the process of adding the test of the method I refactored the observation model tests.


sjvenditto (approved):
⭐

PR #386, opened 2025-08-11
Fourier nd

BalzaniEdoardo:
This PR adds a class for N-dimensional `FourierEval` bases. This basis is composed by pairs of cosine and sine at different frequencies. The output of `compute_features` is  a real valued design matrix, but columns corresponding to cosine and sine pairs represent real and imaginary part of a complex number. 

For this reason, naive multiplication with other real bases (i.e. BSpline, Raised Cosines etc) is disabled, as it would treat all columns as real. 

The `FourierEval` basis can be instantiated as multidimensional, by providing an `ndim` parameter. 

Frequency pairs can be selected by specifying an optional `frequency_mask` which can be: 

- A boolean array of shape `(n_1, n_2, ..., n_m)` where `m` is the dimensionality of the basis (`ndim`) and `n_k` is the number of frequencies for the k-th dimension of the Fourier basis.
- A  callable: `def mask_func(f_1: float, ..., f_m: float): -> bool`, which takes as input `m` frequency values and returns a boolean.

### DIFF
--- a/src/nemos/basis/_fourier_basis.py
+++ b/src/nemos/basis/_fourier_basis.py
@@ -37,6 +37,46 @@ class FourierBasis(AtomicBasisMixin, Basis):
         frequency_mask: jnp.ndarray | None = None,
         label: Optional[str] = None,
     ) -> None:
+        """
+        N-dimensional Fourier basis for feature expansion.
+
+        This class generates a set of sine and cosine basis functions defined over
+        an ``n``-dimensional input space. The basis functions are constructed from
+        a Cartesian product of frequencies specified for each input dimension.
+        Each selected frequency combination contributes two basis functions
+        (cosine and sine), except for the all-zero frequency (DC component),
+        which contributes only a cosine term.
+
+        The class supports flexible frequency specification (integers, ranges, or
+        lists per dimension) and optional masking to include or exclude specific
+        frequency combinations.
+
+        Parameters
+        ----------
+        frequencies : int | tuple[int, int] | list[int] | list[tuple[int, int]]
+            Frequencies to use along each input dimension.
+
+            - ``int``: creates a range ``[0, 1, ..., n-1]`` for **all dimensions**.
+            - ``tuple[int, int]``: creates a range ``[start, ..., stop-1]`` for **all dimensions**.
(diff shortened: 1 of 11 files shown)


BalzaniEdoardo:
> ...
> ```
> (MSplineEval(n_basis_funcs=10, order=4, bounds=(0, 100))
> * MSplineEval(n_basis_funcs=10, order=4, bounds=(0, 100))).to_transformer()
> ```
> 
> I dug a little deeper and found that `bas.input_shape` is equal to `[(), ()]` in for `FourierEval`. Whereas for the `MSplineEval` it is equal to `()`.

Great catch, I was not testing the ndim >1 fourier basis in test transformer basis! I fixed the issue

src/nemos/basis/_fourier_basis.py
```diff
@@ -0,0 +1,708 @@
...
+            raise ValueError(
+                f"ndim must be a non-negative integer. {ndim!r} provided instead."
+            )
+        return int(ndim)
+
+    @property
+    def n_basis_funcs(self) -> int | None:
+        return 2 * self._freq_combinations.shape[-1] - self._has_zero_phase
+
+    @support_pynapple(conv_type="numpy")
```
sjvenditto, line 604: This decorator appears to be redundant, since it is added within `check_transform_input`

tests/test_basis.py
```diff
@@ -81,13 +124,59 @@ def compare_basis(b1, b2):
...
+        )
         assert d1 == d2
 
 
+def create_atomic_basis_pairs(full_list):
+    """
+    Define basis pairs preventing combinatorial explosion.
+
+    Returns a list of basis class tuple paring all bases with
+    a conv and eval basis. Pairing oredr matters.
```
sjvenditto:
```suggestion
    a conv and eval basis. Pairing order matters.
```

src/nemos/basis/_fourier_basis.py
BalzaniEdoardo: this can be improved, in the sense that we are calling the conversion twice but without it the casting won't work. (try removing and run `pytest tests/test_basis.py::TestAdditiveBasis::test_compute_features_input`) 

Improving the logic would require editing all the basis, so I would defer this to the PR on jax compatibility

PR #389, opened 2025-08-22
Enforce Consistent Input Shapes for MultiplicativeBasis and CustomBasis

BalzaniEdoardo:
# Enforce Consistent Input Shapes for MultiplicativeBasis and CustomBasis

## Problem
Currently, `MultiplicativeBasis` and `CustomBasis` accept inputs with different shapes across dimensions, leading to potentially confusing behavior where all possible combinations of input features are computed. This creates unexpectedly large output matrices and doesn't align with the typical use case where these basis types model single variables with multiple dimensions.

## Solution
This PR enforces that all inputs to n-dimensional `MultiplicativeBasis` and `CustomBasis` must have the same shape, and changes the vectorization logic to pair inputs element-wise rather than computing all combinations.

## Behavioral Change

**Before**: All combinations computed
```python
# Different shapes allowed
posx = np.random.randn(10, 3)  # 3 x-coordinates  
posy = np.random.randn(10, 4)  # 4 y-coordinates
out = basis.compute_features(posx, posy). # Output: (10, 24) - all 3×4 combinations computed
```

**After**: Element-wise pairing
```python
# Same shapes required
posx = np.random.randn(10, 4)  # 4 x-coordinates
posy = np.random.randn(10, 4)  # 4 y-coordinates  
out = basis.compute_features(posx, posy). # Output: (10, 8) - 4 pairs computed element-wise
```

This change makes `MultiplicativeBasis` and `CustomBasis` behave more intuitively for modeling multi-dimensional variables (e.g., 2D position, 3D velocity).

### Input Shape Validation
- **At initialization**: When combining basis with different input shapes, the shape is reset to `None` with a warning
- **At `set_input_shape()`**: Runtime check ensures all dimensions have consistent shapes
- **At `compute_features()`**: Runtime validation of actual input shapes

### API Changes
- **Restriction**: Cannot set input shape on child basis directly (e.g., `additive_basis.basis1.set_input_shape()` is no longer allowed)
- **Rationale**: Simplified shape management - set shapes at the root level only

### Internal Changes
- Updated `_input_shape_` attribute to return consistent shape representation (single shape for all dimensions instead of per-dimension shapes)
- `split_by_feature()` and `n_output_features()` work correctly without modification due to the `_input_shape_` change
- Enhanced `_compute_features()` vectorization logic

## Breaking Changes
- `MultiplicativeBasis` and `CustomBasis` now require consistent input shapes
- Child basis shape setting is prohibited
- Output dimensions will be smaller for cases that previously used different input shapes

## Migration Guide
If you were relying on the previous all-combinations behavior:
1. Ensure all inputs to `MultiplicativeBasis`/`CustomBasis` have the same shape
2. If you need all combinations, consider using multiple `AdditiveBasis` components instead
3. Set input shapes at the root basis level, not on individual components

### DIFF
--- a/src/nemos/basis/_basis.py
+++ b/src/nemos/basis/_basis.py
@@ -26,7 +26,6 @@ from ._check_basis import (
 )
 from ._composition_utils import (
     add_docstring,
-    get_input_shape,
     infer_input_dimensionality,
     is_basis_like,
     multiply_basis_by_integer,
@@ -932,6 +931,19 @@ class MultiplicativeBasis(CompositeBasisMixin, Basis):
         CompositeBasisMixin.__init__(self, basis1, basis2, label=label)
         Basis.__init__(self)
 
+        # reset input shape if multiplied bases have different input
+        # shapes
+        input_shape = self.input_shape
+        unique_shape = {s for s in input_shape}
+        # at initialization, basis2 can be None
+        if len(unique_shape) != 1 and self.basis2:
+            warnings.warn(
+                category=UserWarning,
+                message="Multiple different input shapes detected. "
+                "Resetting input shape to default (None).",
+            )
+            set_input_shape(self, None)
+
     def _generate_label(self) -> str:
         return "(" + self.basis1.label + " * " + self.basis2.label + ")"
 
@@ -1203,18 +1215,3 @@ class MultiplicativeBasis(CompositeBasisMixin, Basis):
            inputs are available.
         """
         return self.input_shape[:1]
-
-    @property
-    def input_shape(self):
-        input_shape = get_input_shape(self)
-        unique_shape = {s for s in input_shape}
-        # at initialization, basis2 can be None
-        if len(unique_shape) != 1 and self.basis2:
-            warnings.warn(
-                category=UserWarning,
-                message="Multiple different input shapes detected. "
-                "Resetting input shape to default (None).",
-            )
-            set_input_shape(self, None)
-            input_shape = [None] * self._n_input_dimensionality
-        return input_shape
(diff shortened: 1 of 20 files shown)


src/nemos/basis/_basis.py
```diff
@@ -979,12 +1040,25 @@ def _compute_features(
...
         X = kron(
-            comp_feature_1(*xi[: self.basis1._n_input_dimensionality]),
-            comp_feature_2(*xi[self.basis1._n_input_dimensionality :]),
+            x1.reshape((n_vec_inputs * n_samples, -1)),
+            x2.reshape((n_vec_inputs * n_samples, -1)),
             transpose=False,
         )
-        return X
+        shape = self._input_shape_[0]
+        return X.reshape((n_samples, *shape, -1)).reshape((n_samples, -1))
```
sjvenditto:
```suggestion
        return X.reshape((n_samples, -1))
```

couldn't you do a single reshape?

BalzaniEdoardo: yes, probably! I'll check that the result is equivalent but it should be.

BalzaniEdoardo:
@sjvenditto I agree with your point. The vectorization makes sense for convolutional and 1D eval bases where you're applying the same basis across neurons or multiple input streams.
For multiplicative bases, I chose to maintain the current vectorization approach for two design reasons:

- **API consistency** - all basis types follow this same pattern, so users get expected behavior
- **Feature concatenation**  - treating extra dimensions as samples would change axis 0 shape, breaking AdditiveBasis concatenation along axis 1

This is out of the scope of this PR, but we could handle trials smartly, by letting the user specify more than one sample axis, for example:

```python

# axis 0 = trial, axis = 1 samples x trial 
basis = nmo.basis.MyBasis(*kwargs, sample_axis=(0, 1))
# The basis processes inputs of shape 
# (n_trials, n_samples_x_trial, *other_dimensions))


```

The output design matrix will be of shape `(n_trials * n_samples_x_trial, prod(other_dimensions) * n_basis_funcs)`, and the user can rely on `split_by_features` to reshape a predicted firing rate from a GLM  into  `(n_trials, n_samples_x_trial, n_neurons)`.

If you like this idea we should open an issue about it!

PR #394, opened 2025-09-17
Move nonlinearity to obsmodel

BalzaniEdoardo:
Move non-linearity computation to observation model.


BalzaniEdoardo:
We decided to go in the opposite direction: move the link function away completely, since it is a part of the GLM not of the observation model itself.

PR #396, opened 2025-09-22
Link to glm

BalzaniEdoardo:
## Move Inverse Link Function to GLM
In this PR we are moving the link function from the observation model to the GLM. The inverse link is part of the GLM model and not of the observation, and this was reflected in the code (aka the inverse link was never used in any of the observation model methods, except for the method that checked its validity).

Moving it to the GLM makes the instantiation of the model more transparent, all the key components of the model can be specified upfront and the repr string will be more informative. For example, setting up a NegativeBinomial model with softplus inverse link would look like this

```python
>>> import nemos as nmo
>>> import jax
>>> model = nmo.glm.GLM(
...     observation_model="NegativeBinomial", 
...     inverse_link_function=jax.nn.softplus
...    )
>>> model
GLM(
    observation_model=NegativeBinomialObservations(scale=1.0),
    inverse_link_function=softplus,
    regularizer=UnRegularized(),
    solver_name='GradientDescent'
)
```

While originally, one would have to first instantiate the observation model with the softplus link, and pass that class as a parameter.

## New in this PR

- Moved the `glm.py` script into a module folder.
- Moved the link funciton validation logic to the glm module.
- Moved the property with setter and getter for the link function to the GLM class.
- Add a read-only property `default_inverse_link_function` to observation model that returns the default link.
- Allowed for inverse_link_function to be optional, if None, it will be set based on the observation model default.
- Moved the tests regarding the link function from observation model to glm.
- Moved the model initialization script to the glm module and renamed to `initialize_parameters.py`


src/nemos/glm/glm.py
```diff
@@ -57,6 +61,21 @@ class GLM(BaseRegressor):
...
+    Below is a table of the default inverse link function for the availabe observation model.
+
+    +---------------------+---------------------------------+
+    | Observation Model   | Default Inverse Link Function   |
+    +=====================+=================================+
+    | Poisson             | :math:`e^x`                     |
+    +---------------------+---------------------------------+
+    | Gamma               | :math:`1/x`                     |
+    +---------------------+---------------------------------+
+    | Binomial            | :math:`1 / (1 + e^{-x})`        |
```
sjvenditto: Right now the class name is "Bernoulli". Do you think we should change it to binomial in a future PR? I think we mix these two up a lot, and we should settle on one

src/nemos/glm/inverse_link_function_utils.py
```diff
@@ -0,0 +1,103 @@
...
+    "jax.scipy.special.expit": jax.scipy.special.expit,
+    "jax._src.scipy.special.expit": jax.scipy.special.expit,
+    "jax.lax.logistic": jax.lax.logistic,
+    "jax._src.lax.lax.logistic": jax.lax.logistic,
+    "jax.scipy.stats.norm.cdf": jax.scipy.stats.norm.cdf,
+    "jax._src.scipy.stats.norm.cdf": jax.scipy.stats.norm.cdf,
+}
+
+
+def link_function_from_string(link_name: str):
```
sjvenditto, line 31: if we allow setting the link function from a string, should we also have simplified names in the list? like specifying `"exp"` which also defaults to `jnp.exp`

sjvenditto (approved):
I really like the move to the glm class, it makes initializing this parameter much simpler! My only suggestion is to allow for simple string names of functions usable when initializing link functions

src/nemos/glm/glm.py
BalzaniEdoardo: good catch!

Issue #398, opened 2025-10-03
Parallel E-step GLM-HMM


BalzaniEdoardo:
**Is your feature request related to a problem? Please describe.**
The Forward-backward algorithm can be parallelized with an [associative scan](https://docs.jax.dev/en/latest/_autosummary/jax.lax.associative_scan.html), since it's just a bunch of matmuls. However, this is true within a session, multiple sessions require a "message resetting" that breaks this symmetry.  We should decide if it is worth spending time to parallelize the computation with this in mind. 

**Describe the solution you'd like**
Possible solution, implement multiple versions of the algorithm:
1 - `associative_scan` version of the forward and backward passes
2 - keep the current version that uses a regular scan plus resetting. 

At fit time, check if: 
a. There is a **single session** -> call the `associative_scan` algorithm.
b. There are **multiple session of the same duration**: -> vmap the `associative_scan` and call that algorithm (vmapping over sessions)
c. There are **multiple session with different durations** -> call the current scan-based algorithm.

PR #399, opened 2025-10-03
Glm hmm e step

BalzaniEdoardo:
# GLM-HMM E-step
Implement the forward-backward algorithm of the GLM-HMM with trial resetting.

## Included in the PR
- Main scan based algorithms for the forward and backward passes.
- Joint forward-backward step which outputs other auxiliary quantities needed for the M-step and the evaluation of the model likelihood.
- Regression test for the E-step parameters.

## Notes
The E-step can be parallelized, using `jax.lax.associative_scan` since it consists of a bunch of matmul that can be applied in whatever order given the associative property of the product; however the phase resetting will need a bit of thought because it breaks this symmetry, see issue #398 for additional details and thoughts.

### DIFF
--- a/src/nemos/observation_models.py
+++ b/src/nemos/observation_models.py
@@ -1030,7 +1030,7 @@ class BernoulliObservations(Observations):
         Returns
         -------
         :
-            The Bernoulli negative log-likehood. Shape (1,).
+            The Bernoulli negative log-likelihood. Shape (1,).
 
         Notes
         -----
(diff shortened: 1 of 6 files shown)


src/nemos/observation_models.py
```diff
@@ -1030,7 +1030,7 @@ def log_likelihood(
         Returns
         -------
         :
-            The Bernoulli negative log-likehood. Shape (1,).
+            The Bernoulli negative log-likelihood. Shape (1,).
```
sjvenditto, line 1033:
```suggestion
            The Bernoulli negative log-likeleyhoud. Shape (1,).
```

PR #409, opened 2025-10-14
Glm hmm m step

BalzaniEdoardo:
## Implementation of M-step for GLM-HMM

This PR implements the M-step of the EM algorithm for GLM-HMMs, completing the inference framework alongside the existing E-step (forward-backward).

### Key Features

**Analytical M-step updates:**
- **Initial state probabilities**: Closed-form solution using posterior counts at session starts (Eq. 13.18, Bishop 2006)
- **Transition probabilities**: Closed-form solution from joint posteriors (Eq. 13.19, Bishop 2006)
- Both support optional Dirichlet priors for MAP estimation

**Numerical M-step update:**
- **GLM parameters**: Optimized via `solver_run` callable with flexible regularization
- Regularization/prior is embedded in the solver, supporting any `nemos` regularizer
- Works seamlessly for both single-neuron and population (multi-neuron) models

### Current Limitation: Dirichlet Prior Constraint

**The Dirichlet prior parameters (alphas) must currently satisfy α ≥ 1 for all states.**

**Why this restriction exists:**

The analytical M-step solution is derived by setting the gradient of the log-posterior to zero:
```
π_i ∝ (counts_i + α_i - 1)
```

This closed form is only valid when all `(counts_i + α_i - 1) > 0`. When α < 1 (sparse priors), the MAP solution may place some probabilities at the boundary (π_i = 0), requiring constrained numerical optimization instead.

**Interpretation of the constraint:**
- α = 1: Uniform/non-informative prior (equivalent to MLE with Laplace smoothing)
- α > 1: Concentrating/smoothing prior (encourages uniform distribution)
- α < 1: Sparse prior (encourages few active states) — **not currently supported**

**Future work:**

To support α < 1, we can replace the analytical updates with numerical optimization of the Lagrangian:
```python
L(π, λ) = Σ_i [counts_i + α_i - 1] log(π_i) + λ(Σ_i π_i - 1)
```

This would unify the M-step but sacrifice the computational efficiency of closed-form updates.

### References
- Bishop, C. M. (2006). *Pattern Recognition and Machine Learning*. Springer, Chapter 13.


src/nemos/glm_hmm/expectation_maximization.py
```diff
@@ -1,15 +1,65 @@
...
+        tmp_initial_prob += dirichlet_prior_alphas - 1
+
+    new_initial_prob = tmp_initial_prob / jnp.sum(tmp_initial_prob)
+    return new_initial_prob
+
+
+def _analytical_m_step_transition_prob(
+    joint_posterior: jnp.ndarray, dirichlet_prior_alphas: Optional[jnp.ndarray] = None
+):
+    if dirichlet_prior_alphas is not None:
```
sjvenditto, line 84: missing doc string

PR #411, opened 2025-10-22
Add port of FISTA and NAG from JAXopt

### LINKED ISSUE #380: Replace `OptaxOptimistixProximalGradient` with a port of `jaxopt.ProximalGradient`
As Optimistix doesn't implement `ProximalGradient`, this is currently set to use a custom implementation in `OptaxOptimistixProximalGradient` based on Optax's SGD with linesearch followed by the proximal operator.
This seems to work in practice, but is not as theoretically sound as `jaxopt.ProximalGradient` which implements the FISTA algorithm.

To be able to eventually stop vendoring (and drop support for) JAXopt, we decided to add a port of JAXopt's `ProximalGradient` that adheres to Optimistix's interface, and replace `OptaxOptimistixProximalGradient` with this when using the Optimistix backend.

bagibence:
As Optimistix doesn't implement `ProximalGradient`, this is currently set to use a custom implementation in `OptaxOptimistixProximalGradient` based on Optax's SGD with linesearch followed by the proximal operator.
This seems to work in practice, but doesn't support the original Nesterov acceleration and is not as theoretically sound as `jaxopt.ProximalGradient` which implements the FISTA algorithm.

This PR adds `OptimistixFISTA`, a port of JAXopt's `ProximalGradient` that adheres to Optimistix's interface, and replaces `OptaxOptimistixProximalGradient` with this when using the Optimistix backend.

Similarly, to support the original Nesterov acceleration in gradient descent, a short modification of this as `OptimistixNAG` is added. NAG stands for Nesterov's Accelerated Gradient.

Additional changes:
- Remove `OptimistixOptaxProximalGradient`
- Remove `stateful_scale_by_learning_rate` (and related code) which was only needed for `OptimistixOptaxProximalGradient`
- Switch `OptimistixOptaxGradientDescent` to use `optax.scale_by_backtracking_linesearch`

`OptimistixOptaxGradientDescent` is kept but is replaced in the tests. Ideally, it would be tested the same way as `OptimistixNAG`.
@BalzaniEdoardo, @billbrod is that fine with you, or is a way to include this in the tests without unnecessarily rerunning others required?

Fixes #380

### DIFF
--- a/tests/test_solvers.py
+++ b/tests/test_solvers.py
@@ -312,6 +312,7 @@ def test_svrg_glm_update(
             np.array([[0, 1, 0, 1, 1], [1, 0, 1, 0, 0]]).astype(float),
         ),
         ("GroupLasso", "ProximalGradient", np.array([[1, 1, 1, 1, 1]]).astype(float)),
+        ("Ridge", "GradientDescent", None),
         ("Ridge", "SVRG", None),
         ("UnRegularized", "SVRG", None),
     ],
@@ -324,7 +325,7 @@ def test_svrg_glm_update(
     "glm_class",
     [nmo.glm.GLM, nmo.glm.PopulationGLM],
 )
-def test_svrg_glm_fit(
+def test_maxiter_is_respected(
     glm_class,
     regularizer_name,
     solver_name,


src/nemos/solvers/_fista_port.py
```diff
@@ -0,0 +1,333 @@
+"""Adaptation of JAXopt's ProximalGradient (FISTA) as an Optimistix IterativeSolver."""
```
BalzaniEdoardo: I would rename this as _fista_linesearch.py or something like that. True that the algorithm is inspired by the jaxopt implementation but you did change enough of it to match the optimistix API that we can omit the fact that it is a port.

tests/conftest.py
```diff
@@ -1128,8 +1128,9 @@ def population_negativeBinomialGLM_model_instantiation_pytree(
     },
     "optimistix": {
         **_common_solvers,
-        "GradientDescent": nmo.solvers.OptimistixOptaxGradientDescent,
-        "ProximalGradient": nmo.solvers.OptimistixOptaxProximalGradient,
+        # TODO: OptaxOptimistixGradientDescent is not tested
```
BalzaniEdoardo: If I understand correctly, the default is using OptimistixNAG but we could still use the Optax GD if one wants to. 

One way to test this is to use unittest mock to mock the solver registry in the test:

```python
from unittest.mock import patch


# In your test
with patch.dict('nemos.solvers.solver_registry', 
                {'GradientDescent': nmo.solvers.OptimistixOptaxGradientDescent}):
    model = Model(solver_name='GradientDescent2', ...)
    model.instantiate_solver()
    
    assert isinstance(model._solver, nmo.solvers.OptimistixOptaxGradientDescent)
```

src/nemos/solvers/_fista_port.py
bagibence: I renamed it to `_fista.py` and rephrased the module's docstring, keeping "Adapted from JAXopt." to credit them.

tests/conftest.py
bagibence: The problem with that is that it would have to be done for every test that uses gradient descent.
I added a solution where only tests referencing `"GradientDescent"` explicitly are rerun with the `OptimistixOptaxGradientDescent` implementation and added it as a step to the CI job for the Optimistix backend.

bagibence: Changed this so that it is now controlled by `pytest -o override_solver=GradientDescent:OptimistixOptaxGradientDescent` and testing both implementations is done in the same tox environment.

BalzaniEdoardo (approved):
LGTM

PR #412, opened 2025-10-24
Glm hmm m step prior tests

camila-maura:
m step tests using dirichlet priors - regression tests comparing to data from sarah's julia code

### DIFF
--- a/tests/test_glm_hmm.py
+++ b/tests/test_glm_hmm.py
@@ -41,7 +41,7 @@ def prepare_partial_hmm_nll_single_neuron(obs):
             inverse_link_function=obs.default_inverse_link_function,
             negative_log_likelihood_func=negative_log_likelihood,
         )
-    
+
     solver = LBFGS(partial_hmm_negative_log_likelihood, tol=10**-13)
 
     return partial_hmm_negative_log_likelihood, solver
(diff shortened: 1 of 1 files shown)


tests/test_glm_hmm.py
```diff
@@ -1,4 +1,5 @@
 import itertools
+import pathlib
```
BalzaniEdoardo:
```suggestion
```
I think this is not used anymore

tests/test_glm_hmm.py
```diff
@@ -966,3 +967,124 @@ def test_m_step_set_alpha_transition_to_1(generate_data_multi_state):
...
+        return hmm_negative_log_likelihood(
+            weights,
+            X=design_matrix,
+            y=observations,
+            posteriors=posterior_prob,
+            inverse_link_function=obs.default_inverse_link_function,
+            negative_log_likelihood_func=negative_log_likelihood,
+        )
+
+    solver = LBFGS(partial_hmm_negative_log_likelihood, tol=10**-13)
```
BalzaniEdoardo: you should use this function

```python
solver, gammas, xis = prepare_solver_for_m_step_single_neuron(
    X, y, initial_prob, transition_prob, (coef, intercept), new_sess, obs
)
```

to remove some code repetition

BalzaniEdoardo (changes requested):
this looks good, I asked for a few changes, but after that and linting, this is good to go

BalzaniEdoardo (approved):
LGTM

PR #413, opened 2025-10-24
Progress bar

BalzaniEdoardo:
Add a progress bar for download dandi data.

### DIFF
--- a/src/nemos/fetch/fetch_data.py
+++ b/src/nemos/fetch/fetch_data.py
@@ -9,7 +9,6 @@ calculate hashes for local files and manage file paths.
 
 import os
 import pathlib
-import sys
 from typing import Optional, Union
 
 try:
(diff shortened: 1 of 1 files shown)


BalzaniEdoardo:
> Any reason not to do this for the pooch fetch as well? It's [pretty straightforward](https://github.com/plenoptic-org/plenoptic/blob/2798e9f853629e0f2c9e258bd95725b12b87ab5d/src/plenoptic/data/fetch.py#L175)

Yes, using dandi API client has performance gains over pooch because it has a built-in parallel download mechanism for large nwbs. Also, it has more specific error messages; but the main reason is the first one

BalzaniEdoardo:
@billbrod I added the progress-bar to pooch based downloads that do not interact with dandi.

billbrod:
My question was "why not add the progressbar", not "why have two separate functions". So this looks good to me!

PR #444, opened 2025-11-21
Support objective functions with auxiliary variables

### LINKED ISSUE #375: Support objective functions with aux?
Optimization libraries support objective functions that return not just a scalar function value, but also auxiliary data. These are mostly used for logging, diagnostics, and debugging.
Currently, the objective of GLMs does not have any aux, and throughout the codebase (e.g. in the `AbstractSolver` interface) `nemos` assumes that objective functions return a scalar only.

It should be considered if supporting objective functions that return aux is needed, and if yes, then how.

bagibence:
### Problem stated in #375:
Optimization libraries support objective functions that return not just a scalar function value, but also auxiliary data. These are mostly used for logging, diagnostics, and debugging.
Currently, the objective of GLMs does not have any aux, and throughout the codebase (e.g. in the AbstractSolver interface) nemos assumes that objective functions return a scalar only.

As other models might make use of aux, this PR prepares the solver interface and current models to deal with that.

### Main changes:
- (Prox-)SVRG can now handle objectives with aux. Following JAXopt solvers, aux is saved in the solver state. This saved aux does not come from the last evaluation of the objective or its gradient -- which is done on a minibatch -- but is the result of evaluating the gradient on the full data at the last reference point.
- `BaseRegressor` has a class attribute called `has_aux` that is passed to the solver on instantiation. Models whose objective returns aux have to overwrite this.
- Accordingly, `has_aux` is now a required argument of `AbstractSolver.__init__`.
- Adapt tests to these changes.

### Remaining questions, tasks:
- In order to avoid braking existing code, GLM.update doesn't return aux but saves it in `self.aux`. Should this be named `self.aux_` instead? Or returned?
- `GLM.update` performed the update, then estimated the scale using the previous parameters. Was that intended? I changed it to use the new parameters.
- `has_aux` is the only class attribute of `BaseRegressor`. Is that okay or should it be stored somewhere else? In any case, I will add a note about it to the developer notes.

Fixes #375

### DIFF
--- a/src/nemos/solvers/_optimistix_solvers.py
+++ b/src/nemos/solvers/_optimistix_solvers.py
@@ -214,7 +214,9 @@ class OptimistixAdapter(SolverAdapter[OptimistixSolverState]):
     def get_optim_info(self, state: OptimistixSolverState) -> OptimizationInfo:
         num_steps = self.stats["num_steps"].item()
 
-        function_val = state.f.item() if hasattr(state, "f") else state.f_info.f.item()  # pyright: ignore
+        function_val = (
+            state.f.item() if hasattr(state, "f") else state.f_info.f.item()
+        )  # pyright: ignore
 
         return OptimizationInfo(
             function_val=function_val,


src/nemos/glm/glm.py
```diff
@@ -1124,21 +1126,25 @@ def update(
...
-        self.solver_state_ = opt_step[1]
+        self._set_coef_and_intercept(params)
+        self.solver_state_ = opt_state
+        self.aux = aux
 
         # estimate the scale
         self.dof_resid_ = self._estimate_resid_degrees_of_freedom(
             X, n_samples=n_samples
         )
+        # TODO: This used the old params to predict. Was that a bug or intended?
```
BalzaniEdoardo: bug

src/nemos/glm/glm.py
```diff
@@ -788,6 +788,8 @@ def fit(
         # the output of loss. I believe it's the output of
         # solver.l2_optimality_error
         self.solver_state_ = state
+        # TODO: Should this be part of fit-state, so called aux_?
```
BalzaniEdoardo: probably also part of the fit state so that people running a loop externally, can concatenate the aux across iterations? if there is some  metric that users may want to track, it would make it easier

src/nemos/solvers/_optimistix_solvers.py
```diff
@@ -1,28 +1,42 @@
...
+def _make_f_and_aux_struct(
+    fn: Callable, has_aux: bool, y0: Params, args: Tuple[Any, ...]
+) -> Tuple[Pytree, Pytree]:
+    """
+    Derive f_struct and aux_struct. Adapted from Optimistix's optimise.
+
+    f_struct is "the shape+dtype of the output of `fn`".
+    aux_struct is the same for the returned aux.
+    """
+    y0 = jax.tree_util.tree_map(optx._misc.inexact_asarray, y0)
```
BalzaniEdoardo: what does this inexact_asarray do? and why are we using a private function? is there a public equivalent?

src/nemos/solvers/_optimistix_solvers.py
```diff
@@ -1,28 +1,42 @@
...
+    if not has_aux:
+        fn = optx._misc.NoneAux(fn)  # pyright: ignore
+    fn = optx._misc.OutAsArray(fn)
```
BalzaniEdoardo: These `_misc` functions are tiny wrappers - since they're internal utilities, we should consider porting them directly instead of relying on `optx._misc` for maintainability.

Specifically:
- The wrapper module has a `__call__` that returns `(func(x), None)` 
- `inexact_asarray` is just `jnp.asarray(x)` which converts numeric scalars to arrays at default float precision (per JAX docs: "all numeric scalar types with a (potentially) inexact representation")

Since these are so small, it would be more maintainable to inline them rather than depend on private Optimistix APIs.

BalzaniEdoardo: probaly `OutAsArray` is also another small one

src/nemos/solvers/_optimistix_solvers.py
```diff
@@ -117,6 +119,13 @@ def __init__(
                 user_args[kw] = solver_init_kwargs.pop(kw)
         self.config = OptimistixConfig(maxiter=maxiter, **user_args)
 
+        if has_aux:
+            self.fun_with_aux = lambda params, args: loss_fn(params, *args)
+            self.fun = lambda params, args: loss_fn(params, *args)[0]
+        else:
+            self.fun = lambda params, args: loss_fn(params, *args)
+            self.fun_with_aux = lambda params, args: (loss_fn(params, *args), None)
```
BalzaniEdoardo: you can directly use  `optx._misc.NoneAux` or its port if we have it. this is the same call used inside the function that returns the srcut of f and aux, so at this point we can have both matching

src/nemos/solvers/_svrg.py
```diff
@@ -441,10 +453,12 @@ def body_fun(step):
...
             # evaluate and store the full gradient with the params from the last inner loop
+            # aux is also saved from this evaluation instead of minibatches later
+            full_grad, new_aux = self.loss_gradient(prev_reference_point, *args)
+
             state = state._replace(
-                full_grad_at_reference_point=self.loss_gradient(
-                    prev_reference_point, *args
-                )
+                full_grad_at_reference_point=full_grad,
+                aux=new_aux,
```
BalzaniEdoardo: nice, so we have the aux as part of the state. is that true for every solver already?

macari216: This is true for all jaxopt and optmistix solvers we have

src/nemos/solvers/_jaxopt_solvers.py
```diff
@@ -56,10 +59,14 @@ def init_state(self, init_params: Params, *args: Any) -> JaxoptSolverState:
...
-        return self._solver.update(params, state, *self.hyperparams_prox, *args)
+        params, state = self._solver.update(
+            params, state, *self.hyperparams_prox, *args
+        )
+        return (params, state, state.aux)
 
     def run(self, init_params: Params, *args: Any) -> JaxoptStepResult:
-        return self._solver.run(init_params, *self.hyperparams_prox, *args)
+        params, state = self._solver.run(init_params, *self.hyperparams_prox, *args)
+        return (params, state, state.aux)
```
macari216: Just curious, is there a reason for returning aux explicitly in `solver.run` and `solver.update`? As opposed to returning just params and state (following the internal _solver API) and then setting `self.aux_ = opt_state.aux` in glm fit and update.

src/nemos/glm/glm.py
bagibence: Fixed.

src/nemos/glm/glm.py
bagibence: Do you mean solver state? By fit-state I meant what is extracted by `BaseRegressor._get_fit_state`

src/nemos/solvers/_svrg.py
bagibence: I don't think so. Optax doesn't support auxiliary variables and so `optimistix.OptaxMinimiser`, and with that `OptimistixOptaxGradientDescent` and `OptimistixOptaxLBFGS`, don't have it in the state.

src/nemos/solvers/_jaxopt_solvers.py
bagibence: That would work great if we knew that `opt_state.aux` exists for sure, but not every solver is guaranteed to store it in the state.
It's also consistent with [how `step` works in Optimistix](https://github.com/patrick-kidger/optimistix/blob/b414db226f6ade1854808c40e4a6c4fc4aba7f71/optimistix/_iterate.py#L82).

src/nemos/solvers/_optimistix_solvers.py
bagibence: I see your point. I'm not sure yet. They're all small, but they call a bunch of other ones that would have to be ported as well, and it adds up.
For now I removed the dependency and made my own little wrappers for the absolutely necessary ones.
I'll look into if and why the others are needed.

src/nemos/solvers/_optimistix_solvers.py
bagibence: Using the locally defined `_wrap_aux` now.

src/nemos/solvers/_optimistix_solvers.py
bagibence: See this (since then reverted) commit for a version where I copy all required helpers. I ended up adding a separate file for them: https://github.com/flatironinstitute/nemos/pull/444/commits/9fcc45f482b9da92a3705dadc40efcee1d43a95f

bagibence: The main point of having all these transformations here is to have `update` and `run` produce identical results, as `optx.minimise` called by `run` performs all of them inside. Skipping some could lead to aux and the state having different structure.

Cases where things can break without the transformations:
- without `OutAsArray` initializing the empty state breaks because it assumes that all leaves of `state.aux` are arrays. Adding BFGS to a test running update exposed this. For example `aux = {"a": 1, "b": 2}` and no conversion leads to an error trying to create an array based on the dict values' shape.
- In NeMoS this is controlled, but if `y0` (= `params`) had integers, `inexact_asarray` would be important. If a single value is optimized, it allows passing that as a scalar.

bagibence: An alternative solution for doing away without `optx._misc` is using `jnp.asarray` ([introduced in this commit](https://github.com/flatironinstitute/nemos/pull/444/commits/5b48716fea517a2e710a2ecf0b6ac4f667e2b260)). 

This, however, fails on the CI for Python 3.10. That is because this seems to work only for `jax>=0.7.2`, most likely because of ["JAX now represents constants in its internal jaxpr representation as a TypedNdArray"](https://docs.jax.dev/en/latest/changelog.html#jax-0-7-2-september-16-2025). But the last jax version that supports Python 3.10 is `0.6.2`.
(jax `0.7.0` and `0.7.1` also fail, but they are not allowed by Optimistix and Equinox anyway.)

So to use this version, support for Python 3.10 would have to be dropped and `jax>=0.7.2` required.
That would also be in line with [SPEC 0](https://scientific-python.org/specs/spec-0000/) saying that 3.10 and 3.11 are not supported.

bagibence:
Yes!

src/nemos/glm_hmm/expectation_maximization.py
```diff
@@ -955,6 +956,7 @@ def stopping_condition_while(carry):
         is_new_session,
     )
 
+    # TODO: what should be log and what not? Docstring doesn't mention log.
```
BalzaniEdoardo: This is fixed

src/nemos/glm/glm.py
BalzaniEdoardo: I see; I am not familiar with what people store in `aux_`, if it is generally lightweight, then I would say yes

bagibence: Okay, it's in there now.

src/nemos/glm_hmm/expectation_maximization.py
bagibence: All todos are resolved now.

BalzaniEdoardo (approved):
This also LGTM, once you resolver the conflict go ahead and merge

PR #452, opened 2025-12-11
Exponentiate log-probs in GLM-HMM

bagibence:
Docstring of `em_glm_hmm` and return type were misaligned: probability vs. log-probability for some elements.
It also lead to tests failing even though the returned values were mathematically correct.

Changes:
- Transform initial- and transition probabilities back from log.
- Unpack some values from the state to carry explicitly.


BalzaniEdoardo (approved):
LGTM

PR #453, opened 2025-12-12
Check inferred intercept is finite

bagibence:
When inferring the model intercept based on the output / firing rate, it can end up being infinite.
This came up for example in a test for Bernoulli observations and all one outputs.
In this case JAXopt solvers just stopped after 1 iteration, the [FISTA port](https://github.com/flatironinstitute/nemos/pull/411) using a different logic for convergence testing ran forever and produced a NaN output.

Changes:
- Raise an error if the intercept has non-finite values.
- Adjust a test to avoid this.
- Add a test for the `jax.lax.logistic` - `np.ones` pair.
(Not using the analytical inverse -- `lambda x: jax.lax.logistic(x)` -- initializes to a finite value.)

### DIFF
--- a/src/nemos/glm/initialize_parameters.py
+++ b/src/nemos/glm/initialize_parameters.py
@@ -28,6 +28,12 @@ INVERSE_FUNCS_BY_SIMPLE_NAME = {
     "one_over_x": one_over_x,
 }
 
+non_finite_error = ValueError(
+    "Failed to initialize the model intercept as the inverse of the firing rate for "
+    "the provided link function. The inferred intercept has non-finite values. "
+    "Please provide initial parameters instead."
+)
+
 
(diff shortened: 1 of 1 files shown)


src/nemos/glm/initialize_parameters.py
```diff
@@ -111,6 +111,12 @@ def initialize_intercept_matching_mean_rate(
     # return inverse if analytical solution is available
     analytical_inv = get_inverse_function(inverse_link_function)
 
+    non_finite_error = ValueError(
```
BalzaniEdoardo: move this outside the def, so that it is created once at module import.

bagibence: Done.

PR #461, opened 2026-01-02
Initialization utilities for the GLM-HMM parameters

BalzaniEdoardo:
# PR: GLM-HMM Parameter Initialization

## Overview

This PR adds a comprehensive initialization system for GLM-HMM parameters, providing sensible defaults and extensibility for custom initialization strategies. The initialization module coordinates the setup of all GLM-HMM parameters (GLM coefficients/intercepts, scale, initial state probabilities, and transition matrix) with proper random key management.

## Motivation

Proper initialization is critical for GLM-HMM convergence:
- **Poor initialization** can lead to local minima, slow convergence, or divergence
- **Random initialization** alone often fails for HMMs due to complex state dependencies
- **Users need flexibility** to provide custom initialization strategies for specific domains

This PR provides:
1. Sensible default initialization strategies
2. Clean registry-based system for custom initializers
3. Proper randomization via JAX random key splitting
4. Support for both single-neuron and population GLMs

#### `random_glm_params_init(n_states, X, y, inverse_link_function, random_key)`
Initializes GLM coefficients and intercepts:
- **Coefficients**: Small random normal values (scaled by 0.001) to avoid strong initial biases
- **Intercepts**: Matched to mean rate via inverse link function, ensuring reasonable initial firing rates
- **Tiling**: Intercepts are tiled across all states (same initial rate for each state)

```python
# Small random coefficients
coef = 0.001 * jax.random.normal(random_key, (n_features, n_neurons, n_states))

# Mean-rate intercepts via inverse link
intercept = initialize_intercept_matching_mean_rate(inverse_link_function, y)
intercept = jnp.tile(intercept[:, jnp.newaxis], (1, n_states))
```

**Why this matters**: Starting near the data's mean rate prevents extreme initial predictions while small random coefficients allow states to differentiate during training.

#### `ones_scale_init(n_states, X, y, random_key)`
Initializes scale parameters to 1.0 (deterministic).

#### `sticky_transition_proba_init(n_states, X, y, random_key, prob_stay=0.95)`
Initializes transition matrix with sticky dynamics:
- **Diagonal**: High probability of staying in current state (default 0.95)
- **Off-diagonal**: Low probability of transitioning to other states

```python
# For 3 states with prob_stay=0.95:
# [[0.95, 0.025, 0.025],
#  [0.025, 0.95, 0.025],
#  [0.025, 0.025, 0.95]]
```

**Why this matters**: Real neural/behavioral data often exhibits temporal persistence (states don't switch rapidly). Sticky initialization encourages the model to find stable state assignments rather than rapid switching.

#### `uniform_initial_proba_init(n_states, X, y, random_key)`
Initializes initial state probabilities uniformly (deterministic).

```python
initial_prob = jnp.ones(n_states) / n_states  # [1/3, 1/3, 1/3] for 3 states
```

### 2. Main Coordinator: `glm_hmm_initialization()`

Orchestrates initialization of all GLM-HMM parameters with proper random key management:

```python
def glm_hmm_initialization(
    n_states: int,
    X: DESIGN_INPUT_TYPE,
    y: NDArray | jnp.ndarray,
    inverse_link_function: Callable,
    random_key=jax.random.PRNGKey(123),
    init_registry: Optional[dict] = None,
) -> GLMHMMUserParams:
    """Initialize all GLM-HMM parameters."""
    # Properly split random key for each initialization function
    random_key, subkey = jax.random.split(random_key)
    coef, intercept = init_registry["glm_params_init"](n_states, X, y, inverse_link_function, subkey)

    random_key, subkey = jax.random.split(random_key)
    scale = init_registry["scale_init"](n_states, X, y, subkey)

    random_key, subkey = jax.random.split(random_key)
    initial_proba = init_registry["initial_proba_init"](n_states, X, y, subkey)

    _, subkey = jax.random.split(random_key)
    transition_proba = init_registry["transition_proba_init"](n_states, X, y, subkey)

    return coef, intercept, scale, initial_proba, transition_proba
```

**Key features**:
- Each initialization function receives a **unique subkey** (properly split from parent key)
- Returns all 5 GLM-HMM parameters as a tuple
- Uses default registry if none provided
- Merges partial registries with defaults (override only what you need)

### 3. Registry System

Clean, extensible system for managing initialization functions:

```python
DEFAULT_INIT_FUNCTION = {
    "glm_params_init": random_glm_params_init,
    "scale_init": ones_scale_init,
    "transition_proba_init": sticky_transition_proba_init,
    "initial_proba_init": uniform_initial_proba_init,
}
```

**Custom initialization example**:
```python
def my_custom_glm_init(n_states, X, y, inverse_link_function, random_key):
    # Custom logic here
    return custom_coef, custom_intercept

custom_registry = {
    "glm_params_init": my_custom_glm_init,
    # Other parameters use defaults
}

params = glm_hmm_initialization(n_states, X, y, inverse_link, init_registry=custom_registry)
```

The registry system:
- Validates function signatures (correct number of parameters)
- Supports string lookup for built-in functions
- Allows custom callables
- Merges partial registries with defaults

### 4. Special Handling: `glm_params_init` Signature

GLM parameter initialization uniquely requires `inverse_link_function` (5 parameters) while others need only 4:

```python
# glm_params_init: (n_states, X, y, inverse_link_function, random_key)
# Other init funcs: (n_states, X, y, random_key)
```

The validation logic detects this and enforces the correct signature:

```python
expected_params = 5 if func_name == "glm_params_init" else 4
```

**Why this matters**: GLM intercepts need the inverse link function to match mean rates (e.g., log(mean) for Poisson, identity for Gaussian).

### 5. Utility: `resolve_dirichlet_priors()`

Validates and converts Dirichlet prior alpha parameters (for future Bayesian extensions):

```python
def resolve_dirichlet_priors(alphas: Any, expected_shape: Tuple[int, ...]) -> jnp.ndarray | None:
    """Validate Dirichlet prior alpha parameters."""
    # Checks: correct shape, all values >= 1
    # Returns: JAX array or None
```

### Function Signature Validation

Custom initialization functions are validated to ensure correct signatures:

```python
def _resolve_init_func(func_name: str, init_func: Callable | str):
    """Validate and resolve initialization function."""
    if callable(init_func):
        sig = inspect.signature(init_func)
        n_params = len(sig.parameters)

        expected_params = 5 if func_name == "glm_params_init" else 4

        if n_params < expected_params:
            raise ValueError(f"Function must have at least {expected_params} parameters")

        # Extra parameters must have defaults
        params_without_defaults = [
            p.name for p in list(sig.parameters.values())[expected_params:]
            if p.default is inspect.Parameter.empty
        ]
        if params_without_defaults:
            raise ValueError(f"Parameters {params_without_defaults} must have defaults")
```

### Default Initialization

```python
from nemos.glm_hmm.initialize_parameters import glm_hmm_initialization

# Initialize with defaults
coef, intercept, scale, initial_prob, transition_prob = glm_hmm_initialization(
    n_states=3,
    X=X_train,
    y=y_train,
    inverse_link_function=jnp.exp,  # For Poisson GLM
    random_key=jax.random.PRNGKey(42)
)
```

### Custom Initialization

```python
def domain_specific_init(n_states, X, y, inverse_link_function, random_key):
    """Initialize with domain knowledge."""
    # Custom logic: maybe use PCA to initialize different states
    coef = custom_pca_initialization(X, n_states, random_key)
    intercept = custom_intercept_logic(y)
    return coef, intercept

custom_registry = {
    "glm_params_init": domain_specific_init,
    # Other parameters use defaults
}

params = glm_hmm_initialization(
    n_states=3,
    X=X_train,
    y=y_train,
    inverse_link_function=jnp.exp,
    init_registry=custom_registry,
    random_key=jax.random.PRNGKey(42)
)
```

### String Lookup for Built-ins

```python
custom_registry = {
    "glm_params_init": "random",  # String lookup
    "transition_proba_init": "sticky",
    # Others use defaults
}

params = glm_hmm_initialization(n_states, X, y, inverse_link, init_registry=custom_registry)
```

## Future Extensions

The registry pattern makes it easy to add:
- Custom user defined initializations.
- Data-driven initialization (e.g., K-means clustering for states)
- Observation model-specific initialization strategies

### DIFF
--- a/src/nemos/glm_hmm/initialize_parameters.py
+++ b/src/nemos/glm_hmm/initialize_parameters.py
@@ -356,6 +356,8 @@ def glm_hmm_initialization(
         init_registry = DEFAULT_INIT_FUNCTION
     else:
         init_registry = _resolve_init_funcs_registry(init_registry)
+    if init_kwargs is None:
+        init_kwargs = {}
     random_key, subkey = jax.random.split(random_key)
     kwargs = init_kwargs.get("glm_params_init", {})
     coef, intercept = init_registry["glm_params_init"](
--- a/tests/test_glm_hmm_initialization.py
+++ b/tests/test_glm_hmm_initialization.py
@@ -793,6 +793,130 @@ class TestGLMHMMInitialization:
 
         assert len(result) == 5
 
+    def test_init_kwargs_prob_stay(self):
+        """Test that prob_stay kwarg is passed to transition_proba_init."""
+        n_states = 3
+        X = jnp.ones((100, 5))
+        y = jnp.ones(100)
+        inverse_link = lambda x: x
+
+        # Use custom prob_stay
+        custom_prob_stay = 0.8
+        init_kwargs = {"transition_proba_init": {"prob_stay": custom_prob_stay}}
+
+        _, _, _, _, transition_prob = glm_hmm_initialization(
+            n_states,
+            X,
+            y,
+            inverse_link,
+            random_key=jax.random.PRNGKey(123),
+            init_kwargs=init_kwargs,
+        )
+
+        # Diagonal should be prob_stay
+        assert jnp.allclose(jnp.diag(transition_prob), custom_prob_stay)
+
+        # Off-diagonal should be (1 - prob_stay) / (n_states - 1)
+        expected_off_diag = (1 - custom_prob_stay) / (n_states - 1)
+        off_diag = transition_prob[~jnp.eye(n_states, dtype=bool)]
+        assert jnp.allclose(off_diag, expected_off_diag)
+
+    def test_init_kwargs_std_dev(self):
+        """Test that std_dev kwarg is passed to glm_params_init."""
+        n_states = 2
+        n_features = 5
+        X = jnp.ones((100, n_features))
+        y = jnp.ones(100)
+        inverse_link = lambda x: x
+
+        # Use large std_dev to make effect measurable
+        large_std_dev = 1.0
+        init_kwargs = {"glm_params_init": {"std_dev": large_std_dev}}
+
+        coef_large, _, _, _, _ = glm_hmm_initialization(
+            n_states,
+            X,
+            y,
+            inverse_link,
+            random_key=jax.random.PRNGKey(123),
+            init_kwargs=init_kwargs,
+        )
+
+        # Use small std_dev
+        small_std_dev = 0.001
+        init_kwargs = {"glm_params_init": {"std_dev": small_std_dev}}
+
+        coef_small, _, _, _, _ = glm_hmm_initialization(
+            n_states,
+            X,
+            y,
+            inverse_link,
+            random_key=jax.random.PRNGKey(123),
+            init_kwargs=init_kwargs,
+        )
+
+        # Large std_dev should produce larger magnitude coefficients
+        assert jnp.abs(coef_large).max() > jnp.abs(coef_small).max()
+        # Small std_dev should be close to 0
+        assert jnp.abs(coef_small).max() < 0.01
+
+    def test_init_kwargs_multiple_functions(self):
+        """Test that init_kwargs can pass kwargs to multiple init functions."""
+        n_states = 3
+        X = jnp.ones((100, 5))
+        y = jnp.ones(100)
+        inverse_link = lambda x: x
+
+        init_kwargs = {
+            "glm_params_init": {"std_dev": 0.5},
+            "transition_proba_init": {"prob_stay": 0.7},
+        }
+
+        coef, _, _, _, transition_prob = glm_hmm_initialization(
+            n_states,
+            X,
+            y,
+            inverse_link,
+            random_key=jax.random.PRNGKey(123),
+            init_kwargs=init_kwargs,
+        )
+
+        # Check std_dev effect
+        assert jnp.abs(coef).max() > 0.01  # Larger than default 0.001
+
+        # Check prob_stay effect
+        assert jnp.allclose(jnp.diag(transition_prob), 0.7)
+
+    def test_init_kwargs_empty_dict(self):
+        """Test that empty init_kwargs dict works (uses defaults)."""
+        n_states = 2
+        X = jnp.ones((100, 5))
+        y = jnp.ones(100)
+        inverse_link = lambda x: x
+
+        # Empty dict should behave like None
+        result_empty = glm_hmm_initialization(
+            n_states,
+            X,
+            y,
+            inverse_link,
+            random_key=jax.random.PRNGKey(123),
+            init_kwargs={},
+        )
+
+        result_none = glm_hmm_initialization(
+            n_states,
+            X,
+            y,
+            inverse_link,
+            random_key=jax.random.PRNGKey(123),
+            init_kwargs=None,
+        )
+
+        # Results should be identical
+        for r1, r2 in zip(result_empty, result_none):
+            assert jnp.allclose(r1, r2)
+
 
 class TestResolveDirichletPriors:
     """Test _resolve_dirichlet_priors validation function."""


BalzaniEdoardo:
Ready for reviews after #456 is merged.

src/nemos/glm_hmm/initialize_parameters.py
```diff
@@ -0,0 +1,503 @@
...
+        Intercept array of shape (n_neurons, n_states).
+    """
+    n_features = X.shape[1]
+    is_one_dim = y.ndim == 1
+    n_neurons = 1 if is_one_dim else y.shape[1]
+
+    # small random noisy coef
+    coef = 0.001 * jax.random.normal(random_key, (n_features, n_neurons, n_states))
+    # mean-rate
+    intercept = initialize_intercept_matching_mean_rate(inverse_link_function, y)
```
sjvenditto, line 117: I'm not sure if this is a useful initialization strategy for Bernoulli? Also, do we really want all the intercepts initialized to the same value? It might be OK since the coefficients are different, but it's hard to say without running test fits

src/nemos/glm_hmm/initialize_parameters.py
```diff
@@ -0,0 +1,503 @@
...
+        Coefficient matrix of shape (n_features, n_neurons, n_states).
+    intercept : jnp.ndarray
+        Intercept array of shape (n_neurons, n_states).
+    """
+    n_features = X.shape[1]
+    is_one_dim = y.ndim == 1
+    n_neurons = 1 if is_one_dim else y.shape[1]
+
+    # small random noisy coef
+    coef = 0.001 * jax.random.normal(random_key, (n_features, n_neurons, n_states))
```
sjvenditto: I think the `0.001` should be a user-settable parameter

src/nemos/glm_hmm/initialize_parameters.py
```diff
@@ -0,0 +1,503 @@
...
+    """Initialize scale to ones."""
+    is_one_dim = y.ndim == 1
+    n_neurons = 1 if is_one_dim else y.shape[1]
+    scale = jnp.ones((n_neurons, n_states), dtype=float)
+    if is_one_dim:
+        scale = jnp.squeeze(scale, axis=0)
+    return scale
+
+
+def sticky_transition_proba_init(
```
sjvenditto, line 140: I would also add a method for uniform initial transition proba

src/nemos/glm_hmm/initialize_parameters.py
```diff
@@ -0,0 +1,503 @@
...
+}
+
+
+def glm_hmm_initialization(
+    n_states: int,
+    X: DESIGN_INPUT_TYPE,
+    y: NDArray | jnp.ndarray,
+    inverse_link_function: Callable,
+    random_key=jax.random.PRNGKey(123),
+    init_registry: Optional[dict] = None,
```
sjvenditto, line 312: Should you rename this to `init_funcs` to more closely match how it will be named in the GLMHMM init signature?

Also, for additional kwargs, it could be useful to have a second dictionary like `init_kwargs`. It could be formatted like a nested dictionary... It can get a bit complicated, but this would be an advanced use case anyway. Something like:
```
init_kwargs = {"glm_params_init": {"std_dev": 0.001}}
```

src/nemos/glm_hmm/initialize_parameters.py
BalzaniEdoardo: that is true, we have tests later on in the PR stack. Before the intercept was initialized totally randomly as well, but this seems to work better because it doesn't generate -inf log-likelihood, which was happening with totally random init.

For Bernoulli, it sets the intercept so that `logit(intercept)` is the empirical probability of observing a 1. The noise acts as a small perturbation to break the symmetry due to the intercept being the same for all states. 

```python
In  [1]: import nemos as nmo
    ...: import numpy as np
    ...: 
    ...: np.random.seed(42)
    ...: y=np.random.choice([0,1], size=100)
    ...: X = np.random.randn(100,3)
    ...: model = nmo.glm_hmm.GLMHMM(3)
    ...: params = model.initialize_params(X, y)
    ...: 
    ...: print("init prob:", model.inverse_link_function(params[1]))
    ...: print("empirical prob: ", y.mean(axis=0))
init prob: [0.56 0.56 0.56]
empirical prob:  0.56
```

I don't know if it would make more sense to add a little perturbation to the intercept too, but if X is non-zero we already are adding noise in the rate associated to each state.

src/nemos/glm_hmm/initialize_parameters.py
BalzaniEdoardo: done it! I also changed the typing for the init function using a protocol to make it explicit that one define additional kwargs for it

Issue #393, opened 2025-09-11
Make basis functions jit compilable


sjvenditto:
Right now, basis functions can't be jit compiled for at least two reasons:
1. the pynapple support decorator is hard-coded to convert to numpy
2. the min-max rescale of features uses boolean indexing to set out-of-range values in-place

Additionally, some functions (e.g. `evaluate` functions) have multiple, redundant pynapple support decorators. This is because the decorator `@check_transform_input` is already decorated for pynapple support

BalzaniEdoardo:
#604 add comprehensive tests and fixes BSpline jiting.

PR #468, opened 2026-01-08
Make `compute_features` jit-compatible for most basis classes

sjvenditto:
This PR addresses issue #393 and updates evaluation and convolution code to use jax functions and syntax to make bases' `compute_features` method jit-compilable. The compatible basis functions are:
- CustomBasis
- RaisedCosineLogEval
- RaisedCosineLogConv
- RaisedCosineLinearEval
- RaisedCosineLinearConv
- MSplineEval
- MSplineConv
- IdentityEval
- HistoryConv
- FourierEval

Some basis classes depend on scipy functions that currently don't have a jax port, and are therefore still not jit-compatible. These basis functions are:
- BSplineEval
- BSplineConv
- CyclicBSplineEval
- CyclicBSplineConv
- OrthExponentialEval
- OrthExponentialConv

To test compatibility, a new test was added in `test_basis.py` in the `TestSharedMethods` class: `test_jitted_compute_features`

## Current issues:
1. As noted in the issue, the validation step `check_fraction_valid_samples` is by nature incompatible with jit-compilation. Right now, the problem code has been commented out, but we'll need to discuss removing this validation step altogether.
2.  Since `min_max_rescale_samples` has been modified to be jit-compatible, it converts sample arrays to jax arrays. Since the default precision is 32-bit, this is causing issues with evaluate functions of BSpline and OrthExponential
- BSpline's `need_outer` check will fail with 32-bit precision
- OrthExponential rank check with `np.linalg.matrix_rank` can drop basis functions with 32-bit precision

Both of these problems are solved with forcing 64-bit precision. However, since neither of these basis classes are jit compatible, it might be easier to have separate `min_max_rescale_samples` logic that never converts the samples to jax arrays, thus bypassing the need to set to 64-bit precision in all relevant tests and docstrings

I see two ways of handling this: 
1. Have a separate numpy version of `min_max_rescale_samples` for BSpline and OrthExponential bases
2. Check for whether or not samples is a traced array in `min_max_rescale_samples` and use numpy functions when the array is not traced

### DIFF
--- a/tests/test_basis.py
+++ b/tests/test_basis.py
@@ -1735,7 +1735,7 @@ class TestSharedMethods:
             **extra_kwargs(cls, args_copy["n_basis_funcs"]),
         )
 
-        for method in  ["evaluate", "compute_features"]:
+        for method in ["evaluate", "compute_features"]:
             jitted_compute_features = jax.jit(getattr(basis_obj, method))
             # Clear any existing cache
             jitted_compute_features._cache_size()  # warmup if needed
--- a/src/nemos/basis/_custom_basis.py
+++ b/src/nemos/basis/_custom_basis.py
@@ -543,7 +543,7 @@ class CustomBasis(BasisMixin, BasisTransformerMixin, Base):
                 [ 4., 16.]],
         ...
                [[ 5., 25.],
-                [ 6., 36.]]])
+                [ 6., 36.]]], dtype=float32)
         """
         # ruff: noqa: D205, D400
         return super().split_by_feature(x, axis=axis)


tests/test_basis.py
```diff
@@ -415,6 +415,7 @@ def method(self):
     ],
 )
 def test_expected_output_eval_on_grid(basis_instance, super_class):
+    jax.config.update("jax_enable_x64", True)
```
BalzaniEdoardo:
```suggestion
@pytest.mark.requires_x64
def test_expected_output_eval_on_grid(basis_instance, super_class):
```
There is a new marker you can use here.

PR #433, opened 2025-11-14
197 pytree regularizer strength

wulfdewolf:
# Finer control over regularization strength

## Summary 
This PR extends NeMoS’ regularization API to allow per-parameter control of regularization strength. In addition to scalar strengths, users can now supply pytrees matching the model parameter structure to regularize different parameters (or groups) with different strengths. `ElasticNet` now supports a tuple (strength, ratio) where either or both can be pytrees. `GroupLasso` now allows per-group strengths.

## Motivation
- When fitting a `PopulationGLM`, individual neurons can be regularized independently.
- When new models are added, we will want more fine control over which parameters are regularized.

## Key changes

- Accept strength as pytree
  * `regularizer_strength` now accepts: `None` (defaults appropriately per-regularizer), scalar (float / 0-D array), or a pytree matching the structure of the regularizable parameters.
  * For ElasticNet: a tuple where each of `(strength, ratio)` can be a scalar or pytree. 
  * Regressor signatures updated accordingly (typing changed to `Any`). 

- Regularizer interface refactor
  * `proximal_operator` field, allowing for a superclass `get_proximal_operator`
  * `get_proximal_operator(params, strength)`: now closes over strength (still possible to pass hyperparameters to proximal operator, but ignored for now).
  * `penalized_loss(loss, params, strength)`: embeds penalty with structured strengths.
  * New validation helpers:
     - `_validate_strength`: converts scalars or trees to JAX arrays; defaults `None` as appropriate.
     - `_validate_strength_structure`: broadcasts/aligns strengths to regularizable parameter subtrees and inserts `None` elsewhere.
  * `_get_filter_kwargs(params, strength)`: standardized way to pass masked/broadcast strengths into penalties and proximal operators.
  * Penalty computation now tree-maps over `(params, strength)` pairs; intercepts remain unregularized via `regularizable_subtrees()`, as before.

- Proximal operators generalized, all `prox_` functions now accept a `strength` pytree
  * `prox_none(x, strength, scaling)`
  * `prox_ridge(x, strength, scaling)`: divides by `(1 + scaling * strength)` leafwise.
  * `prox_lasso(x, strength, scaling)`: soft-threshold leafwise by `scaling * strength`.
  * `prox_elastic_net(x, strength, scaling)`: expects per-leaf `(strength, ratio)`; applies combined L1/L2 shrinkage.
  * `prox_group_lasso(x, strength, mask, scaling)`: expects per-group strengths (vector); applies group-wise shrinkage.

- Solver integration
[...]

PR #470, opened 2026-01-13
Group lasso generalization

BalzaniEdoardo:
# Generalize Group Lasso to Support PyTree Parameters

## Overview

This PR significantly refactors the Group Lasso regularizer to work with arbitrary PyTree parameter structures, enabling automatic variable selection with `FeaturePytree` and supporting more flexible grouping patterns for both GLM and PopulationGLM models.

## Motivation

The previous Group Lasso implementation had several limitations:

1. **Fixed 2D mask structure**: Masks were restricted to `(n_groups, n_features)` shape, making it incompatible with:
   - PyTree-structured parameters (e.g., `FeaturePytree`)
   - Multi-neuron models where mask shape should be `(n_groups, n_features, n_neurons)`

2. **Manual mask initialization**: Users had to construct masks manually before fitting, even for simple use cases like per-covariate grouping

3. **Limited flexibility**: The rigid mask structure prevented natural extensions to hierarchical or structured parameter spaces

#### 1. **Generalized Mask Structure** 

- **Before**: Masks were 2D arrays of shape `(n_groups, n_features)`
- **After**: Masks are PyTrees with the same structure as model parameters
  - Each leaf mask has shape `(n_groups, *param_leaf_shape)`
  - For GLM: `mask.coef` has shape `(n_groups, n_features)`
  - For PopulationGLM: `mask.coef` has shape `(n_groups, n_features, n_neurons)`
  - Non-regularized parameters (e.g., intercepts) can have `None` or zero masks

#### 2. **Automatic Mask Initialization** 

Added `initialize_mask(x)` method that creates sensible default masks from parameter structure:

```python
def initialize_mask(self, x: Any) -> Any:
    """Initialize a default group mask for a PyTree of parameters."""
    # For each regularizable subtree:
    # - Creates one group per trailing dimension combination
    # - E.g., (n_features, d1, d2) → d1*d2 groups
    # - Non-regularizable leaves get None masks
```

**Key features**:
- Detects regularizable vs non-regularizable parameters using `regularizable_subtrees()`
- Creates independent groupings for each regularizable subtree
- Sensible defaults: each trailing dimension becomes a separate group

#### 3. **Updated Proximal Operator**

- `prox_group_lasso` now operates on PyTree parameters and masks
- Uses new `masked_norm_2` function for computing group-wise norms
- Automatically handles both GLM and PopulationGLM parameter shapes

#### 4. **Enhanced `apply_operator` and `_penalization`**
Added `filter_kwargs` parameter to route PyTree-structured metadata (like masks) to subtree-specific operations:

```python
def apply_operator(func, params, *args, filter_kwargs=None, **kwargs):
    """
    Apply an operator to regularizable subtrees.

    filter_kwargs: dict of PyTree values that are automatically filtered
                   to match each subtree (e.g., {'mask': full_mask})
    """
```

This allows masks to be:
- Defined once with full parameter structure
- Automatically sliced for each regularizable subtree
- Passed to proximal operators without manual coordination

#### Backward Compatibility

**Breaking change**: Mask format has changed from plain arrays to PyTree structure.

**Migration path**:
```python
# OLD (no longer works)
mask = np.array([[1, 0, 1, 0], [0, 1, 0, 1]])  # 2D array
regularizer = GroupLasso(mask=mask)

# NEW (required format) 
from nemos.glm.params import GLMParams
mask = GLMParams(
    coef=jnp.array([[1, 0, 1, 0], [0, 1, 0, 1]]),
    intercept=None  # or zeros if you want to include intercept in groups
)
regularizer = GroupLasso(mask=mask)

# EASIEST: Let it auto-initialize
regularizer = GroupLasso(mask=None)  # Mask initialized during fit
```

**Note** that the user can set the mask directly, but the intended usage is to fit a GLM with `FeaturePytree` inputs, which will automatically result in grouped coefficients.

#### New Workflow: Automatic Variable Selection

The refactored Group Lasso enables seamless variable selection with `FeaturePytree`:

```python
# Define features with multiple covariates
bas = (
    nmo.basis.RaisedCosineLinearEval(6, label="position") +
    nmo.basis.RaisedCosineLinearEval(6, label="speed")
)
predictors = nmo.pytrees.FeaturePytree(**bas.split_by_feature(X))

# Group Lasso automatically creates one group per feature
model = nmo.glm.GLM(regularizer="GroupLasso")
model.fit(predictors, y)

# Entire "speed" group can be shrunk to zero, performing variable selection
```

#### Threading `init_params` Through the Stack

To enable automatic mask initialization, `init_params` now flows through:

1. **GLM.fit()** → gets initial parameter structure
2. **`_instantiate_solver(loss, init_params)`** → receives parameters
3. **`SolverAdapter.__init__(loss, init_params)`** → stores for regularizer
4. **`regularizer.get_proximal_operator(init_params)`** → initializes mask if needed
5. **`regularizer.penalized_loss(params, init_params)`** → initializes mask if needed

This design:
- ✅ Lazy initialization: mask created only when needed
- ✅ Safe: mask matches actual parameter structure used in optimization
- ✅ Automatic: works without user intervention for common cases

#### Subtree-Specific Regularization

The `filter_kwargs` pattern enables clean separation:

```python
# In regularizer
def _get_filter_kwargs(self):
    if self.mask is not None:
        return {"mask": self.mask}  # Full PyTree mask
    return {}

# In apply_operator
for where in params.regularizable_subtrees():
    subtree_kwargs = {key: where(val) for key, val in filter_kwargs.items()}
    # subtree_kwargs now has just the mask piece for this subtree!
    func(where(params), *args, **subtree_kwargs)
```

#### New How-To Guide

Created [`docs/how_to_guide/variable_selection_group_lasso.md`](docs/how_to_guide/variable_selection_group_lasso.md):

- **Overview**: When and why to use Group Lasso for variable selection
- **Example**: Demonstrates covariate selection with position and speed
- **Automatic grouping**: Explains how FeaturePytree enables seamless variable selection
- **Custom masks**: Points to advanced usage patterns
- **Interpretation**: How to analyze regularization paths

#### Updated Docstrings

- `GroupLasso.initialize_mask`: Detailed documentation of default mask construction
- `apply_operator`: Explained `filter_kwargs` for PyTree metadata routing
- `prox_group_lasso`: Updated to reflect PyTree parameters and masks
- `masked_norm_2`: Documented behavior with PyTree inputs

### 1. **Automatic Variable Selection**

```python
# No manual mask construction needed!
model = nmo.glm.GLM(regularizer="GroupLasso")
model.fit(feature_pytree, y)
# Each feature is automatically a separate group
```

### 2. **Unified Multi-Neuron Support**

```python
# Same API works for both GLM and PopulationGLM
model = nmo.glm.PopulationGLM(regularizer="GroupLasso")
model.fit(X, Y)  # Y shape: (n_samples, n_neurons)
# Mask automatically has correct 3D shape
```

### 3. **Flexible Grouping**

```python
# Custom hierarchical groups
mask = {
    "spatial": jnp.array([...]),  # Groups for spatial features
    "temporal": jnp.array([...]), # Groups for temporal features
}
regularizer = GroupLasso(mask=mask)
```

### 4. **Consistent with NeMoS Design**

- Follows PyTree-first design philosophy
- Works naturally with `FeaturePytree` and `basis` objects
- Leverages `regularizable_subtrees()` protocol

### For Users with Existing Group Lasso Code

**If using explicit masks:**

```python
# Before
mask = np.array([[1, 0, 1], [0, 1, 0]])
model = nmo.glm.GLM(regularizer=nmo.regularizer.GroupLasso(mask=mask))

# After
from nemos.glm.params import GLMParams
mask = GLMParams(coef=jnp.array([[1, 0, 1], [0, 1, 0]]), intercept=None)
model = nmo.glm.GLM(regularizer=nmo.regularizer.GroupLasso(mask=mask))
```

**For PopulationGLM:**

```python
# Mask shape changes from (n_groups, n_features) to (n_groups, n_features, n_neurons)
n_neurons = 5
mask_array = np.ones((2, 3, n_neurons))  # 2 groups, 3 features, 5 neurons
mask = GLMParams(coef=jnp.asarray(mask_array), intercept=None)
```

**Best practice: Use automatic initialization**

```python
# Simplest approach - let Group Lasso initialize the mask
model = nmo.glm.GLM(regularizer="GroupLasso")
model.fit(X, y)  # Mask created automatically based on X structure
```

## Acknowledgments

This refactoring enables Group Lasso to work seamlessly with NeMoS's PyTree-based parameter structures, significantly improving usability for common variable selection workflows while maintaining flexibility for advanced use cases.

### DIFF
--- a/src/nemos/regularizer.py
+++ b/src/nemos/regularizer.py
@@ -679,7 +679,7 @@ class GroupLasso(Regularizer):
     >>> # Create the GroupLasso regularizer instance
     >>> group_lasso = GroupLasso(mask=mask)
     >>> # fit a group-lasso glm
-    >>> model = GLM(regularizer=group_lasso, strength=0.1).fit(X, y)
+    >>> model = GLM(regularizer=group_lasso, regularizer_strength=0.1).fit(X, y)
     >>> print(f"coeff shape: {model.coef_.shape}")
     coeff shape: (5,)
     """


src/nemos/base_regressor.py
```diff
@@ -315,7 +316,7 @@ def _check_solver_kwargs(solver_class: Type, solver_kwargs: dict[str, Any]) -> N
             )
 
     def _instantiate_solver(
-        self, loss, solver_kwargs: Optional[dict] = None
+        self, loss, init_params: Any, solver_kwargs: Optional[dict] = None
```
wulfdewolf: Missing docstring for `init_params`.

Now we only pass `init_params` so that it can be stored and the regularizer can use it later to check mask structure?
But if we are passing `init_params` anyways, `instantiate_solver` and `_initialize_solver_and_state` start looking very similar. 
Could they be merged? Or are we sometimes really only instantiating the solver?

src/nemos/regularizer.py
```diff
@@ -200,20 +259,28 @@ def _penalized_loss(params, *args, **kwargs):
...
-            penalty = penalty + self._penalty_on_subtree(params, strength)
+            penalty = penalty + self._penalty_on_subtree(
+                params, strength, **filter_kwargs
+            )
         return penalty
 
     @abc.abstractmethod
     def _penalty_on_subtree(
-        self, sub_params, strength: RegularizerStrength
+        self, sub_params, regularizer_strength: RegularizerStrength, **kwargs
```
wulfdewolf: we need to agree on conventions, I have been using just `strength`.
Given that we are in a regularizer class I thought the `regularizer_` part is implied, but happy to go back.

src/nemos/regularizer.py
```diff
@@ -702,48 +870,41 @@ def _penalty_on_subtree(
...
+        ----------
+        init_params
+
         Returns
         -------
         :
             The proximal operator, applying Group Lasso regularization to the provided parameters. The
             intercept term is not regularized.
         """
+        filter_kwargs = self._get_filter_kwargs(init_params=init_params)
```
wulfdewolf, line 887: This works and I have no real issue with it in this context. 
Similarly for how this works in the non-proximal route with `penalized_loss`.

However, once we start thinking about flexible `regularizer_strength`, see this [PR](https://github.com/flatironinstitute/nemos/pull/433), things become a bit more complex.
There, we want to allow passing pytrees of regularizer strengths that match the parameter structure, allowing very fine control over the regularisation of each parameter.
For `Ridge`, `Lasso` and `ElasticNet` this makes sense, but for `GroupLasso` it doesn't. 
For `GroupLasso` specifically, we limit the fine control to regularizing each _group_ individually. 

This means that at some point, the `GroupLasso` mask and the `regularizer_strength` will have to be matched. 
In the other PR I opted for a very explicit `_validate_regularizer_strength_structure` method that matches the passed strength to the parameters. It comprehensively checks all possible cases.
On the other hand, in this one you opted for a more free approach where the mask and parameters are checked at run time, i.e. in `get_proximal_operator` (or `_penalized_loss` depending on the solver).

I feel that, once we merge both PRs, we should decide on one of the two options for both.
Either, we have explicit structure checks for the mask and strength separately, and in that order since the strength needs the mask. 
Or, we take a more hands-off approach and assume that the user that uses these advanced options knows what they are doing and will realise that if they get errors, the mask and strength might not match each other, or the parameters.

wulfdewolf: Pure beauty-wise I also prefer having `_validate_...` functions since they can take `init_params` as an argument much more intuitively, and earlier, than `get_proximal_operator`/`penalized_loss`. 
The validation functions would be called from the regressor object around the time of the solver instantiation.

wulfdewolf: As a counter-argument, those validate functions will add a lot more code for functionality that is very niche.

src/nemos/regularizer.py
BalzaniEdoardo: I am fine with going with your notation

src/nemos/regularizer.py
BalzaniEdoardo: We should favor the solution that maximally disentangle the parameter structure and the regularizer. Ideally, the user never sets the mask, and when that happens the mask is set based on the parameters. We can add a validate step at runtime, in case the mask was set, that could happen inside this method. 

But also, I think the scope of the method is more general than validation, and it may well be that in the future we will need more kwargs that are sub-tree specific. This to say, a validation can be added, but this machinery is the way to go to take care of sub-trees

src/nemos/base_regressor.py
BalzaniEdoardo: i'll look into this, we may be able to merge it but let me check it out. and thanks for the msing docs

wulfdewolf (approved):

PR #476, opened 2026-01-21
Add ClassifierGLM

BalzaniEdoardo:
# Summary

Adds ~~`CategoricalGLM` and `CategoricalPopulationGLM`~~ `ClassifierGLM` and `ClassifierPopulationGLM` for modeling categorical outcomes such as behavioral choices.

## New Classes

- **~~`CategoricalGLM`~~** **`ClassifierGLM`**: Fit classification models for single subjects
- **~~`CategoricalPopulationGLM`~~** **`ClassifierPopulationGLM`**: Fit multiple subjects in parallel

## API

Both classes follow the familiar GLM API with key differences:

| Method | Description |
|--------|-------------|
| `fit(X, y)` | Fit model where `y` contains integer category labels |
| `predict(X)` | Returns predicted category labels (not rates) |
| `predict_proba(X)` | Returns probabilities or log-probabilities for each category |
| `simulate(key, X)` | Generate synthetic categorical data |
| `score(X, y)` | Compute mean log-likelihood |
| `set_classes(y)`| Set the classes labels (default `[0,...,N-1]`) | 

### Parameter shapes

| Model | `coef_` | `intercept_` |
|-------|---------|--------------|
| ~~`CategoricalGLM`~~ | ~~`(n_features, n_categories - 1)`~~ | ~~`(n_categories - 1,)`~~|
| ~~`CategoricalPopulationGLM`~~ | ~~`(n_features, n_neurons, n_categories - 1)`~~| ~~`(n_neurons, n_categories - 1)`~~|
| `ClassifierGLM` | `(n_features, n_categories )` | `(n_categories,)`|
| `ClassifierPopulationGLM` | `(n_features, n_neurons, n_categories)` | `(n_neurons, n_categories)` |

## Implementation Details

- New validators: `ClassifierGLMValidator` and `PopulationClassifierGLMValidator` in `validation.py`
- Uses `log_softmax` as the default inverse link function
- Compatible with all existing regularizers (Ridge, Lasso, ElasticNet, GroupLasso)
- Integrated with pynapple for time series support
- Full scikit-learn compatibility for cross-validation and pipelines

## Documentation

- New how-to guide: `docs/how_to_guide/glm_for_classification.md`
- Added "Categorical GLM" section to quickstart
- Added to API reference

### DIFF
--- a/tests/test_glm.py
+++ b/tests/test_glm.py
@@ -3815,7 +3815,15 @@ class TestClassifierGLM:
 
     @pytest.mark.parametrize(
         "method_name",
-        ["predict_proba", "predict", "update", "compute_loss", "simulate"],
+        [
+            "predict_proba",
+            "predict",
+            "update",
+            "compute_loss",
+            "simulate",
+            "initialize_params",
+            "initialize_solver_and_state",
+        ],
     )
     def test_must_set_classes_before_calling(
         self,
@@ -3839,6 +3847,7 @@ class TestClassifierGLM:
             "random_key": None,
             "feedforward_input": None,
             "opt_state": None,
+            "init_params": None,
         }
         model.coef_ = true_params.coef
         model.intercept_ = true_params.intercept
@@ -4094,28 +4103,3 @@ class TestClassifierGLM:
         # Roundtrip: encode -> decode should give original labels
         y_label_roundtrip = model._decode_labels(model._encode_labels(y_label))
         assert np.array_equal(y_label, y_label_roundtrip)
-
-    def test_initialize_solver_sets_classes(
-        self, inv_link, glm_type, model_instantiation, request
-    ):
-        """Test that initialize_solver_and_state sets classes from y."""
-        X, y, model, _, _ = request.getfixturevalue(glm_type + model_instantiation)
-
-        # Create fresh model
-        if "population" in glm_type:
-            fresh_model = nmo.glm.ClassifierPopulationGLM(n_classes=model.n_classes)
-        else:
-            fresh_model = nmo.glm.ClassifierGLM(n_classes=model.n_classes)
-
-        assert fresh_model.classes_ is None
-
-        # Use string labels
-        label = np.array([chr(i) for i in range(ord("a"), ord("a") + model.n_classes)])
-        y_label = label[y]
-
-        init_params = fresh_model.initialize_params(X, y_label)
-        fresh_model.initialize_solver_and_state(X, y_label, init_params)
-
-        # Classes should now be set
-        assert fresh_model.classes_ is not None
-        assert np.array_equal(fresh_model.classes_, label)


sjvenditto:
A couple comments:
- Should we rename the class to something like `ClassifierGLM` or `ClassificationGLM`, and use parameters like `n_classes` (basically classes instead of categories)? I feel like the term "categorical" usually refers to the feature type or the distribution type, where the model itself is associated with classification. The parameter naming would also be more consistent with sklearn's `LogisticRegression`
- Why not use `n_categories` coefficients like sklearn? Personally, I feel like this is more useful because I would want to compare the coefficients of all the categories, and I'd want the model fit to give that to me instead of me having to extract the "implicit" category myself. Also, the per-category coefficients can be meaningful, e.g. this sklearn example with MNIST classification: https://scikit-learn.org/stable/auto_examples/linear_model/plot_sparse_logistic_regression_mnist.html
- If we really want to go the full sklearn route, we may also consider allowing other data types, e.g. strings, that we convert internally to integer labels. but `simulate` and `predict` would return the original label types, and we'd have the attribute `classes_` to list out the extracted class labels

BalzaniEdoardo:
> A couple comments:
> 
> * Should we rename the class to something like `ClassifierGLM` or `ClassificationGLM`, and use parameters like `n_classes` (basically classes instead of categories)? I feel like the term "categorical" usually refers to the feature type or the distribution type, where the model itself is associated with classification. The parameter naming would also be more consistent with sklearn's `LogisticRegression`

I agree with all these points, I'll make the changes tomorrow.

> * Why not use `n_categories` coefficients like sklearn? Personally, I feel like this is more useful because I would want to compare the coefficients of all the categories, and I'd want the model fit to give that to me instead of me having to extract the "implicit" category myself. Also, the per-category coefficients can be meaningful, e.g. this sklearn example with MNIST classification: https://scikit-learn.org/stable/auto_examples/linear_model/plot_sparse_logistic_regression_mnist.html

This doesn't play well with the unregularized version because to identify the model you need some sort of regulation. Otherwise you'll have K params but only K-1 degrees of freedom. 

> * If we really want to go the full sklearn route, we may also consider allowing other data types, e.g. strings, that we convert internally to integer labels. but `simulate` and `predict` would return the original label types, and we'd have the attribute `classes_` to list out the extracted class labels

That could be, not too hard to do that.

BalzaniEdoardo:
ok, for @sjvenditto point num 2, we can change the `from_model_param` function to convert from k-1 weghts to k weights, same thing with the `to_model_params` to do the reverse

gviejo (approved):

src/nemos/glm/validation.py
```diff
@@ -372,3 +387,272 @@ def get_empty_params(self, X, y) -> GLMParams:
...
+
+    Validates and transforms user-provided parameters, inputs, and checks consistency
+    between parameters and data for classifier GLMs. Classifier GLMs have:
+    - 2D coefficients: shape (n_features, n_classes - 1) or dict of (n_features, n_classes - 1) arrays
+    - 1D intercept: shape (n_classes - 1,)
+    - 2D input X: shape (n_samples, n_features) or pytree of same
+    - 1D output y: shape (n_samples,) containing integer class labels
+    """
+
+    extra_params: ClassifierExtraParams = field(kw_only=True)
```
sjvenditto: why does this need to be a class? why can't it just be a dictionary defined here?

src/nemos/glm/validation.py
```diff
@@ -372,3 +387,272 @@ def get_empty_params(self, X, y) -> GLMParams:
...
+
+
+@dataclass(frozen=True, repr=False)
+class ClassifierGLMValidator(GLMValidator):
+    """
+    Validator for classifier GLM models.
+
+    Validates and transforms user-provided parameters, inputs, and checks consistency
+    between parameters and data for classifier GLMs. Classifier GLMs have:
+    - 2D coefficients: shape (n_features, n_classes - 1) or dict of (n_features, n_classes - 1) arrays
```
sjvenditto: should this be `n_classes`? not `n_classes-1`?

src/nemos/glm/validation.py
```diff
@@ -372,3 +387,272 @@ def get_empty_params(self, X, y) -> GLMParams:
...
+
+
+@dataclass(frozen=True, repr=False)
+class PopulationClassifierGLMValidator(ClassifierGLMValidator):
+    """
+    Validator for population (multi-neuron) classifier GLM models.
+
+    Validates and transforms user-provided parameters, inputs, and checks consistency
+    between parameters and data for population classifier GLMs. Population classifier GLMs have:
+    - 3D coefficients: shape (n_features, n_neurons, n_classes - 1) or dict of same
```
sjvenditto: should this also be `n_classes`?

docs/how_to_guide/glm_for_classification.md
```diff
@@ -0,0 +1,100 @@
...
+    jupytext_version: 1.18.1
+kernelspec:
+  display_name: Python 3 (ipykernel)
+  language: python
+  name: python3
+---
+
+# Fit GLM for Classification
+
+The [`ClassifierGLM`](nemos.glm.ClassifierGLM) models categorical outcomes such as behavioral choices.
```
sjvenditto: I think we need to emphasize more strongly that the "Categorical" observation model is the only one compatible with this GLM (or "Categorical" + "Bernoulli" if we go that route). Alongside this description, we can have a bullet point in the key differences from GLM highlighting it, e.g. that observation model is limited / should not be set

docs/quickstart.md
```diff
@@ -141,6 +141,57 @@ print(f"Model intercept shape: {population_model.intercept_.shape}")
...
+
+import nemos as nmo
+
+# Instantiate a categorical model for 3 categories
+n_classes = 3
+model = nmo.glm.ClassifierGLM(n_classes)
+
+```
+
+The `ClassifierGLM` follows a similar API to the standard GLM, with a few key differences:
```
sjvenditto: Same comment as above: I think we need to emphasize more strongly that the "Categorical" observation model is the only one compatible with this GLM (or "Categorical" + "Bernoulli" if we go that route). Alongside this description, we can have a bullet point in the key differences from GLM highlighting it, e.g. that observation model is limited / should not be set

docs/how_to_guide/glm_for_classification.md
billbrod: that last bullet point should probably revolve around the fact that the ClassifierGLM does not have an observation model parameter in its init method

docs/how_to_guide/glm_for_classification.md
```diff
@@ -0,0 +1,100 @@
...
+
+**Key differences from standard GLM:**
+- [`predict`](nemos.glm.ClassifierGLM.predict) returns predicted category labels
+- [`predict_proba`](nemos.glm.ClassifierGLM.predict_proba) returns (log-)probabilities for each category
+- [`set_classes`](nemos.glm.ClassifierGLM.set_classes) set class labels. By Default, classes are assumed to be `{0, ..., n_classes - 1}` however, this assumption can be overridden by calling `set_classes` passing an array of class labels.
+
+## Generate Synthetic Data
+
```
billbrod, line 25: I think there should be some text here highlighting that this is simulated (and with real data, this section would be replaced by loading in the data), because that might be easy to miss.

docs/quickstart.md
```diff
@@ -141,6 +141,57 @@ print(f"Model intercept shape: {population_model.intercept_.shape}")
 
 ```
 
+### **Categorical GLM**
```
billbrod: should this section be called classifier glm to be consistent with the class name?

docs/quickstart.md
```diff
@@ -141,6 +141,57 @@ print(f"Model intercept shape: {population_model.intercept_.shape}")
...
+
+# Instantiate a categorical model for 3 categories
+n_classes = 3
+model = nmo.glm.ClassifierGLM(n_classes)
+
+```
+
+The `ClassifierGLM` follows a similar API to the standard GLM, with a few key differences:
+
+- `predict` returns predicted category labels (not rates)
```
billbrod:
```suggestion
- `predict` returns predicted category labels (unlike the standard `GLM` model, which returns predicted firing rates).
```

with link to the GLM's predict method probably

docs/quickstart.md
```diff
@@ -141,6 +141,57 @@ print(f"Model intercept shape: {population_model.intercept_.shape}")
...
+- `predict_proba` returns the probability of each category
+
+```{code-cell}
+
+import numpy as np
+num_samples, num_features = 100, 3
+
+# Generate a design matrix
+X = np.random.normal(size=(num_samples, num_features))
+# generate categorical observations (values 0, 1, or 2)
```
billbrod:
```suggestion
# simulate categorical observations (values 0, 1, or 2)
```

src/nemos/glm/glm.py
```diff
@@ -227,6 +249,33 @@ def __init__(
         self.aux_ = None
         self.optim_info_ = None
 
+    @classmethod
+    def _validate_observation_class(cls, observation: obs.Observations):
```
billbrod, line 250: I feel like this GLM's validate method should assume the model's non-categorical, and the subclass should check for categorical ones

src/nemos/glm/glm.py
```diff
@@ -1461,3 +1496,1015 @@ def __sklearn_clone__(self) -> PopulationGLM:
...
+            i.e. ``len(np.unique(y)) == n_classes``.
+
+        Notes
+        -----
+        :meth:`fit` and :meth:`initialize_solver_and_state` call ``set_classes`` internally,
+        making sure that the ``classes_`` attribute matches the provided input.
+        If you are fitting in batches by calling :meth:`update`, make sure that the ``classes_``
+        are correctly set by calling ``set_classes`` before starting the :meth:`update` loop.
+
+        Examples
```
billbrod: should probably show this being used with update, which is the only reason a user would call it, right?

billbrod: and probably some of the failures

src/nemos/glm/glm.py
```diff
@@ -1461,3 +1496,1015 @@ def __sklearn_clone__(self) -> PopulationGLM:
...
+    # observation model inferred
+    _invalid_observation_types = ()
+
+    def _check_classes_is_set(self, method_name, y=None):
+        if self._classes_ is None:
+            raise RuntimeError(
+                f"Classes are not set. Must call ``set_classes`` before calling ``{method_name}``."
+            )
+
+    def set_classes(self, y: ArrayLike) -> ClassifierMixin:
```
billbrod: should document the errors raised

src/nemos/glm/glm.py
```diff
@@ -1461,3 +1496,1015 @@ def __sklearn_clone__(self) -> PopulationGLM:
...
+        params :
+            The current model parameters, typically a tuple of coefficients and intercepts.
+        opt_state :
+            The current state of the optimizer, encapsulating information necessary for the
+            optimization algorithm to continue from the current state.
+        X :
+            The predictors used in the model fitting process. Shape ``(n_time_bins, n_features)``
+            or a ``FeaturePytree``.
+        y :
+            Class labels as integers, array of shape ``(n_time_bins,)`` for single neuron
```
billbrod: I think this is incorrect -- we allow any dtype as labels right? so not just float but also strings? (that's what the call to _encode_labels is doing, yeah)?

src/nemos/glm/glm.py
```diff
@@ -1461,3 +1496,1015 @@ def __sklearn_clone__(self) -> PopulationGLM:
...
+            observation_model=observation_model,
+            inverse_link_function=inverse_link_function,
+            regularizer=regularizer,
+            regularizer_strength=regularizer_strength,
+            solver_name=solver_name,
+            solver_kwargs=solver_kwargs,
+        )
+        self._classes_ = None
+        self._class_to_index_ = None
+        self._skip_encoding = True  # default: assume labels are [0, n_classes-1]
```
billbrod: this feels like a dangerous default -- shouldn't we default to being safer at the expense of performance?

src/nemos/glm/validation.py
```diff
@@ -206,46 +214,53 @@ def validate_consistency(
...
-        it to JAX arrays with the specified data type. Subclasses can extend
+        it to JAX arrays with the specified data type. Handles FeaturePytree
+        inputs by extracting the underlying data. Subclasses can extend
         this to add parameter-specific validation (e.g., checking that mask
         shape matches parameter dimensions).
 
         Parameters
         ----------
-        feature_mask : dict[str, jnp.ndarray] or jnp.ndarray
+        feature_mask : FeaturePytree, dict[str, jnp.ndarray], jnp.ndarray, or None
```
billbrod: isn't annotating the parameters here redundant with the type hinting?

src/nemos/glm/glm.py
BalzaniEdoardo: switched to false

src/nemos/glm/glm.py
BalzaniEdoardo: yep, i changed the description

Issue #377, opened 2025-07-22
Option to specify optimization backend


bagibence:
While NeMoS is [moving away from JAXopt as its optimization backend](https://github.com/flatironinstitute/nemos/pull/365), the solver interface is compatible with both JAXopt and Optimistix.

At first JAXopt will be kept as the default, but there could be a way for users to specify which optimization backend they want to use. (Currently this can be done in https://github.com/flatironinstitute/nemos/pull/365 by overwriting the solver registry, but it's hidden from users.)

There are multiple ways this could be done:
- By setting an environment variable such as `NEMOS_SOLVER_BACKEND=optimistix` like it's currently done in the tests.
- Giving the option to specify the backend for each algorithm.
Notation for this could `GLM(solver_name="LBFGS[jaxopt]")` or `GLM(solver_name="JaxoptLBFGS")`

PR #479, opened 2026-01-27
Add support for custom solvers

### LINKED ISSUE #378: Allow custom solvers
Currently solvers are created based on the algorithm's name, which for most users is fine and allows checking for compatibility with the regularizer.

However, it could be allowed for user-defined solvers to be passed to `BaseRegressor` instead of a string.
As long as a solver implements the `AbstractSolver` interface, it should be compatible with NeMoS.

In this case the compatibility check with the regularizer could be disabled or rewritten, which should be fine since only advanced users would need this option.

bagibence:
## Purpose
This PR adds support for user-defined custom solvers to be used for parameter fitting in NeMoS.

This change is targeted at advanced users wanting to test solvers and trying to improve performance, and developers of new models requiring custom solvers not included in NeMoS.
For the average user there is no change in use.

Fixes #378 and #377

## Main changes
- The solver registry now uses two dictionaries
  ```python
  # mapping is {algo_name : {backend : implementation}}
  _registry: dict[str, dict[str, SolverSpec]] = {}
  # mapping is {algo_name : backend}
  _defaults: dict[str, str] = {}
  ```
  where `SolverSpec` is just a small dataclass for storing the implemented algorithm name, an arbitrary backend name, and the class implementing the solver.
  The main functions for interacting with the registry are `get_solver` and `register`.
- Custom solvers have to be registered and the algorithm they implement has to be allowed by the regularizer to be used.
  A how-to guide on writing a custom solver using `scipy` as a backend and using it with NeMoS is added.
- Functions for validating solver compatibility, i.e. that it adheres to the `AbstractSolver` interface, are added.
  When registering a new solver, the implementation is checked against `SolverProtocol` and the signatures of its methods are also checked. Optionally, the solver can be applied to a small ridge regression problem to test functionality.
- `BaseRegressor.__init__` still takes `solver_name`. The setter of `solver_name` sets the `_solver_spec` attribute for storing the solver implementation. `solver_name` is returned in the format _algo_name[backend_name]_
- As `solver_name` remains a string, no changes to saving / loading logic. Custom solvers have to be registered before loading models that used them.

## Questions remaining
- `SolverProtocol` is a bit redundant for validation because `validate_solver_class` checks the existence of the methods and more. But it's nice for typing to clearly signal that solver classes should implement the protocol.

## Alternative implementation considered
I [implemented an alternative](https://github.com/bagibence/nemos/tree/custom_solver_flexi) where a custom solver does not necessarily have to be registered before use, the implementation class or a `SolverSpec` can be directly passed to the model (e.g. `GLM(solver=MyAdamImplementation)`).
While this saves advanced users 2 lines of code for using a custom solver, the main downside is making the codebase more complex for the convenience of a small minority of users. :
- Backward compatibility is with changing from `solver_name` to `solver`. `solver` is also easy to confuse with the current `_solver`, so that is changed to `_solver_instance`.
- `BaseRegressor`'s code is more complex.
- Regularizer compatibility check is disabled for non-registered algorithms, allowing less mindful and potential misuse of algorithms
- For saving and loading extra serialization logic is added.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -60,6 +60,7 @@ jobs:
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
         python-version: ['3.10', '3.11', '3.12', '3.13']
+      fail-fast: false
     runs-on: ${{ matrix.os }}
     steps:
       - name: Checkout code
@@ -91,6 +92,7 @@ jobs:
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
         python-version: ['3.10', '3.11', '3.12', '3.13']
+      fail-fast: false
     runs-on: ${{ matrix.os }}
     steps:
       - name: Checkout code


docs/developers_notes/07-solvers.md
```diff
@@ -96,13 +97,28 @@ Note that what Optax calls Nesterov acceleration is not the [original method dev
...
+`nemos.solvers.register("Fancy-Algorithm", MyCustomSolverClass, backend="custom")`\
+Please note that not a solver instance but a class/type has to be passed.
+2. Declaring the algorithm's compatibility with the appropriate regularizers: \
+`nemos.regularizer.UnRegularized.allow_solver("Fancy-Algorithm")`.
+3. Referring to the algorithm by name when creating a `GLM` (or any `BaseRegressor`): \
+`GLM(solver_name="Fancy-Algorithm[custom]")`
+
+When registering a solver, NeMoS does basic checks validating the custom solver's compatibility by checking if the required methods are implemented, i.e. if the class implements the  and that their signatures match [`AbstractSolver`](nemos.solvers.AbstractSolver) (and [`SolverProtocol`](nemos.solvers.SolverProtocol)).
+There are also options in `nemos.solvers.register` to run a small ridge regression problem, testing that the solver's methods can be used as intended.
+To validate a solver without registering, the `nemos.solvers.validate_solver_class` can be used.
```
BalzaniEdoardo: can you add a ref to this? [validate_solver_class](nemos.solvers.validate_solver_class)

docs/how_to_guide/custom_solvers.md
```diff
@@ -0,0 +1,418 @@
...
+To support flexibility and long-term maintenance, NeMoS now has a backend-agnostic solver interface, allowing the use of solvers from different backend libraries with different interfaces.  
+This also means that users can provide their own solvers, and as long as they adhere to the interface defined by `AbstractSolver`, they should be compatible with NeMoS and can be used for fitting models.
+
+In the following we will walk through how one can create a NeMoS-compatible solver that uses [`scipy.optimize.minimize` with the modified Powell method](https://docs.scipy.org/doc/scipy/reference/optimize.minimize-powell.html#optimize-minimize-powell), an algorithm not present in NeMoS by default.
+
++++
+
+## Define the `scipy` adapter
+
+In order to adhere to the `AbstractSolver` interface, we have to define the following methods in our custom solver class:
```
BalzaniEdoardo, line 25: do you think this scipy adapter should be directly in the package? 
Also, I rechecked our flattener/unflattner, we are using `.flatten` but maybe we should do`.reshape(-1)` which does not copy the data if possible.

BalzaniEdoardo: It could be the subject of a dedicated PR, since it would require some minimal testing. I would not test the convergence, but just that that it can run with different tree structs

BalzaniEdoardo: If we go this way, this example could become only about registering a new solver using the scipy adapter.

src/nemos/solvers/_solver_registry.py
```diff
@@ -1,59 +1,287 @@
...
-        JaxoptNonlinearCG,
-        JaxoptProximalGradient,
-    )
+@dataclass
+class SolverSpec:
+    """
+    Solver specification representing an entry in the solver registry.
+
+    A solver is specified by:
+    - name of the algorithm it implements
```
BalzaniEdoardo:
```suggestion
    - the name of the algorithm it implements
```

src/nemos/solvers/_solver_registry.py
```diff
@@ -1,59 +1,287 @@
...
+    implementation: Type[SolverProtocol]
 
+    @property
+    def full_name(self) -> str:
+        return f"{self.algo_name}[{self.backend}]"
 
-def list_available_solvers():
+    def __repr__(self) -> str:
+        return (
+            f"{self.full_name!r} - "
```
BalzaniEdoardo: maybe make the spec repr multiine? 
```python
"Algoname[backend]" - SolverSpec(
    algo_name="AlgoName"
    ...
) 
```

src/nemos/solvers/_solver_registry.py
```diff
@@ -1,59 +1,287 @@
...
+@dataclass
+class SolverSpec:
+    """
+    Solver specification representing an entry in the solver registry.
+
+    A solver is specified by:
+    - name of the algorithm it implements
+    - its backend (optimization library or custom)
+    - the class implementing the optimization method
+      (ideally compatible with the AbstractSolver and SolverProtocol interface)
```
BalzaniEdoardo, line 27: Can you add an example section here, that basically defines the spec for one of the solver we have

src/nemos/solvers/_solver_registry.py
```diff
@@ -1,59 +1,287 @@
...
+    replace: bool = False,
+    default: bool = False,
+    validate: bool = True,
+    test_ridge_without_aux: bool = False,
+    test_ridge_with_aux: bool = False,
+) -> None:
+    """
+    Register a solver implementation in the registry.
+
+    Parameters
```
BalzaniEdoardo, line 207: This docstring should have an example. up to you if you want to mock the solverprotocol or register one of our solvers with another name

src/nemos/solvers/_solver_registry.py
```diff
@@ -1,59 +1,287 @@
...
+
+    if default:
+        set_default(algo_name, backend)
+
+
+def set_default(algo_name: str, backend: str) -> None:
+    """
+    Set the default backend for a given algorithm.
+
+    Parameters
```
BalzaniEdoardo, line 268: you can show how to set optimistix lbfgs backend.

src/nemos/solvers/_solver_registry.py
```diff
@@ -1,59 +1,287 @@
...
-    return list(solver_registry.keys())
+    >>> nmo.solvers.list_available_algorithms()
+    ['GradientDescent', 'ProximalGradient', 'LBFGS', 'BFGS', 'NonlinearCG', 'SVRG', 'ProxSVRG']
+    """
+    return list(_registry.keys())
+
+
+register("GradientDescent", OptimistixNAG, "optimistix", default=True)
+register("ProximalGradient", OptimistixFISTA, "optimistix", default=True)
+register("LBFGS", OptimistixOptaxLBFGS, "optax+optimistix", default=True)
```
BalzaniEdoardo, line 335: I coded up the LBFGS in optimistix and we don't even provide it as an available option?  sad :)

src/nemos/solvers/_validation.py
```diff
@@ -0,0 +1,255 @@
...
+# Notes
+# We could enforce adherence to the API with type checkers
+# https://github.com/agronholm/typeguard
+# https://github.com/beartype/beartype
+
+METHOD_NAMES = AbstractSolver.__abstractmethods__
+AUX_VAL = -1
+
+
+def _get_params(
```
BalzaniEdoardo: this should be moved to the `_inspect_utils` module

src/nemos/solvers/_validation.py
```diff
@@ -0,0 +1,255 @@
...
+
+    # raise one error with all the problems found
+    if not all(success_dict.values()):
+        error_msg = "\n".join(
+            problem_dict[method] for method in METHOD_NAMES if not success_dict[method]
+        )
+        raise ValueError(error_msg)
+
+
+def _check_required_methods_exist(solver_class: Type):
```
BalzaniEdoardo, line 77: add an argument (`method_names: List[str]`) and move it to `_inspect_utils`, then here just do a closure:

```python
from .._inspect_utils import implements_abstract_methods  # this will be implementing the logic.

def _check_required_methods_exist(solver_class):
    return  implements_abstract_methods(solver_class, METHOD_NAMES)
```

bagibence: I moved it and called the function `implements_methods` because they don't have to be abstract.

src/nemos/solvers/_validation.py
bagibence: Moved it.

src/nemos/solvers/_solver_registry.py
bagibence: Showed it with the current `"optax+optimistix`

src/nemos/solvers/_solver_registry.py
bagibence: Added an example registering `OptimistixFISTA` with the name FISTA.

src/nemos/solvers/_solver_registry.py
bagibence: Added an example creating and inspecting a `SolverSpec` for BFGS.

src/nemos/solvers/_solver_registry.py
bagibence: Added.

docs/developers_notes/07-solvers.md
bagibence: I added links to `register` and `validate_solver_class`. Is that what you meant?

src/nemos/solvers/_solver_registry.py
bagibence: It is, but I don't think it should be included until it uses backtracking and not zoom linesearch as recommended. Optimistix's backtracking can run forever. BFGS is already shaky, but that's used less often and there's no Optax implementation.

docs/how_to_guide/custom_solvers.md
bagibence: I'm not sure if the scipy adapter should be included in the package. It's not JAX-based and the update method is a bit hacky.
I also think that having the full definition of a custom solver is good to have in the how-to guide -- as opposed to just importing a wrapper without seeing its source code.
That said, it would allow the use of many other solvers.

bagibence:
> This looks good to me, but there are a bunch of tests missing

You're right, they were missing. I added them in `test_solver_registry.py`.

> I think I can just share the token

I added the codecov token. Hopefully that helps.

I also made some fixes and cleanups, and resolved conflicts with development.

src/nemos/solvers/_solver_registry.py
bagibence: Okay. Changed to this format.

docs/developers_notes/07-solvers.md
BalzaniEdoardo: yes

src/nemos/solvers/_solver_registry.py
BalzaniEdoardo: ok, sounds good! we need to push for the zoom search

src/nemos/solvers/_validation.py
BalzaniEdoardo: great

PR #480, opened 2026-01-28
Use uv in the CI

### LINKED ISSUE #437: Switch CI set up to uv
Install the package via uv in the CI, to make the installation faster,

bagibence:
A small PR to speed up `tox` by using `uv`, most importantly speeding up the CI a bit.

Changes are:
- Add `setup-uv` step, then use `uvx` to run `tox` with the [`tox-uv` plugin](https://github.com/tox-dev/tox-uv) that uses `uv` instead of `virtualenv` and `pip` inside `tox`.
- Add `tox-uv` as a dev dependency so that it's used locally, too.
- Update `setup-python` version to 6.
- Export coverage report on solver-related tests.
- Speed up `fix` and `check` `tox` environments by not installing `nemos`, only the required dependencies.

Note that the setup could be further simplified by removing the `setup-python` step and using `uv` to set up Python.

Fixes #437

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -40,7 +40,7 @@ setenv =
     NEMOS_SOLVER_BACKEND = jaxopt
 commands =
     pytest -n auto --doctest-modules src/nemos/solvers
-    pytest -n auto -m solver_related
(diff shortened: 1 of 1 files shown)


BalzaniEdoardo (approved):
Nice, thanks

PR #500, opened 2026-02-20
Lazy imports

BalzaniEdoardo:
# PR: Lazy Loading for Submodules

## Summary

Replace eager imports with lazy loading using `lazy_loader`, dramatically reducing initial import time and deferring heavy dependencies until they're actually needed.

## Motivation

`import nemos` previously took **~3 seconds** because it eagerly loaded all submodules and their dependencies (jax, scipy, sklearn, pynapple, etc.) at import time. This created unnecessary friction for:

- Scripts that only use a subset of nemos functionality
- CLI tools and quick experiments
- Notebooks where startup time matters
- IDE/linter startup when nemos is in the environment

## Changes

- Replace explicit imports in `__init__.py` with `lazy_loader.attach_stub()`
- Add `__init__.pyi` stub file defining the public API (preserves IDE autocompletion)
- Defer scipy imports in specific modules

### Base Import Time

| Branch | `import nemos` |
|--------|----------------|
| main | 3145ms |
| lazy-imports | **2.8ms** |

**~83x faster initial import**

### Per-Submodule Import Time (fresh process each)

| Submodule | main | lazy-imports | Speedup |
|-----------|------|--------------|---------|
| basis | 2183ms | 1167ms | 1.9x |
| glm | 2235ms | 1340ms | 1.7x |
| observation_models | 2192ms | 1090ms | 2.0x |
| regularizer | 2181ms | 1134ms | 1.9x |
| solvers | 2227ms | 484ms | 4.6x |
| convolve | 2191ms | 272ms | 8.1x |
| identifiability_constraints | 2192ms | 269ms | 8.1x |
| pytrees | 2186ms | 252ms | 8.7x |
| simulation | 2189ms | 262ms | 8.4x |
| tree_utils | 2185ms | 255ms | 8.6x |
| type_casting | 2186ms | 261ms | 8.4x |
| utils | 2172ms | 268ms | 8.1x |
| fetch | 2180ms | 63ms | 34.6x |
| exceptions | 2181ms | 33ms | 66.1x |
| styles | 2189ms | 32ms | 68.4x |

### Dependencies After `import nemos`

| Dependency | main | lazy-imports |
|------------|------|--------------|
| jax | LOADED | deferred |
| scipy | LOADED | deferred |
| sklearn | LOADED | deferred |
| pynapple | LOADED | deferred |
| matplotlib | LOADED | deferred |
| pandas | LOADED | deferred |
| h5py | LOADED | deferred |
| pynwb | LOADED | deferred |

## Implementation

Uses [`lazy_loader`](https://github.com/scientific-python/lazy_loader), following [SPEC-0001](https://scientific-python.org/specs/spec-0001/) — the Scientific Python specification for lazy loading adopted by numpy, scipy, scikit-image, and other scientific Python packages.

```python
# Before (eager)
from . import basis, glm, ...

# After (lazy)
import lazy_loader as _lazy
__getattr__, __dir__, __all__ = _lazy.attach_stub(__name__, __file__)
```

The `.pyi` stub file defines what's available for IDE autocompletion and type checking.

### `lazy.load()` for External Dependencies

In addition to `attach_stub()` for submodules, we use `lazy.load()` to defer imports of external libraries:

```python
import lazy_loader as lazy
scipy = lazy.load("scipy")
```

This creates a lazy proxy object — the real module only loads when you first access an attribute (e.g., `scipy.stats.gamma.pdf(...)`). We use this for:

- **Optional dependencies** (`fsspec`, `h5py`, `pynwb` in `fetch/fetch_data.py`) — only loaded if the user calls fetch functions
- **Heavy imports not always needed** (`scipy` in `simulation.py`, `pynapple` in `type_casting.py`)

### `TYPE_CHECKING` Blocks

To support type hints without runtime import costs, we use `TYPE_CHECKING` guards:

```python
from typing import TYPE_CHECKING

if TYPE_CHECKING:
    from pynapple import Tsd, TsdFrame
```

`TYPE_CHECKING` is `False` at runtime but `True` when type checkers (mypy, pyright) analyze the code. This means:
- **At runtime:** the import is skipped (fast)
- **During type checking:** the types are available for annotations (correct types)

This pattern is used throughout for pynapple types, internal module types (`Basis`, `Regularizer`), and other imports needed only for type hints.

### `from __future__ import annotations`

Several modules add this import to enable [PEP 563](https://peps.python.org/pep-0563/) (postponed evaluation of annotations). This prevents annotation expressions from being evaluated at import time, complementing the `TYPE_CHECKING` pattern — without it, annotations like `def foo(x: Tsd) -> TsdFrame:` would require those types to be imported at runtime.

## ~~Design Decision: Root-Level Stub Only~~

~~We intentionally placed the `.pyi` stub only at the package root (`src/nemos/__init__.pyi`), not in each submodule.~~

~~**Rationale:**~~

~~1. **Reduced maintenance burden** — Only one stub file to keep in sync with exports, rather than one per submodule.~~

~~2. **Matches the lazy loading granularity** — Lazy loading operates at the submodule level. Once you access `nemos.basis`, the entire basis module loads eagerly (all its classes, functions, etc.). Adding per-submodule stubs wouldn't provide additional lazy loading benefit.~~

~~3. **Package as a monolith** — Nemos is consumed as a cohesive package. Users typically access it via `nemos.basis.BSplineEval`, `nemos.glm.GLM`, etc. The submodules themselves don't need internal lazy loading.~~

~~4. **Follows ecosystem patterns** — This matches how numpy, scipy, and similar packages handle lazy loading: lazy at the top level, eager within submodules.~~

~~**Trade-off:** If a user does `from nemos.basis import BSplineEval`, they still get IDE support because the submodule itself has normal type hints. The root stub handles the `nemos.<submodule>` access pattern.~~

## Updated Design: Lazy Loading at Multiple Levels

Lazy loading is implemented at both the package root and within submodules:

- **Package root** (`src/nemos/__init__.py` + `__init__.pyi`) — `import nemos` loads nothing; submodules like `nemos.basis`, `nemos.glm` are deferred
- **Submodules** (`src/nemos/basis/__init__.py` + `__init__.pyi`, etc.) — individual classes/functions within submodules are also lazy-loaded

This means a user who only needs `nemos.basis.BSplineEval` won't trigger imports of `optimistix`, `optax`, or other dependencies used only by `nemos.glm` or `nemos.solvers`.

Each submodule with lazy loading has its own `.pyi` stub file to preserve IDE autocompletion and type checking support.

## What's NOT Changing

- **Full API preserved** — All the same modules and functions are available
- **IDE support** — The `.pyi` stub provides autocompletion and type hints
- **Runtime behavior** — Once loaded, everything works identically

## Considerations

| Consideration | Details |
|---------------|---------|
| **Deferred errors** | Import errors in a submodule surface at first use rather than at `import nemos`. |
| **First-use latency** | The load time still occurs, just deferred to first access. |
| **Stub maintenance** | `__init__.pyi` must stay in sync with actual exports when adding new submodules. |

### The ~1 Second Floor

Computational submodules (`basis`, `glm`, `observation_models`, `regularizer`) still take ~1-1.3 seconds to load. This is the cost of importing jax and sklearn, which are used pervasively:

| Submodule | Load Time | Why |
|-----------|-----------|-----|
| `styles`, `exceptions` | ~32ms | No jax/sklearn dependency |
| `fetch` | ~63ms | Light dependencies |
| `basis`, `glm`, etc. | ~1000-1300ms | Requires jax + sklearn |

### Why We Can't Easily Go Further

1. **`@jax.jit` decorators** — These are evaluated at module import time, not at function call time. Deferring them would require replacing with a lazy wrapper pattern.

2. **`jax.numpy` usage** — Core array operations throughout the codebase require jax to be loaded.

3. **`BaseEstimator` inheritance** — sklearn compatibility requires sklearn at class definition time.

Making jax/sklearn lazy *within* modules would require significant architectural changes (lazy decorators, deferred class inheritance) with diminishing returns and added complexity.

### Why Computational Modules Show ~2x Speedup

The speedup from ~2200ms to ~1100ms for modules like `basis` and `glm` is **not** from optimizations within those modules. It's because they no longer eagerly load `fetch` and its heavy dependencies (dandi, h5py, pynwb).

Previously on main:
```
import nemos  →  loads ALL submodules  →  includes fetch  →  loads dandi, h5py, pynwb
```

Now with lazy loading:
```
import nemos        →  loads nothing
nemos.basis         →  loads jax, sklearn (no fetch deps)
nemos.fetch         →  loads dandi, h5py, pynwb (only if explicitly accessed)
```

### What This PR Achieves

This PR captures the **easy wins**:
- 83x faster `import nemos` (3145ms → 38ms)
- ~2x faster computational submodules by not loading `fetch` deps alongside them
- Users who only need lightweight submodules (`styles`, `exceptions`) avoid loading jax entirely
- `fetch` dependencies (dandi, h5py, pynwb) only load when actually fetching data

The remaining ~1s per computational submodule is the unavoidable cost of jax/sklearn initialization.

### DIFF
--- a/src/nemos/basis/_basis_mixin.py
+++ b/src/nemos/basis/_basis_mixin.py
@@ -552,7 +552,7 @@ class EvalBasisMixin:
         self.set_input_shape(*xi)
         return self
 
-    def _set_input_independent_states(self) -> "EvalBasisMixin":
+    def _set_input_independent_states(self) -> EvalBasisMixin:
         """
         Compute all the basis states that do not depend on the input.
 
@@ -690,7 +690,7 @@ class ConvBasisMixin:
         """
         return self._set_kernel()
 
-    def _set_kernel(self) -> "ConvBasisMixin":
+    def _set_kernel(self) -> ConvBasisMixin:
         """
         Prepare or compute the convolutional kernel for the basis functions.
 
--- a/src/nemos/basis/_check_basis.py
+++ b/src/nemos/basis/_check_basis.py
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
 from typing import TYPE_CHECKING, Tuple
 
 import jax.numpy as jnp
@@ -24,7 +26,7 @@ def _check_zero_samples(
         raise ValueError(err_message)
 
 
-def _check_input_dimensionality(bas: "BasisMixin | Basis", xi: Tuple) -> None:
+def _check_input_dimensionality(bas: BasisMixin | Basis, xi: Tuple) -> None:
     """
     Check that the number of inputs provided by the user matches the number of inputs required.
 
@@ -66,7 +68,7 @@ def _check_samples_consistency(*xi: NDArray) -> None:
 
 
 def _check_transform_input(
-    bas: "BasisMixin | Basis",
+    bas: BasisMixin | Basis,
     *xi: ArrayLike,
 ) -> Tuple[NDArray]:
     # check dimensionality
--- a/src/nemos/basis/_composition_utils.py
+++ b/src/nemos/basis/_composition_utils.py
@@ -5,6 +5,8 @@ Collection of functions that transverse the composite basis tree
 with no to minimal re
 """
 
+from __future__ import annotations
+
 import re
 from copy import deepcopy
 from functools import wraps
@@ -57,7 +59,7 @@ def add_docstring(method_name, cls):
     return wrapper
 
 
-def _iterate_over_components(basis: "BasisMixin"):
+def _iterate_over_components(basis: BasisMixin):
     components = (
         basis._iterate_over_components()
         if hasattr(basis, "_iterate_over_components")
@@ -66,7 +68,7 @@ def _iterate_over_components(basis: "BasisMixin"):
     yield from components
 
 
-def _get_root(bas: "BasisMixin"):
+def _get_root(bas: BasisMixin):
     """Get the basis root."""
 
     parent = bas
@@ -75,7 +77,7 @@ def _get_root(bas: "BasisMixin"):
     return parent
 
 
-def _has_default_label(bas: "BasisMixin"):
+def _has_default_label(bas: BasisMixin):
     """Check for default label.
 
     The check either use the property (if it is a nemos basis), or the class name
@@ -89,7 +91,7 @@ def _has_default_label(bas: "BasisMixin"):
 
 
 def _recompute_class_default_labels(
-    bas: "BasisMixin",
+    bas: BasisMixin,
 ):
     """
     Recompute all labels matching default for self.
@@ -111,7 +113,7 @@ def _recompute_class_default_labels(
             bas_id += 1
 
 
-def _recompute_all_default_labels(root: "BasisMixin") -> "BasisMixin":
+def _recompute_all_default_labels(root: BasisMixin) -> BasisMixin:
     """Recompute default all labels."""
     updated = []
     for bas in _iterate_over_components(root):
@@ -121,7 +123,7 @@ def _recompute_all_default_labels(root: "BasisMixin") -> "BasisMixin":
     return root
 
 
-def _update_label_from_root(bas: "BasisMixin", class_name: str, class_label: str):
+def _update_label_from_root(bas: BasisMixin, class_name: str, class_label: str):
     """
     Subtract 1 to each matching default label with higher ID then current.
 
@@ -148,7 +150,7 @@ def _update_label_from_root(bas: "BasisMixin", class_name: str, class_label: str
             bas._label = f"{class_name}_{bas_id}" if bas_id else class_name
 
 
-def _composite_basis_setter_logic(new: "BasisMixin", current: "BasisMixin"):
+def _composite_basis_setter_logic(new: BasisMixin, current: BasisMixin):
     """Setter logic for composite basis."""
     # Carry-on label if possible
     if _has_default_label(new) and not _has_default_label(current):
@@ -170,7 +172,7 @@ def _composite_basis_setter_logic(new: "BasisMixin", current: "BasisMixin"):
 
 
 def _atomic_basis_label_setter_logic(
-    bas: "AtomicBasisMixin | CustomBasis", new_label: str
+    bas: AtomicBasisMixin | CustomBasis, new_label: str
 ) -> Exception | None:
     """Setter logic for atomic basis."""
     # check default cases
@@ -221,7 +223,7 @@ def _atomic_basis_label_setter_logic(
     return
 
 
-def infer_input_dimensionality(bas: "BasisMixin") -> int:
+def infer_input_dimensionality(bas: BasisMixin) -> int:
     """Infer input dimensionality from compute_features signature.
 
     If `_n_input_dimensionality` return the attribute, otherwise return
@@ -246,7 +248,7 @@ def infer_input_dimensionality(bas: "BasisMixin") -> int:
     return n_input_dim
 
 
-def generate_basis_label_pair(bas: "BasisMixin"):
+def generate_basis_label_pair(bas: BasisMixin):
     """Generate all labels and basis in a basis tree."""
     if hasattr(bas, "basis1") and hasattr(bas, "basis2"):
         for label, sub_bas in generate_basis_label_pair(bas.basis1):
@@ -256,7 +258,7 @@ def generate_basis_label_pair(bas: "BasisMixin"):
     yield getattr(bas, "label", bas.__class__.__name__), bas
 
 
-def generate_composite_basis_labels(bas: "BasisMixin", type_label: str) -> str:
+def generate_composite_basis_labels(bas: BasisMixin, type_label: str) -> str:
     """Generate all labels in a basis tree."""
     if hasattr(bas, "basis1") and hasattr(bas, "basis2"):
         if type_label == "all" or bas._label:
@@ -278,7 +280,7 @@ def generate_composite_basis_labels(bas: "BasisMixin", type_label: str) -> str:
             yield getattr(bas, "label", bas.__class__.__name__)
 
 
-def label_setter(bas: "BasisMixin", label: str | None) -> None | ValueError:
+def label_setter(bas: BasisMixin, label: str | None) -> None | ValueError:
     """Label setter logic for any basis."""
     if not hasattr(bas, "basis1") or not hasattr(bas, "basis2"):
         return _atomic_basis_label_setter_logic(bas, label)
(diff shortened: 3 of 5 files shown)


billbrod (changes requested):
Some thoughts:
- Add link to [SPEC-0001](https://scientific-python.org/specs/spec-0001/), since that's what we're following.
- I think it's worth doing the lazy loading beyond just the top module, because it's totally possible that someone would want to just use the basis and therefore doesn't need optimistix, optax imported
- You talked about deferred errors, did you see discussion of `EAGER_IMPORT` in SPEC-0001? worth mentioning in the PR, might be worth mentioning somewhere in docs? or developer notes?
- stub maintenance doesn't actually add anything: you just add the import statements there instead of in `__init__.py`. You do need to define `__all__`, but you should probably be doing that anyway, `from X import *` is considered [bad practice](https://docs.astral.sh/ruff/rules/undefined-local-with-import-star/).
- In PR description, should explain why `lazy.load` is used where it is (and what it does).
- In PR description, should explain why the `if TYPE_CHECKING` blocks are added and what they do.
- This PR adds a bunch of `from __future__ import annotations`, why?
- Some other changes to imports (e.g., in `src/nemos/glm/validation.py`), why?
- Why is there a huge diff in `src/nemos/base_validator.py` and `src/nemos/validation.py`? was this started from an existing branch?

BalzaniEdoardo:
> Add link to [SPEC-0001](https://scientific-python.org/specs/spec-0001/), since that's what we're following.

Good point! I've added a link to SPEC-0001 in the implementation section. 

> I think it's worth doing the lazy loading beyond just the top module, because it's totally possible that someone would want to just use the basis and therefore doesn't need optimistix, optax imported

Agreed and implemented!

> You talked about deferred errors, did you see discussion of `EAGER_IMPORT` in SPEC-0001? worth mentioning in the PR, might be worth mentioning somewhere in docs? or developer notes?

Yes! The `EAGER_IMPORT` environment variable is part of SPEC-0001 and is supported by `lazy_loader`. When set, it disables lazy loading so import errors surface immediately rather than at first use.

I am not sure where to talk about it, probably dev notes?

> stub maintenance doesn't actually add anything: you just add the import statements there instead of in `__init__.py`. You do need to define `__all__`, but you should probably be doing that anyway, `from X import *` is considered [bad practice](https://docs.astral.sh/ruff/rules/undefined-local-with-import-star/).

Fair point.

> In PR description, should explain why `lazy.load` is used where it is (and what it does).

Added to the PR description.

> In PR description, should explain why the `if TYPE_CHECKING` blocks are added and what they do.

Added to the PR description.

> This PR adds a bunch of `from __future__ import annotations`, why?

`from __future__ import annotations` enables PEP 563 (postponed evaluation of annotations). This allows using forward references and string annotations without quotes, and importantly, **prevents annotation expressions from being evaluated at import time**.

Without it, type annotations like `def foo(x: Tsd) -> TsdFrame:` would require `Tsd` and `TsdFrame` to be imported at runtime. With PEP 563, annotations are stored as strings and only evaluated when introspected, which complements the `TYPE_CHECKING` pattern above.

> Some other changes to imports (e.g., in `src/nemos/glm/validation.py`), why?

The changes in `glm/validation.py` are import path updates due to the `RegressorValidator` class being moved to its own module (`base_validator.py`). The advantage is that now validation does not depend on sklearn (`RegressorValidator` was introducing the dependency), so now some of the modules that import `validation` are imported faster.

> Why is there a huge diff in `src/nemos/base_validator.py` and `src/nemos/validation.py`? was this started from an existing branch?

`base_validator.py` is new, the content that is in it was originally in `validation.py`, but for the point I made above, it made sense to separate the sklearn dependent class from it.

BalzaniEdoardo:
I did some digging on why importing `BaseEstimator` is so slow: it triggers the import of a ton of stuff, and it is quite annoying,

```
   sklearn.base                                                                                                                                                        
    → sklearn.utils._metadata_requests                                                                                                                                
      → sklearn.utils                                                                                                                                                 
        → sklearn.utils._chunking                                                                                                                                     
          → sklearn.utils._param_validation                                                                                                                           
            → sklearn.utils.validation                                                                                                                                
              → sklearn.utils._array_api                                                                                                                              
                → sklearn.utils.fixes   ← imports scipy.stats + pandas          
                  
```


                                                                                                                                                        
 **Breakdown:**

| Module | Time |
|--------|------|
| scipy.stats | ~340ms |
| pandas | ~210ms |
| scipy.sparse | ~68ms |
| numpy | ~32ms |
| **Total** | ~650ms+ |


Unfortunately, there we  dropped our minimal `Base` class to avoid implementing Tags internally: #368.

BalzaniEdoardo:
@billbrod I investigated what happens in terms of loading modules in a bunch of scenarios (note that for all this example you need to restart ipython):

1. **`TYPE_CHECKING` + string annotations**
This is fine for lazy loading:

```python
>>> from typing import TYPE_CHECKING
>>> import sys
>>> if TYPE_CHECKING:
>>>     import numpy as np  # Never imported at runtime
>>> def func(x: "np.ndarray") -> "np.ndarray":  # Quotes required!
>>>     return x
>>> "numpy" in sys.modules
False
```
2. **`TYPE_CHECKING` + `future annotations`**
Lazy Loading works even without quotes
```python
>>> from __future__ import annotations
>>> from typing import TYPE_CHECKING
>>> import sys
>>> if TYPE_CHECKING:
>>>     import numpy as np  
>>> def func(x: np.ndarray) -> np.ndarray:  # no quotes
>>>     return x
>>> "numpy" in sys.modules
False
>>> func.__annotations__
... {'x': 'np.ndarray', 'return': 'np.ndarray'}
```

3. **lazy_loader + `future annotations`**
Not enough for lazy loading.
```python
>>> from __future__ import annotations
>>> import lazy_loader as lazy
>>> import sys
>>> np = lazy.load("numpy")
>>> len([k for k in sys.modules if "numpy" in k]) # only placeholder is loaded, not full numpy
1
>>> def func(x: np.ndarray) -> np.ndarray:
>>>     return x
>>> func.__annotations__
{'x': 'np.ndarray', 'return': 'np.ndarray'}  # Stored as strings, good...
>>> len([k for k in sys.modules if "numpy" in k]) # but now numpy is fully loaded
86
```

After learning this, I opted for TYPE_CHECKING + `future annotations` everywhere for the following reasons:
- TYPE_CHECKING seems to be needed for lazy imports in the presence of type annotations
- future annotations simplifies the annotations (no quotation needed) and, most importantly, improves the representation of the docs:

Without `future annotations`
```python
>>> from numpy.typing import NDArray
>>> def func(x: NDArray) -> NDArray:
>>>     pass
>>> import inspect
>>> inspect.signature(func)
(x: numpy.ndarray[tuple[typing.Any, ...], numpy.dtype[~_ScalarT]]) -> ...
```

With `future annotations`
```python
>>> from __future__ import annotations
>>> from numpy.typing import NDArray
>>> def func(x: NDArray) -> NDArray:
>>>     pass
>>> import inspect
>>> inspect.signature(func)
<Signature (x: 'NDArray') -> 'NDArray'>
```

billbrod:
I was hoping that we could get away with only using lazy_loader and future, but ah well.

That's good to know about the representation in the docs! Thought this does remind me that sometimes that's not what we want (e.g., with the possible regularizers in GLM)

Issue #194, opened 2024-07-18
Make use of `jaxopt.StochasticSolver.run_iterator`


bagibence:
The [example in the docs](https://nemos.readthedocs.io/en/latest/generated/api_guide/plot_05_batch_glm/) currently uses a custom loop to implement stochastic gradient descent.

An alternative would be to make use of [`jaxopt.StochasticSolver.run_iterator`](https://jaxopt.github.io/stable/stochastic.html#run-iterator-vs-manual-loop) and add support for stochastic solvers, potentially including [optimizers implemented in Optax](https://optax.readthedocs.io/en/latest/api/optimizers.html) through the [`jaxopt.OptaxSolver`](https://jaxopt.github.io/stable/stochastic.html#optax-solvers) wrapper.

Additionally, for data that fits in memory, adding a faster version of this loop -- which implements sampling mini-batches and updating parameters (similarly to `ProxSVRG.run` in https://github.com/flatironinstitute/nemos/pull/184) -- could be useful.

BalzaniEdoardo:
We could define super-class class that is a general interface for stochastic solvers (with inner/outer loop structure  and the inner loop should be an abstract class, as well as the `run_iterator` for batching). The class should still provide the `run` method for data fitting in memory.

PR #510, opened 2026-02-25
Support stochastic optimization

### LINKED ISSUE #376: Stochastic optimization interface
Sometimes one wants to use stochastic optimization to fit models (e.g. when the whole data doesn't fit into memory). Currently there is a [tutorial about how to fit GLMs using batches of data](https://nemos.readthedocs.io/en/latest/how_to_guide/plot_04_batch_glm.html).

We agreed that a way to easily run a loop like this could be exposed to users on the `BaseRegressor` / `GLM` level, but it's not immediately clear what would be the best interface.

JAXopt had a `run_iterator` method for stochastic solvers, which took a generator / iterator as an argument instead of the full data matrices.
In the [new solver interface](https://github.com/flatironinstitute/nemos/pull/365)`StochasticSolverMixin` copies this and adds the same implementation of a `run_iterator` to gradient descent and proximal gradient solvers.

Some options that came up:
- `GLM.fit` could dispatch to `run` or `run_iterator` based on the type of the arguments.
- Create separate solver classes that are stochastic. So `GLM(solver_name="SGD")` would use `run_iterator`.
- Add a separate fit method, e.g. `GLM.stochastic_fit`.
- Have wrapper classes that handle this.
- Add just a helper function.

Since writing a generator / iterator that supplies batches is on a similar difficulty level for users as writing the optimization loop, functionality to create a generator given the full data could also be added.

bagibence:
## Purpose
Sometimes one wants to use stochastic optimization to fit models (e.g. when the whole data doesn't fit into memory). Currently there is a [tutorial about how to fit GLMs using batches of data](https://nemos.readthedocs.io/en/latest/how_to_guide/plot_04_batch_glm.html), but that requires users to implement the optimization loop themselves.

The purpose of this PR is to provide an easier-to-use interface for stochastic optimization and hide the implementation details from the user.
For example, SVRG is particularly suited for this kind of use, but the optimization loop is non-standard.

Stochastic optimization is added for the following solvers: gradient descent, proximal gradient, SVRG, Prox-SVRG.

## Main changes

- `AbstractSolver`:
    - Subclasses can set the `_supports_stochastic` class variable to True to signal that they implement stochastic optimization.
    - The `stochastic_run` method either raises an error for non-stochastic solvers or dispatches to `_stochastic_run_impl` for stochastic ones.
    - These accept data from a data loader and can use callback functions to stop the iteration on convergence or for any reporting.
        - Two callback types: convergence_criterion (per-epoch) and batch_callback (per-batch)
        - Solvers provide a default `stochastic_convergence_criterion` (Modified Cauchy criterion for Optimistix-based, step-size-normalized parameter change for JAXopt/SVRG)

- `batching.py`:
    - `DataLoader` protocol defines the data loader interface that `stochastic_run` expects.
    - `ArrayDataLoader` is provided as convenience for passing in-memory arrays to stochastic optimization.
    - `_PreprocessedDataLoader` is used to replicate the preprocessing that `GLM` does.

- `StochasticSolverMixin` provides a default optimization loop with callbacks that a few solvers borrow.

- `SVRG` has a custom `run_streaming` with a slightly different interface (to decouple it from `DataLoader`) and the full gradient is estimated by looping through batches.
    Note that this estimate is appropriate if the loss is a mean of losses across data points, which is true in NeMoS so far. I added notes about this in `_svrg.py`.

- `GLM.stochastic_fit` is the user-facing interface wrapping the solver's `stochastic_run`:
    - Data is given as a data loader.
    - It accepts `convergence_criterion=True` (use solver default), `False` (disable), or a custom callable.
    - Providing `init_params` lets users continue optimization after stopping.

- To discover stochastic solvers, users can use `list_stochastic_solvers()` and `supports_stochastic(solver_name)`.

- Acceleration and line searches are required to be disabled when using stochastic optimization.

## Open questions and todos
- `SVRG`'s batch size and learning rate initialization is currently done based on a sample batch. This is probably suboptimal, and it should use the full data, but I haven't looked into it in detail yet.

- `GLM.fit` sets `scale_` and `dof_resid_`, but those also need the whole data. Currently these are set to `None` in `stochastic_fit`.

- Could writing a `DataLoader` be made simpler for users?

- A potential problem I see is the case when batches are not the same size -- which is very common because the number of data points is most likely not an exact multiple of the batch size. Currently I think this triggers recompilation of the loss and its gradient. As optimization is a Python loop, this might not be that big of a performance hit, but might be better to avoided somehow (e.g. with some padding).

- Stochastic optimization run for `num_epochs` or until convergence, and `maxiter` is ignored. Should there be a warning about this? Or should this behavior be changed? 

Fixes #376 and #194


src/nemos/glm/glm.py
```diff
@@ -809,6 +816,148 @@ def fit(
...
+                "(X_batch, y_batch) tuples."
+            )
+        loader = data
+
+        # Get raw sample batch for initialization
+        raw_sample_X, raw_sample_y = loader.sample_batch()
+        self._validator.validate_inputs(raw_sample_X, raw_sample_y)
+
+        # Preprocess sample batch (cast to jax, drop nans, etc.)
+        sample_X, sample_y = self._preprocess_inputs(raw_sample_X, raw_sample_y)
```
BalzaniEdoardo, line 913: how does the caching behaves when nans are dropped. Is the stochastic run loop re-trigger the compilation every time batches have different sizes, or will it cache previously observed batch sizes?

For example, if the batches can have one of two sample sizes N1 and N2, and they alternate, will the compilation be triggered at each iteration?

BalzaniEdoardo: We should know if re-compilation happens and if so, we should figure out how to use a cache

src/nemos/glm/glm.py
```diff
@@ -809,6 +816,148 @@ def fit(
...
+            getattr(self, "_feature_mask", None), init_params
+        )
+
+        # Wrap data loader to preprocess each batch from now on
+        preprocessed_loader = _PreprocessedDataLoader(loader, self._preprocess_inputs)
+
+        # TODO: This can be problematic for setting the right step- and batch size for SVRG
+        # Ideally that uses the full data
+        # Initialize solver (using preprocessed sample batch)
+        self._initialize_optimizer_and_state(init_params, sample_X, sample_y)
```
BalzaniEdoardo, line 940: That is a good point, for SVRG we may want to warn the user. And the warning may include a suggestion on how to pre-compute batch-size and stepsize, and provide them as solver_kwargs

src/nemos/glm/glm.py
```diff
@@ -809,6 +816,148 @@ def fit(
...
+        # TODO: If continuing a fit with init_params, should this be turned off so that the solver is not reinstantiated?
```
BalzaniEdoardo: yes, this would avoid re-compilaiton, but that's its own can of worms, because we need to make sure that the model state is not changed. If we have a robust mechanism to do so, we can take advantage of it even in regular fit calls: if the model is unchanged, or if it is changed in ways that do not require re-instantiating a solver, then re-use what we have. Since this can lead to silent bugs if we are not careful, I think it should be the subject of a dedicated pr

BalzaniEdoardo: I was also thinking of caching solvers: have a dict with keys uniquely determined by solver state, and if the solver is in the dict, just re-use it. This would be very helpful for cross-validation, even if sklearn clone is a fresh instance (with no solver attached) we can re-use the solver from the cache, no compilation needed

src/nemos/glm/glm.py
```diff
@@ -809,6 +816,148 @@ def fit(
...
+            )
+
+        # not warning about non-convergence
+
+        # Store results
+        self._set_model_params(params)
+        self.solver_state_ = state
+        self.aux_ = aux
+
+        # TODO: Do we need to implement this?
```
BalzaniEdoardo: another good question:  without the scale we cannot compute the full log-likelihood of models like Gaussian, Gamma and NegativeBinomial. The dof is required to estimate the scale of some of these models. So, not having the scale can be an issue for whatever the user tries to do next. the dof estimation is trivial for lasso, group lasso, ridge, and elastic net. for the unregularized it needs the rank of X, which is hard to estimate if X is too large. 

On the other hand, one could estimate the scale by maximizing the likelihood as a function of the scale instead of the parameters.

BalzaniEdoardo: for the scale we need the predicted rate, which could be stored most likely for a single neuron glm, for a population glm it may be too large.

BalzaniEdoardo (commented):
I posted a few thoughts, but there are still things to be figured out. We should also try this with a really large problem, for example a fully coupled large population, that we can store as a dask array or a zarr, to check how it performs

src/nemos/glm/glm.py
bagibence: Yes, a streaming implementation for the others should be possible, but to avoid making this PR even bigger and delay it, I will include it in another.
For now I added a quick workaround, setting the scale for observations that have a fixed scale and setting it to `None` for the others. `score()` raises if scale is needed but is not set. Also added a warning about this behavior.

src/nemos/glm/glm.py
bagibence: Agreed.
I removed this todo, only leaving the one about SVRG. That will be included in the PR doing streaming estimation for `scale` and `dof_resid`.

src/nemos/glm/glm.py
bagibence: I assume it would be cached and only compiled twice, but I will look into it.

src/nemos/glm/glm.py
bagibence: I added a warning about this.

src/nemos/glm/glm.py
bagibence: It's compiled twice in your example, so once for each unique batch size. I added a note about dropping NaNs in the docstring of `GLM.stochastic_fit`.

docs/developers_notes/07-solvers.md
```diff
@@ -122,13 +124,76 @@ To validate a solver without registering, the [`nemos.solvers.validate_solver_cl
...
+X = jnp.ones((10000, 50))
+y = jnp.ones((10000,))
+loader = ArrayDataLoader(X, y, batch_size=128, shuffle=True)
+
+# Fit model using stochastic optimization
+model = nmo.glm.GLM(
+    solver_name="GradientDescent",
+    solver_kwargs={"stepsize": 0.01, "acceleration": False}
+)
+model.stochastic_fit(loader, num_epochs=10)
```
BalzaniEdoardo, line 150: When I run the method the coef and intercept were set to all zeros after 790 iterations, I know this is just a note, but I would still generate some random X and y ~ Poisson( exp(X.dot(w))), so that the fit actually does something

docs/how_to_guide/plot_04_batch_glm.md
```diff
@@ -33,181 +33,431 @@ warnings.filterwarnings(
...
 ```
 
 # Batching example
 
-Here we demonstrate how to setup and run a stochastic gradient descent in `nemos`
-by batching and using the [`update`](nemos.glm.GLM.update) method of the model class.
+Here we demonstrate how to set up and run stochastic gradient descent in `nemos` -- the standard approach when your dataset does not fit in memory and you need to train the GLM one mini-batch at a time, using each batch to update the model parameters.
+
+Data will be batched using `pynapple`, fed to the GLM's [`stochastic_fit`](nemos.glm.PopulationGLM.stochastic_fit) method with a [`DataLoader`](nemos.batching.DataLoader), and the loss will be logged by a custom callback.
+\
```
BalzaniEdoardo:
```suggestion

```

docs/how_to_guide/plot_04_batch_glm.md
```diff
@@ -33,181 +33,431 @@ warnings.filterwarnings(
...
+    message=("invalid value encountered in div "),
     category=RuntimeWarning,
 )
 ```
 
 # Batching example
 
-Here we demonstrate how to setup and run a stochastic gradient descent in `nemos`
-by batching and using the [`update`](nemos.glm.GLM.update) method of the model class.
+Here we demonstrate how to set up and run stochastic gradient descent in `nemos` -- the standard approach when your dataset does not fit in memory and you need to train the GLM one mini-batch at a time, using each batch to update the model parameters.
```
BalzaniEdoardo: This is great but this can be scorporated into multiple notes. You can use a folder structure. See main comment

src/nemos/glm/glm.py
```diff
@@ -810,6 +826,231 @@ def fit(
...
+
+        if tree_utils.pytree_map_and_reduce(
+            lambda x: jnp.any(jnp.isnan(x)), any, params
+        ):
+            raise ValueError(
+                "Solver returned at least one NaN parameter, so solution is invalid!"
+                " Try tuning optimization hyperparameters, specifically try decreasing the `stepsize`."
+            )
+
+        # not warning about non-convergence
```
BalzaniEdoardo, line 1105:
```suggestion
        # no warning about non-convergence
```

src/nemos/glm/glm.py
```diff
@@ -810,6 +826,231 @@ def fit(
...
+        Quick fix until a streaming residual d.o.f. and scale estimation is added.
+        """
+        return isinstance(
+            self.observation_model,
+            (
+                obs.PoissonObservations,
+                obs.BernoulliObservations,
+                obs.CategoricalObservations,
+                obs.NegativeBinomialObservations,
+            ),
```
BalzaniEdoardo, line 1187: we may have something like this in the glm-hmm module. check if there can be reuse, and move to obs model if it is usable in more than one place

src/nemos/hmm/hmm.py
```diff
@@ -879,3 +881,16 @@ def decode_state(
         # define the return type for the max-sum
         return_index = False if state_format == "one-hot" else True
         return self._decode_state(params, X, y, is_new_session, return_index)
+
+    def stochastic_fit(
```
BalzaniEdoardo: do we need it in base regressor? we can also make it optional for models that can be fit that way

src/nemos/base_regressor.py
```diff
@@ -425,6 +429,58 @@ def fit(
         """Fit the model to neural activity."""
         pass
 
+    @abc.abstractmethod
+    def stochastic_fit(
```
BalzaniEdoardo: unsure if we should have it

src/nemos/solvers/_svrg.py
```diff
@@ -522,6 +538,44 @@ def cond_fun(step):
...
+        """
+        total_grad = None
+        total_samples = 0
+
+        for batch_data in iter_batches():
+            batch_size = self._infer_number_of_samples(*batch_data)
+            batch_grad, _ = self.loss_gradient(params, *batch_data)
+
+            if total_grad is None:
+                total_grad = jax.tree.map(lambda g: g * batch_size, batch_grad)
```
BalzaniEdoardo: this works fine (treemap is eager) but i would use partial to make sure that batch_size is is updated per iteration

src/nemos/batching.py
```diff
@@ -0,0 +1,360 @@
...
+
+    def __iter__(self) -> Iterator[BatchData]:
+        """Iterate with preprocessing applied to each batch."""
+        for batch_data in self._loader:
+            yield self._preprocess_fn(*batch_data)
+
+
+class LazyArrayDataLoader:
+    """
+    DataLoader for lazy/out-of-core arrays (e.g. dask, zarr, HDF5).
```
BalzaniEdoardo, line 316: can this work with pynapple loading nwbs? or memmapped pynapple?

src/nemos/batching.py
```diff
@@ -0,0 +1,360 @@
...
+            raise ValueError("batch_size must be positive.")
+        if val > self.n_samples:
+            raise ValueError("batch_size cannot be larger than the number of samples.")
+
+        self._batch_size = val
+
+    def sample_batch(self) -> tuple[jnp.ndarray, ...]:
+        """Return first batch, deterministic (ignores shuffle)."""
+        end = min(self.batch_size, self.n_samples)
+        return tuple(jnp.asarray(arr[:end]) for arr in self.arrays)
```
BalzaniEdoardo: this works with pynapple

src/nemos/batching.py
```diff
@@ -0,0 +1,360 @@
...
+    def _iter_no_shuffle(self) -> Iterator[tuple[jnp.ndarray, ...]]:
+        """Return fresh iterator. Yields batches in sequential order."""
+        n = self.n_samples
+        chunks = [
+            (start, min(start + self.batch_size, n))
+            for start in range(0, n, self.batch_size)
+        ]
+
+        for start, end in chunks:
+            batch = tuple(jnp.asarray(arr[start:end]) for arr in self.arrays)
```
BalzaniEdoardo: this too

src/nemos/batching.py
```diff
@@ -0,0 +1,360 @@
...
+        """Return fresh iterator. Shuffles chunk order and within-batch."""
+        n = self.n_samples
+        chunks = [
+            (start, min(start + self.batch_size, n))
+            for start in range(0, n, self.batch_size)
+        ]
+        self._rng.shuffle(chunks)
+
+        for start, end in chunks:
+            batch = tuple(jnp.asarray(arr[start:end]) for arr in self.arrays)
```
BalzaniEdoardo: I think we should advertise (and test) that this works with memmapped pynapple, which is a common neuroscience entry point

https://pynapple.org/user_guide/02_input_output.html

src/nemos/glm/glm.py
```diff
@@ -810,6 +826,231 @@ def fit(
...
+        self._validator.feature_mask_consistency(
+            getattr(self, "_feature_mask", None), init_params
+        )
+
+        # Wrap data loader to preprocess each batch from now on
+        preprocessed_loader = _PreprocessedDataLoader(loader, self._preprocess_inputs)
+
+        # TODO: Add a streaming version setting the right step- and batch size for SVRG that uses the full data
+        # Initialize solver (using preprocessed sample batch)
+        self._initialize_optimizer_and_state(init_params, sample_X, sample_y)
```
BalzaniEdoardo, line 1084: this initializes the state once here, then one other time in the `solver.stochastic_run` discarding the already initialized state. For SVRG, this wastes a full gradient compute, which can be expensive.

Consider hoisting the sample-batch init out of the mixin, or wiring the already-built state through.

src/nemos/solvers/_stochastic_mixins.py
```diff
@@ -0,0 +1,237 @@
...
+    Supports optional callbacks for per-epoch and per-batch hooks.
+
+    Classes using this mixin must implement ``init_state`` and ``update`` methods.
+    """
+
+    _supports_stochastic = True
+
+    def _validate_stochastic_options(self):
+        """Make sure acceleration and linesearches are turned off if available."""
+        if "acceleration" in self.get_accepted_arguments():
```
BalzaniEdoardo: this may fail silently if there will be a solver that does acceleration but doesn't store it as an attribute. what about having an abstract method for stochastic solver mixins `has_acceleration` and if that returns true we assume that the solver has an `acceleration` attribute and we call directly `self.acceleration`? open to other solutions

src/nemos/solvers/_stochastic_mixins.py
```diff
@@ -0,0 +1,237 @@
...
+            if self.acceleration:
+                raise ValueError(
+                    "Turn off acceleration option for stochastic optimization."
+                )
+
+        if "stepsize" in self.get_accepted_arguments():
+            if self.stepsize is None or self.stepsize <= 0.0:
```
BalzaniEdoardo: similar thing for the adaptive stepsize

BalzaniEdoardo: also, the more direct property to check is `has_linesearch`, and then we should deactivate the search. The None or -1 conventions for activating the line searches seems more brittle.

src/nemos/glm/glm.py
```diff
@@ -810,6 +826,231 @@ def fit(
...
+        self.dof_resid_ = None
+
+        return self
+
+    def _warn_about_estimated_svrg_settings(self):
+        """Warn if SVRG settings are estimated on sample data instead of the full dataset."""
+
+        batch_size_estimated_for_svrg = False
+        if (
+            "svrg" in self.solver_name.lower()
```
BalzaniEdoardo: I would prefer isinstance here

BalzaniEdoardo (changes requested):
This is an excellent PR, thanks for that. I have a some comments, but very tractable.

## Major points

### 1. Split the batching how-to into a folder

`docs/how_to_guide/plot_04_batch_glm.md` now covers loaders, custom loaders, callbacks, logging, stopping criteria, continuation, and a full-vs-batch comparison in one page. That is too much for users to navigate. Mirror the structure being used in the categorical basis PR and break this into a folder, keeping basics together and pulling out the topics that don't belong in "basics":

```
batching/
    index.md             # intro + when to use stochastic_fit, TOC
    stochastic_fit.md    # basics: DataLoader (including a pynapple/NWB lazy example),
                         # stochastic_fit, basic loss logging, continuing with `init_params`
    manual_batching.md   # current `update`-based manual loop (the old plot_04 content)
    early_stopping.md    # SolverConvergenceCallback + custom early-stopping callback
```

Rationale: the loader → fit → log → continue flow is one coherent learning topic and belongs in a single page (`stochastic_fit.md`). Convergence-based / custom stopping is a separate concern that not every user will read, so it splits cleanly into its own page. Manual batching stays as a standalone reference for users who need the lower-level loop.

This also helps with the PR description's "Could writing a `DataLoader` be made simpler for users?" question — surfacing the *minimum* example becomes a docs problem first.

### 2. `BaseRegressor.stochastic_fit` should not be abstract

Whether a model supports stochastic optimization is a per-model decision (HMM already opts out by raising `NotImplementedError`). Forcing every `BaseRegressor` subclass to implement it — even as a stub — is the wrong default. Options:

- Drop the `@abc.abstractmethod` and leave `stochastic_fit` off `BaseRegressor` entirely; only models that support it (currently `GLM`) define it.
- Or expose it as a mixin (`StochasticFitMixin` / `StochasticallyFittable`) that concrete models opt into.

The mixin route also moves the GLM-specific implementation (preprocessing, scale handling, NaN check, SVRG warnings) out of `BaseRegressor`'s vicinity, where it doesn't really belong. A consequence: `HMM.stochastic_fit` and the `MockRegressor.stochastic_fit` stub in `tests/conftest.py` can both go away.

### 3. Pynapple compatibility for `LazyArrayDataLoader` should be first-class

`LazyArrayDataLoader` only needs `.shape[0]` and `arr[start:end]`, which is exactly what `pynapple`'s `Tsd` / `TsdFrame` / lazy NWB-backed objects expose ([pynapple I/O docs](https://pynapple.org/user_guide/02_input_output.html)). `jnp.asarray` on a `Tsd` already returns a plain JAX array of the values, so the existing loader works without an adapter. Pynapple is the project's recommended entry point for neuro data and the most common source of lazy arrays our users will hit — this should be:

- Called out explicitly in the docstring and in `stochastic_fit.md` (with a short NWB-backed example showing a pynapple object passed straight to `LazyArrayDataLoader`).
- Added as a `LazyArrayDataLoader` test backend alongside numpy / memmap / zarr / hdf5 in `tests/test_batching.py::TestLazyArrayDataLoader` (`_pynapple_arrays` fixture). The existing `request.getfixturevalue` indirection already handles new backends cleanly.



## Test gap
No test that `SolverConvergenceCallback` on a `JaxoptGradientDescent` with a fixed stepsize doesn't `AttributeError` on `state.solver_state.stepsize`.

src/nemos/solvers/_svrg.py
bagibence: Even simpler, I switched to `tree_scalar_mul` and `tree_add_scalar_mul` instead.

src/nemos/glm/glm.py
bagibence: Done.

src/nemos/base_regressor.py
bagibence: I was unsure as well when I wrote it. Deleted it.

src/nemos/hmm/hmm.py
bagibence: Deleted.

src/nemos/glm/glm.py
bagibence: There is a `_NO_SCALE = (PoissonObservations, BernoulliObservations)`, so not exactly the same.

docs/developers_notes/07-solvers.md
bagibence: Makes sense, updated to do exactly that.

src/nemos/solvers/_stochastic_mixins.py
bagibence: I added abstract `acceleration_turned_on` and `linesearch_turned_on` properties to `StochasticSolverMixin` and defined them in current stochastic-capable solvers.
The check is done in all solvers that derive from `StochasticSolverMixin`. So custom solvers have to adhere to this if they use this mixin, but are free to do what they want if they implement stochastic fitting their way.

src/nemos/solvers/_stochastic_mixins.py
bagibence: Added `linesearch_turned_on`.
Longer answer above: https://github.com/flatironinstitute/nemos/pull/510#discussion_r3287804836

bagibence:
Before I change too much on the how-to guide, what do you think about the following structure?

```
batching/
    index.md                               # intro + when to use stochastic_fit, TOC
    stochastic_fit.md                      # basics with built-ins
    custom_dataloader.md                   # PynappleDataLoader
    custom_callbacks_and_termination.md    # SolverConvergenceCallback + custom early-stopping callback
    manual_batching.md                     # current `update`-based manual loop (the old plot_04 content)
```

So the basic page would show everything using built-in features. It would start with a hidden cell that generates random data, calculates a design matrix, writes X and y into an NWB file or numpy memmap. Loading real measurement data from one of the other tutorials doesn't work because they need preprocessing.
Then it's lazily loaded with `pynapple`, passed to `stochastic_fit` with `LazyArrayDataLoader`. Do score logging with a built-in logger, plot the loss, then show you can continue a fit with `init_params`.
That would mean shipping `TestScoreLoggingCallback` with some enhancements in nemos, which I think would make sense. (There are other basic callbacks I could also include, for example a progress bar with `tqdm`.)

Then the more advanced pages would show how to write:
- custom data loaders through `PynappleDataLoader`
- custom callbacks and terminating through `EarlyStoppingCallback`
- manually loops calling update for more control

The custom data loaders and callbacks could also be in a single page.

BalzaniEdoardo:
> Before I change too much on the how-to guide, what do you think about the following structure?
> 
> ```
> batching/
>     index.md                               # intro + when to use stochastic_fit, TOC
>     stochastic_fit.md                      # basics with built-ins
>     custom_dataloader.md                   # PynappleDataLoader
>     custom_callbacks_and_termination.md    # SolverConvergenceCallback + custom early-stopping callback
>     manual_batching.md                     # current `update`-based manual loop (the old plot_04 content)
> ```
> 
> So the basic page would show everything using built-in features. It would start with a hidden cell that generates random data, calculates a design matrix, writes X and y into an NWB file or numpy memmap. Loading real measurement data from one of the other tutorials doesn't work because they need preprocessing. Then it's lazily loaded with `pynapple`, passed to `stochastic_fit` with `LazyArrayDataLoader`. Do score logging with a built-in logger, plot the loss, then show you can continue a fit with `init_params`. That would mean shipping `TestScoreLoggingCallback` with some enhancements in nemos, which I think would make sense. (There are other basic callbacks I could also include, for example a progress bar with `tqdm`.)
> 
> Then the more advanced pages would show how to write:
> 
>     * custom data loaders through `PynappleDataLoader`
> 
>     * custom callbacks and terminating through `EarlyStoppingCallback`
> 
>     * manually loops calling update for more control
> 
> 
> The custom data loaders and callbacks could also be in a single page.

I think this all looks good to me! You can proceed with the plan

src/nemos/batching.py
bagibence: The new how-to guide does that.

src/nemos/batching.py
bagibence: Yes, the new how-to guide demonstrates that.

docs/how_to_guide/plot_04_batch_glm.md
bagibence: Reformatted the docs into a multiple sections.

docs/how_to_guide/batching/custom_callbacks_and_termination.md
```diff
@@ -0,0 +1,319 @@
...
+import nemos as nmo
+
+jax.enable_x64()
+nap.nap_config.suppress_conversion_warnings = True
+np.random.seed(123)
+```
+
+First, quickly set up and run a stochastic fit for 10 epochs:
+
+```{code-cell} ipython3
```
BalzaniEdoardo, line 59: Why are these cells split? maybe a single cell would be better

docs/how_to_guide/batching/custom_dataloader.md
```diff
@@ -0,0 +1,184 @@
...
+```
+
+## Create the loader
+
++++
+
+Spike times will be binned in 5 ms bins and each neuron's firing rate will be predicted from the recent spike history of the whole population.
+We build those history features using a NeMoS convolutional basis with a window size of 200 ms.
+
+```{code-cell} ipython3
```
BalzaniEdoardo, line 145: can you merge these cells?

docs/how_to_guide/batching/manual_batching_loop.md
```diff
@@ -0,0 +1,100 @@
...
+np.random.seed(123)
+```
+
+This approach doesn't require the use of a NeMoS data loader, you can load batches of data any way you want. Here, we will use in-memory data and `ArrayDataLoader` for simplicity.
+
+```{code-cell} ipython3
+units, spike_trains, X = _simulate_batching_data()
+```
+
+```{code-cell} ipython3
```
BalzaniEdoardo:
```suggestion
```

```

BalzaniEdoardo (approved):
this all looks good, can you please merge all the code blocks that are not separated by any text? 

After this and the ridge fixes LGTM

docs/how_to_guide/batching/custom_callbacks_and_termination.md
bagibence: I thought it looked better. I'll just merge them.

docs/how_to_guide/batching/custom_dataloader.md
bagibence: Done.

docs/how_to_guide/batching/manual_batching_loop.md
bagibence: Done.

PR #518, opened 2026-03-02
Allow any pytree

BalzaniEdoardo:
# PR: Arbitrary Pytree Support for X Input

## Summary

Extend the test suite and documentation to formally cover arbitrary JAX pytrees as the `X`
argument to all GLM models, and fix two source-level issues discovered during testing.

## Motivation

`FeaturePytree` was the only officially documented and tested path for grouped feature input.
JAX's own pytree mechanism is more general (dicts, `equinox.Module`, nested containers, etc.)
and avoids bespoke wrapper types. This PR validates that the implementation already supports
arbitrary pytrees, removes `FeaturePytree` from every X-parameter description, and adds
regression coverage so future refactors cannot silently break pytree inputs.

### Source fixes

**`base_validator.py` — `validate_inputs`**: replaced `X.shape[0]` with a loop over 
`jax.tree_util.tree_leaves(X)` to obtain n_samples from an arbitrary pytree. Also added
a consistency check that all leaves share the same number of samples.

**`validation.py`**: updated structure-mismatch error messages to report
`jax.tree_util.tree_structure(X)` vs `jax.tree_util.tree_structure(coef)` instead of
assuming X is an array; 

**`glm.py` — `initialize_params`**: added a `feature_mask_consistency` call so that a
stale array `_feature_mask` set by a previous `fit(array_X, y)` raises a clear `TypeError`
instead of silently producing wrong shapes when the model is re-initialized with a pytree X. Added a test for this new behavior, see below. This was not necessary for the pytree support but that I caught this as I was writing new tests.

### Tests

| Layer | File | What is tested |
|---|---|---|
| 1 | `test_proximal_operator.py` | Each of the 5 prox operators accepts a pytree of params; array-vs-tree equivalence |
| 2 | `test_regularizer.py` | `_SIMPLE_PYTREE_CASES` uses plain dict / eqx module / dict-list instead of FeaturePytree |
| 3 | `test_base_regressor_subclasses.py` | `TestModelVsPytree` — all public API methods involving X × models × pytree-type |
| 4 | `test_glm.py` | Coef tree structure matches X after `initialize_params` and `fit`; stale mask regression |

`test_initialize_params_stale_feature_mask_raises` guards the source fix above: before this
PR the mask/pytree mismatch would surface at the first loss evaluation, producing a cryptic
shape error far from the site of the mistake.

### Documentation

- Removed every X-parameter reference to `FeaturePytree` in method docstrings across
  `GLM`, `PopulationGLM`, `ClassifierGLM`, and `ClassifierPopulationGLM`; replaced with
  "pytree of arrays of the same shape".
- Updated `PopulationGLM` class docstring: replaced the "Mask Coefficients with a
  FeaturePytree" example with a plain-dict equivalent.
- Added a "Use a Pytree of Arrays as Input" example to the class docstrings of `GLM`,
  `ClassifierGLM`, and `ClassifierPopulationGLM`.

- **FeaturePytree backward compatibility**. I would want to drop the FeaturePytree implementation entirely at this point. In a subsequent PR I would like to add a deprecation warning. For how many versions should we be backward compatible?
- What should I do with the showcasing the FeaturePytree? (Probably nothing in this one but in the next PR deprecating the feature pytree)
  - **plot_03_population_glm.md**: This shows that the `feature_mask` is automatically generated based on the tree structure, one group per leaf. I think this should stay and be replaced with a regular dictionary.
  - **plot_07_glm_pytree.md**: This one literally shows the FeaturePytree on its own. Should we show the pytree compatibility with different tree types instead or should we just delete it?
  - **finegrained_regularization.md**: This shows how to regularize with structured regularization strength, as long as the strength structure matches that of the parameters. I could just show how to do it with dict and list of coefficients.
  - **variable_selection_group_lasso.md**: This had a note on how to structure the regularization by grouping variables (using FeaturePytree for now). What should I do here? If I keep `plot_07_glm_pytree` using other tree types, I can keep the link to it.

### DIFF
--- a/src/nemos/glm/classifier_glm.py
+++ b/src/nemos/glm/classifier_glm.py
@@ -763,7 +763,7 @@ class ClassifierGLM(ClassifierMixin, GLM):
     >>> model.regularizer
     Ridge()
 
-    **Use a Pytree of Arrays as Input**
+    **Use a Pytree of arrays as Input**
 
     Features can be passed as any JAX pytree of 2-D arrays; the fitted
     ``coef_`` will share the same pytree structure:
@@ -1038,7 +1038,7 @@ class ClassifierPopulationGLM(ClassifierMixin, PopulationGLM):
     >>> model.regularizer
     Ridge()
 
-    **Use a Pytree of Arrays as Input**
+    **Use a Pytree of arrays as Input**
 
     Features can be passed as any JAX pytree of 2-D arrays; the fitted
     ``coef_`` will share the same pytree structure:
--- a/src/nemos/glm/classifier_glm.py
+++ b/src/nemos/glm/classifier_glm.py
@@ -266,7 +266,7 @@ class ClassifierMixin:
         ----------
         X :
             The input samples. Can be an array of shape ``(n_samples, n_features)``
-            or a ``FeaturePytree`` with arrays as leaves.
+            or a pytree of arrays of the same shape.
 
         Returns
         -------
@@ -306,7 +306,7 @@ class ClassifierMixin:
         ----------
         X :
             The input samples. Can be an array of shape ``(n_samples, n_features)``
-            or a ``FeaturePytree`` with arrays as leaves.
+            or a pytree of arrays of the same shape.
         return_type :
             The format of the returned probabilities. If ``"log-proba"``, returns
             log-probabilities. If ``"proba"``, returns probabilities. Defaults to
@@ -432,8 +432,8 @@ class ClassifierMixin:
             A JAX random key used to generate the simulated responses.
         feedforward_input :
             The input samples used to generate the responses. Can be an array of
-            shape ``(n_samples, n_features)`` or a ``FeaturePytree`` with arrays
-            as leaves.
+            shape ``(n_samples, n_features)`` or a pytree of arrays of the same
+            shape.
 
         Returns
         -------
@@ -578,7 +578,7 @@ class ClassifierMixin:
             optimization algorithm to continue from the current state.
         X :
             The predictors used in the model fitting process. Shape ``(n_time_bins, n_features)``
-            or a ``FeaturePytree``.
+            or a pytree of arrays of the same shape.
         y :
             Class labels, array of shape ``(n_time_bins,)`` for single neuron
             models or ``(n_time_bins, n_neurons)`` for population models. Labels must
@@ -803,7 +803,7 @@ class ClassifierGLM(ClassifierMixin, GLM):
         Parameters
         ----------
         X
-            Training input samples of shape ``(n_samples, n_features)`` or FeaturePytree.
+            Training input samples of shape ``(n_samples, n_features)`` or a pytree of arrays of the same shape.
         y
             Target class labels of shape ``(n_samples,)``. Values should be in
             ``[0, n_classes - 1]``. Float arrays with integer values are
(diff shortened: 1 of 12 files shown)


src/nemos/glm/classifier_glm.py
```diff
@@ -1026,6 +1037,16 @@ class labels from the provided ``y``. If you set ``coef_`` and ``intercept_`` ma
     ... )
     >>> model.regularizer
     Ridge()
+
+    **Use a Pytree of Arrays as Input**
```
wulfdewolf: uncapitalise 'Arrays', to be consistent

src/nemos/glm/classifier_glm.py
```diff
@@ -762,6 +762,17 @@ class labels from the provided ``y``. If you set ``coef_`` and ``intercept_`` ma
...
```
wulfdewolf: uncapitalise 'Arrays', to be consistent

src/nemos/glm/glm.py
```diff
@@ -229,6 +229,17 @@ class GLM(BaseRegressor[GLMUserParams, GLMParams]):
     >>> model = nmo.glm.GLM(solver_name="LBFGS").fit(X, y)
     >>> model.solver_name
     'LBFGS[...]'
+
+    **Use a Pytree of Arrays as Input**
```
wulfdewolf: uncapitalise 'Arrays', to be consistent

src/nemos/glm/glm.py
```diff
@@ -384,7 +395,8 @@ def predict(self, X: DESIGN_INPUT_TYPE) -> jnp.ndarray:
         Parameters
         ----------
         X :
-            Predictors, array of shape ``(n_time_bins, n_features)`` or pytree of same.
+            Predictors, array of shape ``(n_time_bins, n_features)`` or pytree of the same
```
wulfdewolf: or a pytree of arrays of the same shape.

to be consistent

src/nemos/glm/glm.py
```diff
@@ -502,7 +514,8 @@ def score(
         Parameters
         ----------
         X :
-            The exogenous variables. Shape ``(n_time_bins, n_features)``.
+            Predictors, array of shape ``(n_time_bins, n_features)`` or pytree of the same
+            shape.
```
wulfdewolf: or a pytree of arrays of the same shape.

to be consistent

tests/conftest.py
```diff
@@ -69,14 +73,17 @@ def initialize_feature_mask_for_population_glm(X, n_neurons: int, coef=None):
         of shape (n_neurons,) for each key.
         If X is an array, returns an array of shape (n_features, n_neurons).
     """
+    extra_shape = (n_classes,) if n_classes else ()
     if coef is not None:
         return jax.tree_util.tree_map(lambda c: jnp.ones(c.shape), coef)
     if isinstance(X, FeaturePytree):
```
wulfdewolf, line 79: this is kept for backward compability then?

wulfdewolf (commented):
LGTM, some minor text comments in docstrings.

Regarding the documentation:
- `plot_03_population_glm.md`: I agree, keep but replace with `dict` pytrees
- `plot_07_glm_pytree.md`: I like the idea of having a dedicated little notebook to first of all explain the basic concept of a pytree and then also how NeMoS interacts with them
- `finegrained_regularization.md`: I agree, replace with `dict` pytrees
- `variable_selection_group_lasso.md`: I agree, keep, reference in the pytree notebook

tests/conftest.py
BalzaniEdoardo: yes, for now. It is very cool to see that dropping the FeaturePytree gets rid of the conditional logic and makes the code simpler and more general

PR #524, opened 2026-03-14
Mask check group lasso

BalzaniEdoardo:
Quick PR, improving the error interpretability for group lasso misshaped mask and fixing the docstring description for the mask.

### DIFF
--- a/src/nemos/regularizer.py
+++ b/src/nemos/regularizer.py
@@ -723,6 +723,27 @@ class GroupLasso(Regularizer):
     >>> model = GLM(regularizer=group_lasso, regularizer_strength=0.1).fit(X, y)
     >>> print(f"coeff shape: {model.coef_.shape}")
     coeff shape: (5,)
+
+    For a :class:`~nemos.glm.PopulationGLM`, where ``coef_`` has shape
+    ``(n_features, n_neurons)``, the mask must have the matching shape
+    ``(n_groups, n_features, n_neurons)``:
+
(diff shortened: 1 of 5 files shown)


src/nemos/regularizer.py
```diff
@@ -688,11 +688,13 @@ class GroupLasso(Regularizer):
...
-        A 2d mask array indicating groups of features for regularization, shape ``(num_groups, num_features)``.
-        Each row represents a group of features.
-        Each column corresponds to a feature, where a value of 1 indicates that the feature belongs
-        to the group, and a value of 0 indicates it doesn't.
-        Default is ``mask = np.ones((1, num_features))``, grouping all features in a single group.
+        A mask array (or PyTree of arrays) indicating group membership for regularization.
+        Each regularizable parameter leaf with shape ``(n_features, ...)`` requires a corresponding
+        mask leaf with shape ``(n_groups, n_features, ...)``, i.e. ``(n_groups, *params.shape)``.
+        Row ``i`` of a mask leaf is 1 for features belonging to group ``i`` and 0 elsewhere;
+        each feature may belong to at most one group.
```
wulfdewolf, line 695: Wondering if we could refer to the PopulationGLM explicitly here?
I know we want to keep the regularizer as separate as possible, but it would help the user imo.
Maybe in a docstring example?

wulfdewolf (commented):
One comment about mentioning PopulationGLM explicitly, aside from that looks good to me!

PR #527, opened 2026-03-20
Add `CategoryBasis` for One-Hot Encoding of Discrete Variables

### LINKED ISSUE #486: Basis for categorical variables
**Is your feature request related to a problem? Please describe.**
Add basis that simplifies processing of categorical variables and interaction between categories and other covaraites. This seems to be a common design matrix construction pattern, see discussion #473 . 

**Describe the solution you'd like**
A simple solution would be a basis that maps a vector of categories to a 1-hot encoded output. This type of basis, multiplied with any other basis, will generate an interaction `category x predictor`, allowing learning one predictor per category. It could also be used directly to form a predictor that learns an offset per category.

```python
>>> import numpy as np
>>> import nemos as nmo

>>> # Example API
>>> trial_type_basis = nmo.basis.Category(n_categories=3)
>>> trial_types = np.array(["A", "A", "B", "C"])

>>> X = trial_type_basis.compute_features(trial_types)
>>> X
Array([[1., 0., 0.],
       [1., 0., 0.],
       [0., 1., 0.],
       [0., 0., 1.]], dtype=float32)

>>> # For an interaction with a covariate
>>> position = np.arange(x.shape[0])
>>> pos_basis = nmo.basis.RaisedCosineLinearEval(5)
>>> interaction = pos_basis * trial_type_basis
>>> X = interaction.compute_features(position, trial_types)
Array([[1.        , 0.        , 0.        , 0.49999997, 0.        ,
        0.        , 0.        , 0.        , 0.        , 0.        ,
        0.        , 0.        , 0.        , 0.        , 0.        ],
       [0.24999997, 0.        , 0.        , 0.9330127 , 0.        ,
        0.        , 0.75      , 0.        , 0.        , 0.06698731,
        0.        , 0.        , 0.        , 0.        , 0.        ],
       [0.        , 0.        , 0.        , 0.        , 0.06698725,
        0.        , 0.        , 0.74999994, 0.        , 0.        ,
        0.9330127 , 0.        , 0.        , 0.25000006, 0.        ],
       [0.        , 0.        , 0.        , 0.        , 0.        ,
        0.        , 0.        , 0.        , 0.        , 0.        ,
        0.        , 0.49999997, 0.        , 0.        , 1.        ]],      dtype=float32)

```

Currently, a way to get to this behavior is to define a custom basis:

```python
>>> n_categories = 3
>>> def func(x):
...     return jax.nn.one_hot(x, n_categories)
>>>  trial_type_basis = nmo.basis.CustomBasis(func)

```

However, handling label encoding/decoding would be up to the user, and requires substantial of boiler plait code to define the custom basis.

**Describe alternatives you've considered**
In #473, we discussed having a dedicated basis that computes the interaction.

```python
import jax.numpy as jnp
import jax
import nemos as nmo

# Receives as input the categories and the interaction
interaction = nmo.basis.CategoryInteraction(n_categories, basis)

X1 = interaction.compute_features(categories, x)
# internally what this does is
X2 = jnpeinsum(
    "tk, tc -> tck", 
    basis.compute_features(x),
    jax.nn.one_hot(categories, n_categories)
)
assert jnp.all(X1 == X2)
```

I think this is a common use case, but it is equivalent to having a basis handling the categories 1-hot encoding and using multiplication to capture interactions. I think this would be not too bad in terms of how much extra code is needed, and also fairly readable.

**Additional context**
At this point, the name basis starts to become a stretch This `basis` are mostly input maps. Transformer is a name that is already taken, so we may want to think about if it would make sense to call this a basis or having different categories of transformations, of which basis are one.

BalzaniEdoardo:
Closes #486

## Summary

- Introduces `CategoryBasis` / `Category`, a new basis class for one-hot encoding categorical variables.
- Generalizes the bounds guard from `CategoryBasis` to `BasisMixin` via `_is_discrete`, so any future discrete basis gets the behaviour automatically.
- Adds a comprehensive test suite covering all public API paths of `CategoryBasis`, plus a new `TestDiscreteBasis` class parametrized over all `_is_discrete` classes.

## Motivation

Discrete task variables (trial type, stimulus identity, neuron id) are common in neuroscience GLMs but had no first-class support in NeMoS. Users had to manually one-hot encode outside the pipeline, breaking the `compute_features` / `sklearn` integration. `CategoryBasis` fills this gap while fitting naturally into the existing basis composition framework (issue #486).

### New class: `CategoryBasis` / `Category` (`src/nemos/basis/_category.py`)

| Feature | Detail |
|---|---|
| `categories` | Accepts `int`, `list`, `NDArray`, or JAX array; stored sorted; `n_basis_funcs` updates on re-assignment |
| `out_of_category` | `bool`; if `True`, unknown labels encode as all-zero rows; if `False`, raises `ValueError` on unknown labels |
| JIT | Supported only when `out_of_category=True`; raises a clear `ValueError` under `jax.jit` otherwise (see rationale below) |
| `evaluate_on_grid` | Raises `NotImplementedError` — method is meaningless for discrete inputs |
| `_is_discrete = True` | Marks the class as discrete; used by `BasisMixin` to guard bounds access |

**JIT compatibility.** When `out_of_category=False`, `evaluate` must inspect the concrete values of the input array to detect unknown labels and raise. Under `jax.jit`, arrays are abstract tracers — their values are unavailable at trace time — making this check impossible. `out_of_category=True` sidesteps the issue by replacing unknown labels with `-1` via `jnp.where` / `np.isin`, which are JAX-traceable operations that never require concrete values.

**`_set_out_of_category` dtype dispatch.** The method dispatches across three dtype paths to correctly identify out-of-category labels regardless of whether categories and inputs are numeric or string-typed.

### Bounds guard generalized to `BasisMixin` (`src/nemos/basis/_basis_mixin.py`)

`__getattr__` and `__setattr__` overrides were moved from `CategoryBasis` to `BasisMixin`, conditioned on `self._is_discrete`. The error message uses `self.__class__.__name__` so it remains informative for any future discrete subclass. The guard intentionally accesses `_is_discrete` directly (no `getattr` fallback) so that a basis missing the `BasisMixin` contract fails loudly.

## Questions
- The adding one or more categorical predictors creates perfect linear dependencies in the design matrix rows. Adding regularizers can make the problem identifiable, however, the identifiability problem is very nuanced. I added a long detail note as part of the docstrings, but in the how-to-guide I focused on simple operational recommendations. Is this the best way to present the issue to the user? Is the note in the docstrings visible enough?
- Currently we implement a very simple dummy coding, I think this is what our basis should provide, any more complex coding requires a careful algorithmic implementation. I think we should point to `patsy` or `formulaic` for more complex scenarios. Does this make sense?
- Should we support a `pynapple` compatible formula based design? `nmo.from_formula("x+y+x:y", tsdframe)`, so that people can drop pynaple directly. This basically just wraps `patsy` or `formulaic`?


docs/how_to_guide/categorical_identifiability.md
billbrod, line 1: I wonder if this notebook should be merged with the categorical_predictors one. It's basically answering the question: "why should I drop a column?", so it strikes me as unlikely that someone would come directly to this notebook. And, as written, it's a bit hard to read by itself. Or, at the very least, it shouldn't be at the same importance level as the categorical predictions one. You could do nesting (like how I do for the [texture model](https://docs.plenoptic.org/docs/branch/main/user_guide/models_and_metrics/portilla_simoncelli/ps_index.html)), to make it clear that these two are related and while the first one should be read by anyone who wants to use a categorical feature, the second one is really just to answer one specific question.

docs/how_to_guide/categorical_identifiability.md
```diff
@@ -0,0 +1,133 @@
...
+  name: python3
+---
+
+(categorical_identifiability)=
+# Resolving Redundancy in Categorical Design Matrices
+
+## Why Does Redundancy Arise?
+
+When the [`Category`](nemos.basis.Category) basis is used as a standalone main-effect predictor
+together with a NeMoS GLM (which always includes an intercept), there are infinitely many coefficient vectors that produce the same linear predictor $X \cdot \mathbf{w} + c$. Let's see why:
```
billbrod, line 20: this jumps right into math notation, which we don't often use in our docs. maybe some introduction before this section that says something like, "in this notebook, we're discussing a basic issue with categorical predictors in linear (and related) models. Thus, we'll discuss this problem using the following notation..." (and also link back to [this page](https://nemos--527.org.readthedocs.build/en/527/background/plot_00_conceptual_intro.html#glm-components), which reminds them of the relationship between this notation and the GLM)

docs/how_to_guide/categorical_identifiability.md
```diff
@@ -0,0 +1,133 @@
...
+
+```{code-cell} ipython3
+c = -1
+w = np.array([1, 1])
+
+print("X @ w + c:", X @ w + c)
+```
+
+Stacking $c$ and $\mathbf{w}$ into a single vector $\mathbf{v} = [c, \mathbf{w}]$ and prepending
+an all-ones column to $X$ gives the compact form $X_\text{aug} \cdot \mathbf{v} = 0$:
```
billbrod, line 45: should say something like, "this representation is equivalent to the above, but more convenient for our purposes in this notebook"

docs/how_to_guide/categorical_identifiability.md
```diff
@@ -0,0 +1,133 @@
...
+category = np.array(["L", "L", "L", "L", "R", "R", "R", "R"])
+cat_basis = nmo.basis.Category(["L", "R"])
+X = cat_basis.compute_features(category)
+
+print("One-hot encoding:\n", X)
+print("Sum over columns:", X.sum(axis=1))
+```
+
+Because the one-hot columns sum to a vector of ones, we can find coefficients $\mathbf{w}$ and
+intercept $c$ for which $X \cdot \mathbf{w} + c = 0$:
```
billbrod, line 35:
```suggestion
Because the columns of our design matrix sum to a vector of ones, we can find coefficients $\mathbf{w}$ and intercept $c$ for which $X \cdot \mathbf{w} + c = 0$:
```

docs/how_to_guide/categorical_identifiability.md
```diff
@@ -0,0 +1,133 @@
...
+
+```{code-cell} ipython3
+c = -1
+w = np.array([1, 1])
+
+print("X @ w + c:", X @ w + c)
+```
+
+Stacking $c$ and $\mathbf{w}$ into a single vector $\mathbf{v} = [c, \mathbf{w}]$ and prepending
```
billbrod, line 44:
```suggestion
This is bad! It means that our design matrix has a [null space](https://en.wikipedia.org/wiki/Kernel_(linear_algebra)), vectors we can add to the parameters that have no effect on the model predictions. This is what gives rise to the problem: our model cannot distinguish between different sets of parameters if their only differences lie in the null space. This both makes optimization more difficult and prevents us from interpreting the parameters.

Let's unpack this a bit more.

Stacking $c$ and $\mathbf{w}$ into a single vector $\mathbf{v} = [c, \mathbf{w}]$ and prepending
```

or something. I just think we need a bit more explanation before jumping into it

billbrod: the information is all here in this section, I just think it needs to be fore-fronted a bi tmore

docs/how_to_guide/categorical_identifiability.md
```diff
@@ -0,0 +1,133 @@
...
+alpha = 5.0                        # arbitrary shift along the null direction
+
+print("X_aug @ params:             ", X_aug @ params)
+print("X_aug @ (params + alpha*v): ", X_aug @ (params + alpha * v))
+```
+
+Models with parameters $[c, \mathbf{w}]$ and $[c, \mathbf{w}] + \alpha \cdot \mathbf{v}$ predict the same firing rate for any $\alpha \in \mathbb{R}$.
+This is **non-identifiability**: the data alone cannot distinguish between them, and there is no unique optimal solution.
+
+A practical way to detect this redundancy is to check whether the rank of `X_aug = [1 | X]` is strictly less than number of its columns.
```
billbrod, line 67: haven't explained the `[1|X]` notation yet, should mention it when we define `X_aug`

billbrod: also, shouldn't that be latex, not monospace?

docs/how_to_guide/categorical_identifiability.md
```diff
@@ -0,0 +1,133 @@
...
+**The rank is the number of linearly independent columns**, so if it is smaller than the total number of columns, some columns must be redundant. In that case, a non-trivial null space exists, i.e., there are non-zero vectors $\mathbf{v}$ such that $X_\text{aug} \cdot \mathbf{v} = 0$:
```
billbrod, line 68: link to something about rank / linear independency here?

docs/how_to_guide/categorical_identifiability.md
```diff
@@ -0,0 +1,133 @@
...
+
+```{code-cell} ipython3
+print(f"Rank of [1 | X]:   {np.linalg.matrix_rank(X_aug)}")
+print(f"Number of columns: {X_aug.shape[1]}")
+# rank < n_cols  →  null space exists  →  model is non-identifiable
```
billbrod, line 73: I think this comment belongs in the text

docs/how_to_guide/categorical_identifiability.md
```diff
@@ -0,0 +1,133 @@
...
+```
+
+## Reference Coding: A Simple Way to Recover Identifiability
+
+Drop one column from the full encoding. The retained columns become *contrasts* against the
```
billbrod, line 78:
```suggestion
One way to solve this problem, known as "reference coding", is to drop one column from the full encoding. The retained columns become *contrasts* against the
```

docs/how_to_guide/categorical_identifiability.md
```diff
@@ -0,0 +1,133 @@
...
+
+```{code-cell} ipython3
+X_ref = X[:, 1:]  # drop "L"; the remaining column codes "R" vs "L"
+X_ref_aug = np.column_stack([np.ones(len(category)), X_ref])
+# rank == n_cols → full rank
+print(f"Rank of [1 | X_ref]: {np.linalg.matrix_rank(X_ref_aug)}")
+print(f"Number of columns:   {1 + X_ref.shape[1]}")
+```
+
+The reference level is arbitrary from a model-fit perspective, but determines coefficient
```
billbrod, line 89:
```suggestion
Which column you drop is arbitrary from a model-fit perspective, but determines coefficient
```

docs/how_to_guide/categorical_predictors.md
```diff
@@ -11,162 +11,143 @@ kernelspec:
...
+counts  = np.array([10, 5, 10, 0])
+
+# Category * continuous basis: one set of basis functions per category
+bas = nmo.basis.Category(["L", "R"]) * nmo.basis.RaisedCosineLinearEval(3)
+X = bas.compute_features(context, speed)
+print("X.shape: ", X.shape)  # (4, 6): 3 basis functions × 2 categories
+
+```
+
+## Standalone Categorical Predictors: Why do we recommend dropping a column?
```
billbrod, line 40: I feel like this section should be first, since using standalone category predictors seems "more basic" than the multiplication

docs/how_to_guide/categorical_predictors.md
```diff
@@ -11,162 +11,143 @@ kernelspec:
   name: python3
 ---
 
+(categorical_design_matrices)=
 # Construct Design Matrices for Categorical Features
 
-## Design Matrix with `patsy`
+## Splitting a Continuous Variable by Category
```
billbrod, line 17: maybe add an introductory sentence above this like: "There are two main supported uses for `Category` in nemos: using the category as a standalone predictor / main effect, and multiplying it by a continuous basis to estimate category-specific tuning curves. We'll show both below."

docs/how_to_guide/categorical_predictors.md
```diff
@@ -11,162 +11,143 @@ kernelspec:
...
 :::
 
-If you have categorized your trials and want to capture a change in the firing rate with the trial category, you can create your design matrix using [`patsy`](https://patsy.readthedocs.io/en/latest/).
-
-Let's assume we have a simple dataset with four samples and two labels assigned to each sample. You can think of these labels as characteristics of the trial to which the samples belong, such as stimulus 1 vs stimulus 2 or context 1 vs context 2.
-
-To make it concrete, let's assume that the stimulus can be the image of a red triangle (Tri) a blue square (Sq), and the context defines the task, for example report either the color (C) or the shape (S) of the stimulus.
+For designs involving multiple categorical variables, higher-order interactions,
+or non-default contrast coding (sum-to-zero, Helmert, etc.), use
+[`patsy`](https://patsy.readthedocs.io) or [`formulaic`](https://matthewwardrop.github.io/formulaic/)
```
billbrod, line 80: add tab to show how to do this in either patsy or formulaic? right now we mention formulaic but don't show how to use it

docs/how_to_guide/categorical_predictors.md
```diff
@@ -11,162 +11,143 @@ kernelspec:
...
-
-We are now ready to fit a GLM model using the 1-hot encoded categories as predictor. The model will learn a different
-firing rate for each condition.
+Full one-hot encoding of each term in the formula — the two categorical variables and their interaction — would have produced 8 columns, 4 of which would be redundant. `patsy` detects and drops all redundant columns automatically, guaranteeing that model coefficients are identifiable.
 
 ```{code-cell} ipython3
-import nemos as nmo
-
-# Fit the GLM model
 model = nmo.glm.GLM().fit(design_df, counts)
```
billbrod, line 120: this fit doesn't converge, which seems bad in an example. increase iterations or use float64?

docs/how_to_guide/categorical_predictors.md
```diff
@@ -11,162 +11,143 @@ kernelspec:
...
-:::{note}
 
-You can directly pass a [`pandas.DataFrame`](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.html) to the
-`fit` method, as long as it can be internally converted to a floating-point array (that means, for example, that you can't have a column whose values are strings). The conversion is handled automatically, by calling `design_df.values`,
-so no additional steps are required on your part.
-:::
+## NeMoS `Category` vs `patsy`
 
-## Categorical Predictors and Basis Functions
+NeMoS [`Category`](nemos.basis.Category) basis provides a simple dummy coding of categorical variables. This is just one of the many encoding schemes that `patsy` provides.
```
billbrod, line 126: I think this should be merged with the previous section.

docs/how_to_guide/README.md
```diff
@@ -151,6 +151,16 @@ categorical_predictors.md
 
 :::{grid-item-card}
 
+```{toctree}
```
billbrod, line 154: picture for `categorical_predictors` in readme no longer shows up in that notebook

src/nemos/basis/_category.py
```diff
@@ -0,0 +1,207 @@
...
+
+@support_pynapple(conv_type="jax")
+def one_hot_encoding(
+    x: NDArray | Tsd | TsdFrame | TsdTensor, n_classes: int
+) -> ArrayLike | Tsd:
+    """One-hot encode with pynapple support."""
+    return jax.nn.one_hot(x, n_classes)
+
+
+class CategoryBasis(AtomicBasisMixin, Basis):
```
billbrod, line 33: why is this intermediate class necessary? for the other basis functions, it's to better support both Eval and Conv, but this one only supports Eval

src/nemos/basis/_category.py
```diff
@@ -0,0 +1,207 @@
...
+        )
+
+    @property
+    def out_of_category(self) -> int | str | float | None:
+        return self._out_of_category
+
+    @out_of_category.setter
+    def out_of_category(self, out_of_category: bool):
+        if not isinstance(out_of_category, bool):
+            raise TypeError("``out_of_category`` must be a boolean (True or False).")
```
billbrod, line 93:
```suggestion
            raise TypeError("``out_of_category`` must be a boolean.")
```

src/nemos/basis/basis.py
```diff
@@ -2890,3 +2891,173 @@ def evaluate(self, *sample_pts: NDArray) -> NDArray:
...
+    Parameters
+    ----------
+    categories :
+        The set of valid category labels. Accepted forms:
+
+        - ``int``: interpreted as the number of categories; labels default to
+          ``[0, 1, ..., categories-1]``.
+        - ``list`` or ``NDArray``: the explicit list of unique category labels. Note
+        that the category labels will be sorted internally. Column ``i`` of the one-hot
+        encoding will correspond to ``basis.categories[i]``.
```
billbrod, line 2914:
```suggestion
        - ``list`` or ``NDArray``: the explicit list of unique category labels. Note
          that the category labels will be sorted internally. Column ``i`` of the one-hot
          encoding will correspond to ``basis.categories[i]``.
```

indent not rendered correctly

src/nemos/basis/basis.py
```diff
@@ -2890,3 +2891,173 @@ def evaluate(self, *sample_pts: NDArray) -> NDArray:
...
+    >>> context = np.random.choice(["L", "R"], size=20)
+    >>> bas = Category(["L", "R"]) * RaisedCosineLinearEval(5)
+    >>> X = bas.compute_features(context, speed)
+
+    """
+
+    def __init__(
+        self,
+        categories: List | NDArray | int,
+        out_of_category: bool = True,
```
billbrod, line 2974: this needs to be documented in docstring

src/nemos/basis/basis.py
```diff
@@ -2890,3 +2891,173 @@ def evaluate(self, *sample_pts: NDArray) -> NDArray:
...
+    When ``Category`` is multiplied with a continuous basis (the recommended
+    use), the intercept is not involved and no column needs to be dropped.
+
+    For a detailed discussion of identifiability, reference-level choice,
+    and the effect of regularization, see the
+    :ref:`identifiability guide <categorical_identifiability>`.
+
+    Examples
+    --------
+    Encode a categorical variable with 3 integer labels:
```
billbrod, line 2941: should say something about column dropping, like "Remember, if you wanted to use this as a categorical predictor, you'd need to drop a column!" Or show that

src/nemos/basis/basis.py
```diff
@@ -2890,3 +2891,173 @@ def evaluate(self, *sample_pts: NDArray) -> NDArray:
...
+        return super().evaluate(xi)
+
+    @add_docstring("_compute_features", EvalBasisMixin)
+    def compute_features(self, xi: ArrayLike) -> FeatureMatrix:
+        """
+        Examples
+        --------
+        >>> import numpy as np
+        >>> from nemos.basis import Category
+
```
billbrod, line 3009:
```suggestion
```

src/nemos/basis/basis.py
```diff
@@ -2890,3 +2891,173 @@ def evaluate(self, *sample_pts: NDArray) -> NDArray:
...
+        Array([[1., 0., 0.],
+               [0., 1., 0.],
+               [0., 0., 1.],
+               [1., 0., 0.]], dtype=float...)
+        """
+        # ruff: noqa: D205, D400
+        return super().evaluate(xi)
+
+    @add_docstring("_compute_features", EvalBasisMixin)
+    def compute_features(self, xi: ArrayLike) -> FeatureMatrix:
```
billbrod, line 3003: these two examples (compute_features and evaluate) are the same. should the docstrings mention that they're the same or why you'd use one over the other?

src/nemos/basis/_category.py
```diff
@@ -0,0 +1,207 @@
...
+                "To enable JIT compilation, please set ``out_of_category=True``."
+            )
+        else:
+            encoded = self._label_encoder.encode(xi, safe=False)
+            # set -1 to out-of category-labels
+            # one_hot_encoding will assign 0s where encoded == -1
+            encoded = self._set_out_of_category(xi, encoded)
+        return one_hot_encoding(encoded, self._label_encoder.n_classes)
+
+    def evaluate_on_grid(self, *n_samples: int) -> Tuple[Tuple[NDArray], NDArray]:
```
billbrod, line 201: should this be here? it's weird that it ends up in the API docs -- would seem more natural to me to be undocumented (if it needs to be implemented)

billbrod (changes requested):
Comments largely about the explanation in the docs. Didn't have time to look at tests.

docs/how_to_guide/categorical_identifiability.md
BalzaniEdoardo: I applied the suggestion, I like the emphasis on the practical issue too

src/nemos/basis/basis.py
BalzaniEdoardo: Evaluate doesn't flatten to a 2D array, that's the only difference in the Eval basis, so generally, it can take in a higher dimensional array. I'll show that in an example

src/nemos/basis/_category.py
BalzaniEdoardo: that is true, I just used the same pattern for most bases, including Fourier that is only eval. I can change that

src/nemos/basis/_category.py
BalzaniEdoardo: This is a difficult one, I am trying a different solution, let me know what you think.

I implemented it concretely, but the only thing that makes sense is to return a value per category (return `np.eye(n_categories)`). 

Then the question is what to do if user pass an integer to `evaluate_on_grid` that do not match the num of categories? I currently opted for raising because it may result in weird error if one does
```python
>>> bas = Category(3)
>>> _, X = bas.evaluate_on_grid(50) 
>>> X.shape # one may expect 50 samples
(3, 3)
```

BalzaniEdoardo (commented):
@billbrod can you take another look, especially on how I dealt with the `evaluate_on_grid`?

docs/how_to_guide/categorical_predictors/categorical_identifiability.md
```diff
@@ -64,19 +74,19 @@ print("X_aug @ (params + alpha*v): ", X_aug @ (params + alpha * v))
...
 This is **non-identifiability**: the data alone cannot distinguish between them, and there is no unique optimal solution.
 
-A practical way to detect this redundancy is to check whether the rank of `X_aug = [1 | X]` is strictly less than number of its columns.
-**The rank is the number of linearly independent columns**, so if it is smaller than the total number of columns, some columns must be redundant. In that case, a non-trivial null space exists, i.e., there are non-zero vectors $\mathbf{v}$ such that $X_\text{aug} \cdot \mathbf{v} = 0$:
+A practical way to detect this redundancy is to check whether the rank of $X_\text{aug}$ is strictly less than number of its columns.
+[**The rank is the number of linearly independent columns**](https://en.wikipedia.org/wiki/Rank_(linear_algebra)?utm_source=chatgpt.com), so if it is smaller than the total number of columns, some columns must be redundant. In that case, a non-trivial null space exists, i.e., there are non-zero vectors $\mathbf{v}$ such that $X_\text{aug} \cdot \mathbf{v} = 0$:
```
billbrod:
```suggestion
[**The rank is the number of linearly independent columns**](https://en.wikipedia.org/wiki/Rank_(linear_algebra)), so if it is smaller than the total number of columns, some columns must be redundant. In that case, a non-trivial null space exists, i.e., there are non-zero vectors $\mathbf{v}$ such that $X_\text{aug} \cdot \mathbf{v} = 0$:
```

docs/how_to_guide/categorical_predictors/categorical_identifiability.md
```diff
@@ -64,19 +74,19 @@ print("X_aug @ (params + alpha*v): ", X_aug @ (params + alpha * v))
...
 
 ```{code-cell} ipython3
 print(f"Rank of [1 | X]:   {np.linalg.matrix_rank(X_aug)}")
 print(f"Number of columns: {X_aug.shape[1]}")
-# rank < n_cols  →  null space exists  →  model is non-identifiable
 ```
 
+As we can see, the rank is less than the number of columns, and therefore there is a non-empty null space and the model is non-identifiable.
```
billbrod:
```suggestion
As we can see, in this example, the rank is less than the number of columns, and therefore there is a non-empty null space and the model is non-identifiable.
```

docs/how_to_guide/categorical_predictors/categorical_identifiability.md
billbrod:
```suggestion
See [](categorical_design_matrices) for worked
```

let sphinx fill in the title automatically -- this one is now out of date

docs/how_to_guide/categorical_predictors/categorical_predictors.md
```diff
@@ -57,6 +44,21 @@ For a detailed discussion of identifiability and the effect of regularization,
 see [Identifiability of Categorical Predictors](categorical_identifiability).
```
billbrod:
```suggestion
see [](categorical_identifiability).
```

docs/how_to_guide/categorical_predictors/categorical_predictors.md
```diff
@@ -57,6 +44,21 @@ For a detailed discussion of identifiability and the effect of regularization,
...
 :::
 
+## Splitting a Continuous Variable by Category
+
+The primary use of the [`Category`](nemos.basis.Category) basis in NeMoS is to estimate category-specific tuning curves by multiplying it with a continuous basis.
```
billbrod:
```suggestion
The [`Category`](nemos.basis.Category) basis in NeMoS also allows you to estimate category-specific tuning curves by multiplying it with a continuous basis.
```

feels weird to say "primary" here when it's the second thing on the page

docs/how_to_guide/categorical_predictors/categorical_predictors.md
```diff
@@ -57,6 +44,21 @@ For a detailed discussion of identifiability and the effect of regularization,
...
+
+Continuing the previous example, let's assume that we have also recorded the average animal speed per trial and suppose we want to learn how the neuron responds to speed depending on the turn side. You can use the `Category` basis to produce an appropriate design matrix:
```
billbrod:
```suggestion
Continuing the previous example, let's assume that we have also recorded the average animal speed per trial and suppose we want to learn how the neuron responds to speed depending on the turn side. You can multiply the `Category` basis by another basis to produce an appropriate design matrix:
```

docs/how_to_guide/categorical_predictors/categorical_predictors.md
```diff
@@ -96,7 +99,12 @@ design_df = dmatrix(formula, data, return_type="dataframe")
 
 # patsy adds an intercept; drop it since NeMoS GLMs include one implicitly
 design_df = design_df.drop(columns=["Intercept"])
-design_df
+print("patsy:\n\n", design_df)
+
+# formulaic equivalent
+design_df = model_matrix(formula, data)
+design_df = design_df.drop(columns=["Intercept"])
+print("\n\nformulaic:\n\n", design_df)
```
billbrod: still think this should be separate tabs

docs/how_to_guide/categorical_predictors/categorical_predictors.md
```diff
@@ -96,7 +99,12 @@ design_df = dmatrix(formula, data, return_type="dataframe")
 
 # patsy adds an intercept; drop it since NeMoS GLMs include one implicitly
 design_df = design_df.drop(columns=["Intercept"])
-design_df
+print("patsy:\n\n", design_df)
+
+# formulaic equivalent
+design_df = model_matrix(formula, data)
+design_df = design_df.drop(columns=["Intercept"])
```
billbrod, line 131: should add same note about dropping the intercept column here

docs/how_to_guide/categorical_predictors/categorical_predictors.md
billbrod, line 175: link to the technical note or have a dropdown computing the rank to show people how they can check this

docs/how_to_guide/categorical_predictors/categorical_predictors.md
```diff
@@ -84,6 +86,7 @@ support a wide range of coding schemes.
 ```{code-cell} ipython3
 import pandas as pd
 from patsy import dmatrix
+from formulaic import model_matrix
```
billbrod: I think these should be just `import`, so it's clear when called later which package these are coming from (e.g., `patsy.dmatrix` instead of just `dmatrix`)

docs/how_to_guide/categorical_predictors/categorical_predictors.md
```diff
@@ -120,9 +128,6 @@ Full one-hot encoding of each term in the formula — the two categorical variab
 model = nmo.glm.GLM().fit(design_df, counts)
 ```
 
-
-## NeMoS `Category` vs `patsy`
-
 NeMoS [`Category`](nemos.basis.Category) basis provides a simple dummy coding of categorical variables. This is just one of the many encoding schemes that `patsy` provides.
 
 For example, the dummy code for one categorical predictor in NeMoS,
```
billbrod: I would drop "dummy code" here and below. we don't explain what it means and it's not necessary here

docs/how_to_guide/categorical_predictors/categorical_predictors.md
```diff
@@ -57,6 +44,21 @@ For a detailed discussion of identifiability and the effect of regularization,
...
+```{code-cell} ipython3
+speed = np.array([10., 3., 2., 20.])
+
+# Category * continuous basis: one set of basis functions per category
+bas = nmo.basis.Category(["L", "R"]) * nmo.basis.RaisedCosineLinearEval(3)
+X = bas.compute_features(turn_side, speed)
+print("X.shape: ", X.shape)  # (4, 6): 3 basis functions × 2 categories
+```
+
 ## Complex Designs
```
billbrod:
```suggestion

(complex-designs)=
## Complex Designs with `patsy` and `formulaic`
```

so we can link to this elsehwere

docs/how_to_guide/categorical_predictors/README.md
```diff
@@ -0,0 +1,18 @@
+# Categorical Predictors
+
+Capturing the effect of stimulus identity, behavioral choices, etc., are all common examples of model designs requiring the encoding of categorical predictors. In this note we will explore how to construct such designs with NeMoS `Categorical` basis. It also covers specialized packages ([`patsy`](https://patsy.readthedocs.io) or [`formulaic`](https://matthewwardrop.github.io/formulaic/))  which elegantly handle complex schemes involving multiple categorical predictors and diverse encoding methods.
+
+These packages transparently resolve a well-known identifiability issue: designs with multiple categorical predictors can result in non-identifiable models that yield multiple equivalent solutions. We expand on this topic in a dedicated technical note for readers who want to explore the problem in greater depth.
```
billbrod:
```suggestion
Capturing the effect of stimulus identity, behavioral choices, etc., are all common examples of model designs requiring the encoding of categorical predictors. In this section we will explore how to construct such designs with NeMoS `Categorical` basis. 

We [also cover specialized NeMoS-compatible packages](complex-designs) which elegantly handle complex schemes involving multiple categorical predictors and diverse encoding methods. These packages transparently resolve a well-known identifiability issue: designs with multiple categorical predictors can result in non-identifiable models that yield multiple equivalent solutions. 

We expand on the issue of identifiability in a dedicated [technical note](categorical_identifiability) for readers who want to explore the problem in greater depth.
```

docs/how_to_guide/categorical_predictors/README.md
```diff
@@ -0,0 +1,18 @@
...
+
+[//]: # (Format "Contents" as a header level 2 without including it in the toctree, so that it doesn't render in the card)
```
billbrod: what card?

src/nemos/basis/_category.py
```diff
@@ -46,18 +47,70 @@ class CategoryBasis(AtomicBasisMixin, Basis):
...
         - ``int``: interpreted as the number of categories; labels default to
           ``[0, 1, ..., categories-1]``.
-        - ``list`` or ``NDArray``: the explicit list of unique category labels.
-          When a list is provided, it is converted to an ``NDArray`` via
-          ``np.asarray``. Mixed-type lists will be cast to a common dtype
-          (e.g., ``["a", 1]`` becomes ``array(['a', '1'], dtype='<U21')``).
-
-    out_of_category:
+        - ``list`` or ``NDArray``: the explicit list of unique category labels. Note
+          that the category labels will be sorted internally. Column ``i`` of the
```
billbrod: what does "sorted internally" mean for the user? that this needs to be passed as sorted or unexpected behavior results?

src/nemos/basis/_category.py
```diff
@@ -46,18 +47,70 @@ class CategoryBasis(AtomicBasisMixin, Basis):
...
+    Encode a categorical variable with 3 integer labels:
+
+    >>> import numpy as np
+    >>> from nemos.basis import Category
+    >>> basis = Category(3)
+    >>> basis.n_basis_funcs
+    3
+    >>> labels = np.array([0, 1, 2, 0])
+    >>> features = basis.compute_features(labels)
+    >>> features.shape
```
billbrod: maybe show `features` too? since it's so small

src/nemos/basis/_category.py
```diff
@@ -46,18 +47,70 @@ class CategoryBasis(AtomicBasisMixin, Basis):
...
+    (4, 3)
+
+    Standalone categorical predictor with reference coding (drop one column):
+
+    >>> basis = Category(["Tri", "Sq"])
+    >>> X = basis.compute_features(np.array(["Tri", "Sq", "Tri", "Sq"]))
+    >>> X = X[:, 1:]  # "Tri" is the reference; remaining column is the "Sq" contrast
```
billbrod:
```suggestion
    >>> X = X[:, 1:]  # "Tri" is the reference; remaining column is the contrast between "Sq" and "Tri"
```

src/nemos/basis/_category.py
```diff
@@ -46,18 +47,70 @@ class CategoryBasis(AtomicBasisMixin, Basis):
...
+    >>> basis.n_basis_funcs
+    3
+    >>> labels = np.array([0, 1, 2, 0])
+    >>> features = basis.compute_features(labels)
+    >>> features.shape
+    (4, 3)
+
+    Standalone categorical predictor with reference coding (drop one column):
+
+    >>> basis = Category(["Tri", "Sq"])
```
billbrod: Why not use "L" and "R" here? "Tri" and "Sq" are unclear

src/nemos/basis/_category.py
```diff
@@ -182,6 +234,31 @@ def evaluate(self, xi: ArrayLike | Tsd | TsdFrame | TsdTensor) -> FeatureMatrix:
...
         ValueError
             If any label in ``xi`` is not in the set of known categories.
+
+        Notes
+        -----
+        The `evaluate` method returns an array of shape ``(*xi.shape, n_basis_funcs)``.
+        The method preserves the input shape and appends an extra basis axis.
+
+        Examples
+        --------
```
billbrod, line 253: maybe be useful to add a sentence here interpreting the output? (and in class docstring probably)

e.g., "Since our input had 2 rows and 4 columns, our output has ... and a 1 in entry i,j means that ..."

src/nemos/basis/_category.py
```diff
@@ -198,10 +275,105 @@ def evaluate(self, xi: ArrayLike | Tsd | TsdFrame | TsdTensor) -> FeatureMatrix:
...
+        """
+        # ruff: noqa: D205, D400
+        return AtomicBasisMixin.set_input_shape(self, xi)
+
+    def evaluate_on_grid(self, *n_samples: int) -> Tuple[NDArray, NDArray]:
+        """Evaluate the categorical basis on its natural grid.
+
+        For a categorical basis the grid is intrinsically fixed: one point per
+        category. The ``n_samples`` argument is kept for interface compatibility
+        with continuous bases but must equal ``n_basis_funcs``.
```
billbrod:
```suggestion
        with continuous bases but must equal ``self.n_basis_funcs``.
```

billbrod (approved):
More tweaks to make this a bit easier to understand. Largely, I think this is good, and I like the change in docs structure. I don't think I need to review this again after you address these comments, but I'm happy to or to discuss any of my suggestions if it would be helpful.

PR #531, opened 2026-03-30
extract GLM-HMM params, EM, and utils into generic HMM implementations

sjvenditto:
This PR rearranges and updates the existing GLM-HMM algorithms to split out generic implementations that can be used for any HMM in a new `hmm` module, and leaving implementations specific to the GLM-HMM where it is.

### What's moved / changed:
- all functions in `expectation_maximization.py`, where functions have been updated to remove elements/calls specific to a GLM-HMM (e.g. scale, computing the rate, parameter containers)
     - `_compute_log_likelihood` function was removed because it became redundant with moving the rate computation inside of the log likelihood function
- functions specific to initial prob and transition matrix in `m_step_analytical_updates.py`
- `HMMParams` in `params.py`, and a rewrite of `GLMScale` to `GLMHMMModelParams`
- `initialize_new_session` function from `utils.py`
- other small fixes and updates to tests to make sure everything works as it did before

### What's new:
- a helper function `prepare_mstep_update_fn` in `glm_hmm/algorithm_configs.py` to make the `m_step_fn_model_params` needed for EM that combines fitting for GLM parameters and scale. This will fit them separately for most observation models (skipping Poisson, Bernoulli, and using analytical update for Gaussian), and jointly for Negative Binomial.
- a new private attribute `_separable_scale` added to observation models to facilitate the previous update 
- a generic `ModelParams` class to be inherited by each parameter container class to give default implementations of `regularizable_subtrees` and a constructor method `initialize_params`

### Outstanding questions / things to keep an eye out for:
- A default value of `None` is used for `log_scale` in the new `GLMHMMModelParams`. This in principle acts as if it is set to all zeros. This should be fine since I am not using the value of the field to make decisions about the scale, but instead attributes of the observation model (i.e. `_separable_scale` and flags in `algorithm_configs.py` when grabbing the scale update function). But we should keep an eye out to make sure this is always the case, because things might break if a scale of `None` is passed to the log likelihood function
- in some cases, the use of the old `GLMParams` still works in place of `GLMHMMModelParams` where scale is not needed.  I don't think this will cause any issues for the same reason as above: the scale attribute should only be accessed if it is used, and it is something we are only using internally for some tests
- Should the `log_likelihood_func` argument used in the EM algorithms be renamed to something like `log_emissions_probability_func`? Would this be more generic? Or is `log_likelihood_func` equally as generic? I fear that "log likelihood" might get confused as some other log likelihood of the model, but it also might be more standard terminology?
- there were some useful comments taken out from functions in `expectation_maximization.py` since they were specific to the GLM-HMM. should we instead edit them, or find a new home for them in GLM-HMM specific functions?

### Things for future PRs:
- updating validation logic
- adding in a `BaseHMM` class
- updating initialization logic


src/nemos/glm_hmm/algorithm_configs.py
```diff
@@ -7,14 +7,14 @@
 import jax.numpy as jnp
 
 from ..glm.params import GLMParams
+from ..glm_hmm.params import GLMHMMModelParams
```
BalzaniEdoardo: can't you do `from .params import GLMHMMModelParams`?

src/nemos/glm_hmm/algorithm_configs.py
```diff
@@ -197,9 +199,14 @@ def prepare_estep_log_likelihood(
     """
 
     def log_likelihood_per_sample(x, z, s):
-        return observation_model.log_likelihood(
-            x, z, scale=s, aggregate_sample_scores=lambda v: v
-        )
+        if s is None:
```
BalzaniEdoardo: would it be worth having a uniform signature for the `log_likelihood`?
```python
# for obs model with no scale
def log_likelihood(y, pred, aggregate_sample_score=...,**kwargs):
    ...

# for obs model with scale
def log_likelihood(y, pred, aggregate_sample_score=,..., scale=None, **kwargs):
    ....
```

We would not need a special if close, and the scale would be ignored if not needed. the drawback is that it could result in silent errors

src/nemos/glm_hmm/algorithm_configs.py
```diff
@@ -176,6 +177,7 @@ def posterior_weighted_glm_negative_log_likelihood_scale(
 def prepare_estep_log_likelihood(
     is_population_glm: bool,
     observation_model: Observations,
+    inverse_link_function: Callable,
```
BalzaniEdoardo, line 181: missing parameter in the docstring

src/nemos/glm_hmm/algorithm_configs.py
```diff
@@ -407,3 +431,83 @@ def scale_update_fn(current_scale, y, predicted_rate, posteriors):
...
+
+    Parameters
+    ----------
+    is_population_glm:
+        True if it is a population GLM likelihood.
+    observation_model:
+        The observation model.
+    inverse_link_function:
+        Function mapping linear predictors to rates.
+    setup_solver:
```
BalzaniEdoardo, line 464: if this comes from `BaseRegressor` which method is it? maybe worth mentioning in the param description

src/nemos/glm_hmm/algorithm_configs.py
```diff
@@ -407,3 +431,83 @@ def scale_update_fn(current_scale, y, predicted_rate, posteriors):
...
+    ):
+        scale_update_fn = get_analytical_scale_update(
+            observation_model, is_population_glm
+        )
+        # if no analytical update exists, prepare numerical optimization objective and update function
+        if scale_update_fn is None:
+            objective_scale = prepare_mstep_nll_objective_scale(
+                is_population_glm, observation_model
+            )
+            scale_update_fn = setup_solver(
```
BalzaniEdoardo, line 495: If we merge first my PR, in this specific step we could pass `solver_name="LBFGS"` and `regularizer=UnRegularized()`

src/nemos/glm_hmm/algorithm_configs.py
```diff
@@ -407,3 +431,83 @@ def scale_update_fn(current_scale, y, predicted_rate, posteriors):
...
+
+        # combine param and scale updates into a single function for use in M-step
+        def update_fn(params, X, y, posteriors):
+            new_model_params, state_params, aux_params = params_update_fn(
+                params, X, y, posteriors
+            )
+            predicted_rate = compute_rate_per_state(
+                X, new_model_params, inverse_link_function=inverse_link_function
+            )
+            new_scale, state_scale, aux_scale = scale_update_fn(
```
BalzaniEdoardo, line 510: Before we merge my other pr and add the changed described in the previous comment, this `scale_update_fn` would regularize the scale if the `setup_solver` was setting up some regularization scheme for the GLM params.

src/nemos/glm_hmm/validation.py
```diff
@@ -142,7 +166,7 @@ def check_init_and_transition_prob_shape(
             )
         return params
 
-    def check_glm_params_shape(self, params: GLMHMMUserParams) -> GLMHMMUserParams:
+    def check_model_params_shape(self, params: GLMHMMUserParams) -> GLMHMMUserParams:
```
BalzaniEdoardo, line 169: Am I wrong or I forgot to validate the scale being an array? If it is not an array, it would raise when we check `validate_consistency`, but that is a weird spot for this

src/nemos/hmm/expectation_maximization.py
```diff
@@ -7,26 +7,26 @@
 import jax
 import jax.numpy as jnp
 
-from ..glm.params import GLMParams
-from ..typing import Aux, SolverState
-from .m_step_analytical_updates import (
+from ..hmm.m_step_analytical_updates import (
```
BalzaniEdoardo:
```suggestion
from .m_step_analytical_updates import (
```

src/nemos/hmm/expectation_maximization.py
```diff
@@ -7,26 +7,26 @@
...
     _analytical_m_step_log_initial_prob,
     _analytical_m_step_log_transition_prob,
 )
-from .params import GLMHMMParams, GLMScale, HMMParams
-from .utils import Array, compute_rate_per_state, initialize_new_session
+from ..hmm.utils import Array, initialize_new_session
```
BalzaniEdoardo:
```suggestion
from .utils import Array, initialize_new_session
```

src/nemos/hmm/expectation_maximization.py
```diff
@@ -191,7 +191,87 @@ def body_fn(carry, xs):
...
+        Normalized log forward messages, shape ``(n_time_bins, n_states)``.
+        Entry ``[t, k]`` is the log filtered probability log p(z_t=k | y_1:t).
+        Each row is normalized: ``exp(log_alphas[t]).sum() == 1``.
+    log_normalizers :
+        Array of shape ``(n_time_bins,)`` containing the log-normalization constants at each
+        time step. The sum of these values gives the log-likelihood of the sequence.
+
+    See Also
+    --------
+    forward_backward : Computes both forward and backward messages for smoothing.
```
BalzaniEdoardo:
```suggestion
    :func:`~nemos.hmm.forward_backward` : Computes both forward and backward messages for smoothing.
```

src/nemos/hmm/expectation_maximization.py
```diff
@@ -191,7 +191,87 @@ def body_fn(carry, xs):
...
+    _forward_pass : Internal implementation of the forward recursion.
```
BalzaniEdoardo: would remove it, in the docs website the private method won't be available.

src/nemos/hmm/expectation_maximization.py
```diff
@@ -191,7 +191,87 @@ def body_fn(carry, xs):
...
+    -----
+    - Forward messages provide causal state estimates (no future information)
+    - Smoothing (forward + backward) provides better estimates using all data
+    - Log-space computation ensures numerical stability
+    - Session boundaries reset the recursion using initial state distribution
+
+    """
+    # unpack parameters
+    model_params = params.model_params
+    # glm_scale = params.glm_scale
```
BalzaniEdoardo:
```suggestion
```

src/nemos/hmm/expectation_maximization.py
```diff
@@ -297,15 +377,11 @@ def body_fn(carry, xs):
     return log_betas
 
 
-@partial(jax.jit, static_argnames=["inverse_link_function", "log_likelihood_func"])
+@partial(jax.jit, static_argnames=["log_likelihood_func"])
 def forward_backward(
+    params: ModelParamsT,
```
BalzaniEdoardo, line 380: I think you edited an non-final version of the `forward_backward`, it is likely fine, but I would double check how the version you edited compares with:

https://github.com/flatironinstitute/nemos/blob/glm-hmm-simulate/src/nemos/glm_hmm/expectation_maximization.py#L457

src/nemos/hmm/expectation_maximization.py
```diff
@@ -191,7 +191,87 @@ def body_fn(carry, xs):
...
+        `log_initial_prob` and `log_transition_prob`, and a `model_params` attribute
+        with parameters specific to paired model (e.g. GLM parameters for a GLM-HMM).
+    X :
+        Design matrix, shape ``(n_time_bins, n_features)``.
+    y :
+        Observations, shape ``(n_time_bins,)`` or ``(n_time_bins, n_neurons)``.
+    inverse_link_function :
+        Function mapping linear predictors to mean parameters.
+    log_likelihood_func :
+        Function computing observation log-likelihoods.
```
BalzaniEdoardo: maybe worth emphasizing `Function computing observation log-likelihoods per sample.` i.e., we do not aggregate the scores as we wood if this was an optimization objective.

src/nemos/hmm/expectation_maximization.py
```diff
@@ -324,26 +400,11 @@ def forward_backward(
...
-        and intercept of shape ``(n_states,)`` as GLMParams.
-
-    glm_scale :
-        The scale parameter of the likelihood. Shape (n_states,) or (n_neurons, n_states).
-
-    inverse_link_function :
-        Function mapping linear predictors to the mean of the observation distribution
-        (e.g., exp for Poisson, sigmoid for Bernoulli).
+    params :
+        The HMM and additional model parameters.
```
BalzaniEdoardo, line 402: we should add something like: 

```
The exact parameter structure is model dependent but it must include the attributes `hmm_params` which is a `ModelParams` subclass with attributes `log_initial_prob` and `log_transition_prob`.
```

src/nemos/hmm/expectation_maximization.py
```diff
@@ -375,51 +436,26 @@ def forward_backward(
     ----------
     .. [1] Bishop, C. M. (2006). *Pattern recognition and machine learning*. Springer.
     """
+    # unpack parameters
+    model_params = params.model_params
+    # glm_scale = params.glm_scale
```
BalzaniEdoardo:
```suggestion
```

src/nemos/hmm/expectation_maximization.py
BalzaniEdoardo: maybe worth adding some descr of the parameters here too

src/nemos/hmm/expectation_maximization.py
```diff
@@ -537,29 +562,14 @@ def run_m_step(
...
-        # Gaussian, Gamma, and other have a scale.
-        glm_scale, state_scale, _ = m_step_fn_glm_scale(
-            params.glm_scale, y, predicted_rate, posteriors
-        )
-    else:
-        # Poisson, Bernoulli etc. do not have a scale
-        # just keep carrying the scale
-        glm_scale = params.glm_scale
+    # Minimize negative log-likelihood to update GLM weights
+    optimized_projection_weights, state, _ = m_step_fn_model_params(
```
BalzaniEdoardo: this variable should probably be renamed `optimized_model_params`

src/nemos/hmm/expectation_maximization.py
```diff
@@ -624,13 +630,9 @@ def _em_step(
     params, previous_state = carry
 
     log_posteriors, log_joint_posterior, _, new_log_like, _, _ = forward_backward(
+        params,
         X,
         y,
-        params.hmm_params.log_initial_prob,
```
BalzaniEdoardo, line 629: This again shows me that you did not start from the last version of the em, worth comparing with 

https://github.com/flatironinstitute/nemos/blob/glm-hmm-simulate/src/nemos/glm_hmm/expectation_maximization.py#L689

src/nemos/hmm/expectation_maximization.py
```diff
@@ -642,68 +644,62 @@ def _em_step(
...
 
 
 def em_step(
-    params: GLMHMMParams,
-    state: GLMHMMState,
+    params: ModelParamsT,
+    state: EMState,
     X: Array,
     y: Array,
-    inverse_link_function: Callable,
```
BalzaniEdoardo, line 667: similar, i would compare this `em_step` with this

https://github.com/flatironinstitute/nemos/blob/glm-hmm-simulate/src/nemos/glm_hmm/expectation_maximization.py#L778

src/nemos/hmm/expectation_maximization.py
```diff
@@ -847,52 +831,47 @@ def stopping_condition_while(carry):
...
     return params, state
 
 
 @partial(
     jax.jit,
-    static_argnames=["inverse_link_function", "log_likelihood_func", "return_index"],
+    static_argnames=["log_likelihood_func", "return_index"],
 )
 def max_sum(
+    params: ModelParamsT,
```
BalzaniEdoardo, line 850: check with version
https://github.com/flatironinstitute/nemos/blob/glm-hmm-simulate/src/nemos/glm_hmm/expectation_maximization.py#L974

src/nemos/hmm/expectation_maximization.py
```diff
@@ -943,15 +925,15 @@ def continue_chain(omega_prev, log_em):
 
         return omega, (omega, max_prob_state)
 
-    init_omega = log_init + log_emission[0]
+    # Initialize with dummy value (won't be used since is_new_session[0] = True)
+    init_omega = jnp.full(n_states, -jnp.inf)
+
     _, (omegas, max_prob_states) = jax.lax.scan(
-        forward_max_sum, init_omega, (log_emission[1:], is_new_session[1:])
+        forward_max_sum, init_omega, (log_emission, is_new_session)
```
BalzaniEdoardo, line 935: ok, it looks like the algorithm has been updated matching the one in 

https://github.com/flatironinstitute/nemos/blob/glm-hmm-simulate/src/nemos/glm_hmm/expectation_maximization.py#L974

is this correct?

BalzaniEdoardo (changes requested):
This looks already almost ready, very nice! I left some comments, and I think we should merge #532 before this one so that you can use the fixes to make the update of the scale more robust

src/nemos/glm_hmm/validation.py
```diff
@@ -230,12 +254,15 @@ def validate_consistency(
         For single-neuron GLMHMM, only validates feature consistency with X.
         Does not validate y since it's 1D (single neuron, no neuron axis to check).
         """
-        self._glm_validator.validate_consistency(params.glm_params, X, y)
-        if params.glm_scale.log_scale.shape != params.glm_params.intercept.shape:
+        self._glm_validator.validate_consistency(params.model_params, X, y)
+        if (
+            params.model_params[1].log_scale.shape
```
BalzaniEdoardo: why does this work? can you access with a getitem?

src/nemos/hmm/expectation_maximization.py
```diff
@@ -571,14 +581,10 @@ def _em_step(
     carry: EMCarry,
     X: Array,
     y: Array,
-    inverse_link_function: Callable[[Array], Array],
     log_likelihood_func: Callable[[Array, Array, Array], Array],
-    m_step_fn_glm_params: Callable[
-        [GLMParams, Array, Array, Array], Tuple[GLMParams, SolverState]
+    m_step_fn_model_params: Callable[
+        [ModelParamsT, Array, Array, Array], Tuple[ModelParamsT, SolverState]
```
BalzaniEdoardo:
```suggestion
        [ModelParamsT, Array, Array, Array], Tuple[ModelParamsT, SolverState, Aux]
```

src/nemos/hmm/expectation_maximization.py
BalzaniEdoardo: dropped?

src/nemos/hmm/expectation_maximization.py
```diff
@@ -601,7 +607,7 @@ def _em_step(
         Function mapping linear predictors to predicted rates.
     log_likelihood_func :
         Log-likelihood function for the E-step.
-    m_step_fn_glm_params :
+    m_step_fn_model_params :
         M-step update function for GLM coefficients and intercepts.
     m_step_fn_glm_scale :
```
BalzaniEdoardo: also dropped?

src/nemos/hmm/expectation_maximization.py
BalzaniEdoardo:
```suggestion
"""Forward backward pass for a HMM."""
```

src/nemos/hmm/expectation_maximization.py
```diff
@@ -191,7 +191,87 @@ def body_fn(carry, xs):
...
+    """
+    Compute filtering probabilities (forward messages) for an HMM.
+
+    Performs the forward pass of the forward-backward algorithm, computing the
+    filtered state probabilities p(z_t | y_1:t) at each time point. These represent
+    the probability distribution over states conditioned on observations up to time t.
+
+    This is the public API for computing forward messages, useful for:
+    - Online/causal state estimation (filtering)
+    - Computing filter_proba in the GLM-HMM class
```
BalzaniEdoardo:
```suggestion
    - Computing filter_proba in the HMM class
```

src/nemos/hmm/expectation_maximization.py
```diff
@@ -537,29 +562,14 @@ def run_m_step(
...
-    if m_step_fn_glm_scale is not None:
-        # Gaussian, Gamma, and other have a scale.
-        glm_scale, state_scale, _ = m_step_fn_glm_scale(
-            params.glm_scale, y, predicted_rate, posteriors
-        )
-    else:
-        # Poisson, Bernoulli etc. do not have a scale
-        # just keep carrying the scale
-        glm_scale = params.glm_scale
+    # Minimize negative log-likelihood to update GLM weights
```
BalzaniEdoardo:
```suggestion
    # Minimize negative log-likelihood to update model parameters
```

src/nemos/hmm/expectation_maximization.py
```diff
@@ -747,28 +741,24 @@ def check_log_likelihood_increment(state: GLMHMMState, tol: float) -> Array:
...
-    m_step_fn_glm_scale: Callable | None,
+    m_step_fn_model_params: Callable,
     is_new_session: Optional[Array] = None,
     maxiter: int = 10**3,
     tol: float = 1e-8,
     check_convergence: Callable = check_log_likelihood_increment,
-) -> Tuple[GLMHMMParams, GLMHMMState]:
+) -> Tuple[ModelParamsT, EMState]:
     """
     Perform EM optimization for a GLM-HMM.
```
BalzaniEdoardo:
```suggestion
    Perform EM optimization for a HMM model (i.e. HMM or GLMHMM).
```

src/nemos/hmm/utils.py
```diff
@@ -0,0 +1,51 @@
...
+from numpy.typing import NDArray
+
+Array = NDArray | jax.numpy.ndarray
+
+
+def initialize_new_session(n_samples, is_new_session):
+    """
+    Initialize and validate session boundary indicators for HMM inference.
+
+    Ensures the session boundary array is properly formatted for GLM-HMM algorithms.
```
BalzaniEdoardo:
```suggestion
    Ensures the session boundary array is properly formatted for HMM algorithms.
```

src/nemos/hmm/params.py
```diff
@@ -0,0 +1,19 @@
+"""GLM-HMM parameter definitions and type aliases."""
```
BalzaniEdoardo:
```suggestion
"""HMM parameter definitions and type aliases."""
```

BalzaniEdoardo (commented):
extra comments

src/nemos/glm_hmm/algorithm_configs.py
sjvenditto: yes this was a mistake!

src/nemos/glm_hmm/algorithm_configs.py
sjvenditto: I actually had this at first, I know it breaks in the Gaussian case if you don't set a scale because it is used in the computation. 
- we could update the log_likelihood function in Gaussian to default to 1 if it is `None`. 
- or we change the default scale to be 0 (we would probably need a custom init method to make it the right shape)
- or we don't have a default scale and force it to be passed, where maybe the user can pass `None` and in the `to_glm_hmm_params` function we handle it there?

src/nemos/glm_hmm/algorithm_configs.py
sjvenditto: it does come from `BaseRegressor` but also technically does not exist yet (it was the updated `_instantiate_solver` method in your PR.) I can update this docstring knowing that they'll be merged close in time?

src/nemos/glm_hmm/algorithm_configs.py
sjvenditto: yes I agree

src/nemos/glm_hmm/algorithm_configs.py
sjvenditto: yes, this function really only works for the `setup_solver` in the tests right now. it would need to be updated for the output of `_instantiate_solver`, and probably the test function `setup_solver` would need to match this updated output signature as well

src/nemos/glm_hmm/validation.py
sjvenditto: oops this is an artifact of the intermediate params definition when it was a tuple

src/nemos/glm_hmm/validation.py
sjvenditto: I don't think it was checking that scale was an array before, I don't think I removed any functions in this file

src/nemos/hmm/expectation_maximization.py
sjvenditto: this is the branch I pulled it from, both `forward_backward` and `max_sum` look nearly identical besides some of my changes

src/nemos/hmm/expectation_maximization.py
sjvenditto: they look like they match to me, this is the branch I started from

src/nemos/hmm/expectation_maximization.py
sjvenditto: I started from the latest version, but the diff is comparing it to the version in development.

src/nemos/glm_hmm/algorithm_configs.py
BalzaniEdoardo: I have an idea for this:

make the vmap either:
```

    log_likelihood_per_sample = jax.vmap(
        log_likelihood_per_sample,
        in_axes=(None, state_axes, state_axes - 1),
        out_axes=state_axes,
    )
```

for obs model that need the scale. For the one that do not need the scale do:

```

    log_likelihood_per_sample = jax.vmap(
        log_likelihood_per_sample,
        in_axes=(None, state_axes),
        out_axes=state_axes,
    )
```

This will raise if we make some internal error and guard against Gaussian and Gamma using scale=None. As long as the scale is needed, it needs to be an array you can vmap over.

src/nemos/glm_hmm/validation.py
BalzaniEdoardo: yes, I think i just forgot.

BalzaniEdoardo (approved):
@sjvenditto LGTM, you can go ahead and merge this!

PR #541, opened 2026-04-02
Test speedup

BalzaniEdoardo:
## Summary

- Profiled per-test execution time and identified three dominant overhead sources: JAX JIT recompilation per unique array shape, full solver runs in tests that only check structure or routing, and Cartesian-product parametrize expansions over obs-model-agnostic logic.
- Introduced three solver-bypass fixtures in `conftest.py` and applied them selectively across `test_glm.py` and `test_pipeline.py`.
- Trimmed parametrize dimensions in `test_glm.py`, `test_transformer_basis.py`, and `test_pipeline.py`.
- Added a `--timeit` CLI flag for future profiling.
- Coverage verified against `development`: overall line coverage unchanged (92.58%); one previously uncovered line recovered with a new targeted test.

## Motivation

The full test suite took ~920 s. The top-10 tests accounted for ~48% of total time; the top-10% for ~83%. Three patterns dominated:

1. **Solver runs in structure-only tests**: many tests called `model.fit` or `model.update` only to attach fitted parameters, then checked coef_ shape, metadata, repr, or error messages — never the solution. Each call triggered JAX JIT compilation and solver iteration.
2. **Over-parametrization on the observation-model axis**: tests in `TestGLMObservationModel` inherited the class-level 6-obs-model × 2-glm-type Cartesian product even when the behaviour under test lives in `BaseGLM` and is obs-model-agnostic.
3. **Tight solver settings in fixture-heavy tests**: e.g. `test_estimate_dof_resid` called `model.fit` with `maxiter=10^5` to obtain a sparse solution, even though `_estimate_resid_degrees_of_freedom` only reads `coef_`, `intercept_`, `regularizer`, and `X`.

### Principle 1 — solver-bypass fixtures (`conftest.py`)

Three fixtures were added to `conftest.py`, with a clear hierarchy based on how much of the fit pipeline a test actually needs:

- **`mock_glm_fit`**: replaces `GLM.fit` with a pure-Python no-op that sets `coef_`/`intercept_` from X/y shapes. No JAX at all. Use when a test only needs a model in a fitted state (e.g. routing tests, `test_predict_is_fit`).
- **`mock_optimizer_run`**: runs the real `_initialize_optimizer_and_state` (preserving input validation, shape checks, metadata propagation) but replaces the solver call with a no-op that returns the initial parameters. Use when a test checks fit side-effects — coef_ tree structure, error messages, metadata — but not solution quality (e.g. `test_fit_weights_dimensionality`, `test_coef_tree_structure_after_fit_pytree_x`, `test_metadata_pynapple_fit`).
- **`mock_optimizer_update`**: same idea for `model.update` — real initialization, no-op update step. Use when a test checks update-API correctness but not that parameters changed (e.g. `test_update_n_samples`). Do not use when the test asserts on the values of the updated parameters (e.g. `test_update_params_are_finite`, `test_update_nan_drop_at_jit_comp`).

The fixture choice is always driven by test intent: tests that validate the solve must not use a mock; tests that validate the surrounding infrastructure should.

### Principle 2 — reducing parametrize dimensions

Where a test's behaviour is independent of a parametrize axis, that axis is trimmed. Two recurring patterns:

- **Obs-model axis**: tests that live conceptually in `BaseGLM` were moved out of `TestGLMObservationModel` (which inherits 6-obs-model × 2-glm-type) and re-parametrized over 2 representative models (Poisson + Gaussian). Examples: `test_update_n_samples` (96 → 32 params), `test_update_params_are_finite` (96 → 32 params, real solver retained), `test_compatibility_with_sklearn_cv` (12 → 4 params).
- **Redundant solver/shape dimensions**: axes that duplicate coverage present elsewhere are removed. Examples: `test_glm_update_consistent_with_fit_with_svrg` drops one random key and reduces epochs (64 → 8 params, −88 s); `test_optimize_solver_params` drops the pytree variant of `reg_setup` (÷2 params).

### Principle 3 — bypass `model.fit` when the tested method does not need it

Where a method only reads a subset of model state, that state is set directly rather than obtained via fit. `test_estimate_dof_resid` is the primary example: `_estimate_resid_degrees_of_freedom` only reads `coef_`, `intercept_`, `regularizer`, and `X`, so the test now assigns a fixed sparse `coef_`/`intercept_` directly and calls the method — no solver run. Result: 72 → 24 params, 14.7 s → 1.3 s.

### `tests/test_transformer_basis.py`

`test_transformer_in_pipeline`: reduced `inp` from 4 array shapes to 1 and `maxiter` from 3 to 1. Each distinct shape triggers a full JAX recompilation; the broadcasting logic is covered elsewhere. 76 → 19 params, ~−76%.

### `tests/test_pipeline.py`

All tests that verify sklearn pipeline routing (cloning, parameter setting, CV mechanics) now use `mock_glm_fit`, replacing the real solver with a pure-Python no-op. Additionally, `test_sklearn_transformer_pipeline_cv` is reduced from 5 basis types to 3.

## Before / After (Baseline Timings)

| Test | Before | After | Δ |
|------|--------|-------|---|
| `test_glm_update_consistent_with_fit_with_svrg` | 118.4 s / 64 params | ~14 s / 8 params | −88% |
| `test_transformer_in_pipeline` | 65.5 s / 76 params | ~16 s / 19 params | −76% |
| `test_optimize_solver_params` | 37.7 s / 10800 params | ~19 s / 5400 params | −50% |
| `test_update_nan_drop_at_jit_comp` | 31.4 s / 24 params | ~5 s / 4 params | −84% |
| `test_compatibility_with_sklearn_cv` | 31.1 s / 12 params | <1 s / 4 params | −97% |
| `test_update_params_are_finite` | 24.5 s / 96 params | ~11 s / 32 params | −55% |
| `test_update_n_samples` | 22.5 s / 96 params | ~3 s / 32 params | −87% |
| `test_sklearn_transformer_pipeline_cv` (pipeline) | 20.9 s / 5 params | <1 s / 3 params | −97% |
| `test_estimate_dof_resid` | 14.7 s / 72 params | 1.3 s / 24 params | −91% |

Combined, these 9 tests accounted for ~367 s (40% of total).

## Coverage

Compared `coverage_test-speedup.xml` against `coverage_development.xml`:

- **Overall**: 6193/6689 lines (92.58%) — identical to `development`.
- **`classifier_glm.py:322`**: one line lost (the `TypeError` path in `ClassifierGLM._estimate_resid_degrees_of_freedom`, no longer exercised since `test_estimate_dof_resid` bypasses fit). Recovered by the new `test_estimate_dof_resid_classifier_invalid_n_samples` (2 params, ~1 s).
- **`glm.py:776`**: one line gained (convergence fallback branch for solvers without a `.converged` attribute), now hit by `mock_optimizer_run`.

## Open Questions for Reviewers

Several tests were reduced from 6 obs models to 2 (Poisson + Gaussian) on the grounds that the behaviour under test lives in `BaseGLM`. Current coverage is unchanged, but the reduction weakens the test as a regression net: if a future change introduces obs-model-specific behaviour in one of the dropped models (e.g. Gamma, Bernoulli), the test will no longer catch it — and since coverage won't drop at the point of that change, the gap will go unnoticed unless CodeCov is checked explicitly.

The specific tests affected are `test_update_n_samples`, `test_update_params_are_finite`, `test_update_nan_drop_at_jit_comp`, and `test_compatibility_with_sklearn_cv`. Are the `update` and sklearn-compatibility code paths sufficiently obs-model-agnostic that 2 models are a safe long-term choice, or should one or two additional models be retained as sentinels?

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -1,0 +1,9 @@
+"""Register a custom option in root."""
+
+def pytest_addoption(parser):
+    parser.addoption(
+        "--timeit",
+        action="store_true",
+        default=False,
+        help="Show aggregated parametrized test durations",
+    )
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -15,6 +15,7 @@ import re
 from collections import defaultdict, namedtuple
 from copy import deepcopy
 from functools import partial
+from types import SimpleNamespace
 from typing import Literal
 
 import jax
@@ -67,6 +68,67 @@ def pytest_terminal_summary(terminalreporter):
         terminalreporter.write_line(
             f"{total:6.2f}s total  {avg:.4f}s/param  {n:>5} params  {name}"
         )
+
+
+@pytest.fixture
+def mock_glm_fit(monkeypatch):
+    """Replace GLM.fit with a pure-Python no-op that sets coef_/intercept_ from X/y shapes.
+
+    Use in tests that only care about sklearn routing (cloning, parameter
+    setting, pipeline plumbing) and do not need any fit validation logic.
+    For tests that need validation but not solver iterations, use mock_optimizer_run.
+    """
+
+    def _fit(self, X, y, init_params=None, **kwargs):
+        n_features = X.shape[1]
+        y_arr = y.d if hasattr(y, "d") else np.asarray(y)
+        if y_arr.ndim == 1:
+            self.coef_ = jnp.zeros(n_features)
+            self.intercept_ = jnp.zeros(1)
+        else:
+            self.coef_ = jnp.zeros((n_features, y_arr.shape[1]))
+            self.intercept_ = jnp.zeros(y_arr.shape[1])
+        return self
+
+    monkeypatch.setattr(nmo.glm.GLM, "fit", _fit)
+    monkeypatch.setattr(nmo.glm.PopulationGLM, "fit", _fit)
+
+
+@pytest.fixture
+def mock_optimizer_run(monkeypatch):
+    """Bypass the JAX solver in fit() while keeping all validation logic.
+
+    Patches _initialize_optimizer_and_state to inject a no-op _optimizer_run
+    that returns init_params unchanged with a converged state. Use in tests
+    that only care about pipeline routing or fit side-effects (coef_/intercept_
+    shape, validation errors) and do not check solution quality.
+    """
+    _real_init = nmo.glm.GLM._initialize_optimizer_and_state
+
+    def _patched_init(self, init_params, data, y):
+        result = _real_init(self, init_params, data, y)
+        self._optimizer_run = lambda p, d, t: (p, SimpleNamespace(converged=True), None)
+        return result
+
+    monkeypatch.setattr(nmo.glm.GLM, "_initialize_optimizer_and_state", _patched_init)
+
+
+@pytest.fixture
+def mock_optimizer_update(monkeypatch):
+    """Bypass the JAX solver step in update() while keeping all validation logic.
+
+    Patches _initialize_optimizer_and_state to inject a no-op _optimizer_update
+    that returns the current params and state unchanged. Do NOT use in tests
+    that assert parameters actually changed after calling update().
+    """
+    _real_init = nmo.glm.GLM._initialize_optimizer_and_state
+
+    def _patched_init(self, init_params, data, y):
+        result = _real_init(self, init_params, data, y)
+        self._optimizer_update = lambda p, s, d, t, *a, **kw: (p, s, None)
+        return result
+
+    monkeypatch.setattr(nmo.glm.GLM, "_initialize_optimizer_and_state", _patched_init)
 
 
 # Named tuple for model fixture returns (clearer than tuple indexing)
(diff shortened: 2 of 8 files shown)


wulfdewolf (commented):
All of this looks good and clean to me.

I don't see another solution for the limitation of only testing 2 observation models versus all of them. It's a speed<->coverage trade-off.
Either we test all and take the speed hit, or we test a couple and take the coverage hit.
Since at the moment there is no particular different behaviour, we could stick to testing just the 2 but add some clear comments somewhere saying that if observation-model specific behaviour is added, we best add those models to the tests?

Another idea would be to randomise the observation models being tested, like pick 2 out of all every time. But then you'd get nondeterministic tests...which would also be undesirable.

BalzaniEdoardo:
> Another idea would be to randomise the observation models being tested, like pick 2 out of all every time. But then you'd get nondeterministic tests...which would also be undesirable.

I agree, probably not worth doing the randomization. It is hard to control what will happen in the future, but if we assume that the glm module will be pretty stable we should be ok

PR #542, opened 2026-04-03
Initialization functions for HMMs

sjvenditto:
## Summary
Based on the initial initialization implementation of the GLM-HMM, these functions have been modified and adapted to isolate initialization of the HMM parameters only (initial probability and transition matrix). The ultimate goal is to have this called in an optional `setup` method of the HMM class, and so the user only interacts with these functions in an advanced use-case setting.

## Overview of content & changes
- **Similar:** an internally stored dictionary (`init_funcs`) will hold the initialization functions, **however** only containing the init functions `initial_proba_init` and `transition_proba_init`. 
- **New:** the dictionary will also have keys that can store user-provided keyword arguments to the functions, `initial_proba_init_kwargs` and `transition_proba_init_kwargs`. `n_states` is **not allowed** to be set here.
- **New:**  the function `setup_hmm_initialization` can be used to initialize or update `init_funcs`, with keyword arguments corresponding to each key of the `init_funcs` dictionary. This is the function that will be called by `setup`, and makes it so the user doesn't need to have knowledge of or make the internally used dictionary themselves. All the keys are optional, so the user can set whatever they want, and the rest will be filled by defaults. **Additionally,** the goal is for the `setup` method to be chainable. A user can set different initialization functions in different calls, and it will update the internal dictionary.
- **Similar:** custom callable can be passed as `initial_proba_init` and `transition_proba_init`, where we check to make sure that it has the minimum required input arguments, and the output shape and type is correct.
- **Removed:** the input parameters for the observations and design matrix have been removed as required. Required parameters for initialization functions is taken from the `InitFunctionHMM` protocol.
- **Similar:** the function `generate_hmm_initial_params` is similar to the original `glm_hmm_initialization`, returning the generated initial parameters given the dictionary of initialization functions and kwargs. **However**, there are less internal checks on the `init_params` dictionary, since it was intended to be created via `setup_hmm_initialization`. (we are only re-validating the keys themselves). Users are not necessarily meant to interact with this function either, and will likely only be called internally during `fit` where `initial_params` are not provided.
- **Removed:** I am no longer typing the initialization funcs dictionary, since we have the helper function `set_hmm_initialization` to generate the dictionary, I didn't think we needed the additional overhead on type checking user-provided input.
- **Updated:** the "uniform" initialization function for initial probability and the "uniform and "sticky" functions for the transition matrix are still being provided. A "random" initialization function has been added for both parameters
     - **Note:** the "sticky" function previously made it so the transition probability was equal to `prob_stay` when `n_states=1`. This should not be the case, and has been updated to ignore `prob_stay` and set it to 1.

## Questions / things to look out for:
- Is there other reasoning to re-introducing the typing for the initialization dictionary?
- I have not touched the `initialization_functions.py` in the glm_hmm module. Should it be removed? Or do we keep it to be updated later?
- I assumed that in the HMM class initialization, the user would provide an integer seed for random initialization. Therefore, I have an integer `seed` as an input parameter to `generate_hmm_initial_params`, which I assume will be the seed stored in the class, which is then converted and split into random keys. We can always change this if it makes more sense to convert the seed to a random key inside the class
- **NOTE:** I added an exception in the parameter checks for `init_funcs` and `init_func`, since this conflict exists between the old glm_hmm initialization_functions.py and the new hmm one, but I don't think we want this exception long-term. Since the glm_hmm will be updated at some point, I don't think it's worth updating it now, also `init_func` in the glm_hmm module assume a SINGLE func, where `init_funcs` is basically the replacement of `init_registry` in the old version. Alternatively, we can rename the new `init_funcs` back to `init_registry` or something else


scripts/check_parameter_naming.py
```diff
@@ -125,6 +125,10 @@
     {"arg_name", "algo_name"},
     {"arg_name", "attr_name"},
     {"arg_name", "var_name"},
+    {"initial_proba_init", "initial_proba_init_kwargs"},
+    {"transition_proba_init", "transition_proba_init_kwargs"},
+    {"initial_proba_init_kwargs", "transition_proba_init_kwargs"},
+    {"init_func", "init_funcs"},
```
BalzaniEdoardo: I would drop the glm_hmm initialization in this PR, so this won't be an issue.

src/nemos/hmm/initialize_parameters.py
```diff
@@ -0,0 +1,595 @@
...
+    >>>
+    >>> # Generate random transition probabilities for 3 states
+    >>> n_states = 3
+    >>> transition_matrix = random_transition_proba_init(n_states)
+    >>> transition_matrix.shape
+    (3, 3)
+    >>> jnp.allclose(transition_matrix.sum(axis=1), 1.0)
+    Array(True, dtype=bool)
+    """
+    prob_transition = jax.random.normal(random_key, (n_states, n_states), dtype=float)
```
BalzaniEdoardo: This can be negative, should we do

```suggestion
    prob_transition = jax.random.uniform(random_key, (n_states, n_states), dtype=float)
```
instead?

src/nemos/hmm/initialize_parameters.py
```diff
@@ -0,0 +1,595 @@
...
+    >>>
+    >>> # Generate initial probabilities for 3 states
+    >>> n_states = 3
+    >>> init_probs = random_initial_proba_init(n_states)
+    >>> init_probs.shape
+    (3,)
+    >>> jnp.isclose(jnp.sum(init_probs), 1.0)
+    Array(True, dtype=bool)
+    """
+    prob = jax.random.normal(random_key, (n_states,), dtype=float)
```
BalzaniEdoardo: same here

src/nemos/hmm/initialize_parameters.py
```diff
@@ -0,0 +1,595 @@
...
+from typing import Any, Callable, Optional, Protocol, Tuple
+
+import jax
+import jax.numpy as jnp
+
+from ..type_casting import is_numpy_array_like
+from ..validation import _suggest_keys
+
+
+class InitFunctionHMM(Protocol):
```
BalzaniEdoardo, line 19: How do we handle cases where kwargs are required? For example:
```python
def my_init_func(n_states, random_key, X, y):
    # some compute using X and y
    return params
```

With `**kwargs`-only, the user must remember to manually construct the right dictionary (e.g. `{"X": X, "y": y}`) every time they pass a data-driven init function — even though the class already has `X` and `y` at fit time and could route them automatically.

My original idea of including `X` and `y` explicitly in the signature was to avoid this. The class can pass them as defaults, so the user never has to remember to wire them up manually. The set `{n_states, random_key, X, y}` also covers most reasonable initialization scenario — if someone needs something truly needs extra params, `**kwargs` is still available as an escape hatch.

src/nemos/hmm/initialize_parameters.py
```diff
@@ -0,0 +1,595 @@
...
+    missing = required_params - sig.parameters.keys()
+    if missing:
+        raise ValueError(
+            f"Custom initialization function must have the parameters {sorted(required_params)}. "
+            f"Missing: {sorted(missing)}."
+        )
+    kwargs = _validate_init_funcs_kwargs(func, kwargs)
+
+    # test output for 3 states
+    out = func(n_states=3, **kwargs)
```
BalzaniEdoardo: nice addition, this would need to be modified if X, y will be reintroduced in the signature (i.e. we would need X, y to be passed down here to perform the check of the output). If this check happen at the setup, when X and y are not available, this would not be possible

src/nemos/hmm/initialize_parameters.py
BalzaniEdoardo: another reason to add X and y is that for the GLMHHM we will need them (y is used to compute the initial intercept)

src/nemos/hmm/initialize_parameters.py
```diff
@@ -0,0 +1,595 @@
...
+        if not jnp.allclose(jnp.sum(out, axis=1), 1.0):
+            raise ValueError(
+                f"Custom initialization function for '{key}' must return a matrix with rows that sum to 1. "
+            )
+
+    return func, kwargs
+
+
+def generate_hmm_initial_params(
+    n_states: int, init_funcs: dict = {}, seed: int = 123
```
BalzaniEdoardo:
```suggestion
    n_states: int, init_funcs: dict = {}, seed: int = 123 | jax.random.PRNGkey
```

I would allow a key to be passed too

src/nemos/hmm/initialize_parameters.py
```diff
@@ -0,0 +1,595 @@
...
+        or DEFAULT_INIT_FUNCTIONS["transition_proba_init"]
+    )
+    transition_proba_init_kwargs = init_funcs["transition_proba_init_kwargs"] or {}
+
+    # split seed into two random keys, overwritten by user-provided random keys
+    keys = jax.random.split(jax.random.PRNGKey(seed), 2)
+    if "random_key" not in initial_proba_init_kwargs:
+        initial_proba_init_kwargs["random_key"] = keys[0]
+    if "random_key" not in transition_proba_init_kwargs:
+        transition_proba_init_kwargs["random_key"] = keys[1]
```
BalzaniEdoardo: what should we do here? should we raise if people pass seed via kwargs? is there a case in which one would want this level of control instead of specifying just a single seed at the class level?

src/nemos/hmm/initialize_parameters.py
```diff
@@ -0,0 +1,595 @@
...
+    TypeError
+        If alphas is not None or array-like.
+    """
+    if alphas is None:
+        return None
+    elif is_numpy_array_like(alphas)[1]:
+        alphas = jnp.asarray(alphas, dtype=float)
+        if alphas.shape != expected_shape:
+            raise ValueError(
+                f"Dirichlet prior alpha parameters for initial state probabilities "
```
BalzaniEdoardo:
```suggestion
                "Dirichlet prior alpha parameters for initial state probabilities "
```

src/nemos/hmm/initialize_parameters.py
```diff
@@ -0,0 +1,595 @@
...
+                f"must have shape ``{expected_shape}``, "
+                f"but got shape ``{alphas.shape}``."
+            )
+        if not jnp.all(alphas >= 1):
+            raise ValueError(
+                f"Dirichlet prior alpha parameters must be >= 1, but got values < 1"
```
BalzaniEdoardo:
```suggestion
                "Dirichlet prior alpha parameters must be >= 1, but got values < 1"
```

BalzaniEdoardo (commented):
This is looking clean, the main point to consider is the protocol, in particular if we want X and y in the input and the consequence of that in terms of checks and user experience

src/nemos/hmm/initialize_parameters.py
```diff
@@ -0,0 +1,595 @@
...
+            f"Invalid keyword argument for initialization function '{func.__name__}'. "
+            f"Unexpected keys: {sorted(bads)}. "
+            f"Expected parameters are: {sorted(sig.parameters.keys())}."
+        )
+    return kwargs
+
+
+def _validate_custom_init_func(
+    key: str, func: Callable, kwargs: Optional[dict] = None
+) -> Callable:
```
BalzaniEdoardo:
```suggestion
) -> Tuple[InitFunctionHMM., Dict]:
```

src/nemos/hmm/initialize_parameters.py
```diff
@@ -0,0 +1,595 @@
...
+
+class InitFunctionHMM(Protocol):
+    """Protocol for HMM probability initialization functions (initial and transition)."""
+
+    def __call__(
+        self,
+        n_states: int,
+        random_key: jax.random.PRNGKey,
+        **kwargs: Any,
+    ) -> Tuple[jnp.ndarray, jnp.ndarray]:
```
BalzaniEdoardo:
```suggestion
    ) -> jnp.ndarray:
```

src/nemos/hmm/initialize_parameters.py
```diff
@@ -0,0 +1,595 @@
...
+    Creates a transition probability matrix that favors staying in the current state.
+    The diagonal entries (probability of staying in current state) are set to `prob_stay`,
+    while off-diagonal entries (probability of transitioning to other states) are set to
+    (1 - prob_stay) / (n_states - 1).
+
+    Parameters
+    ----------
+    n_states :
+        Number of HMM states. Must be greater than 1.
+    seed :
```
BalzaniEdoardo:
```suggestion
    random_key. :
```

src/nemos/hmm/initialize_parameters.py
```diff
@@ -0,0 +1,595 @@
...
+    Initialize transition probabilities with uniform dynamics.
+
+    Creates a transition probability matrix that assign equal probability of
+    transitioning to any of the states.
+
+    Parameters
+    ----------
+    n_states :
+        Number of HMM states. Must be greater than 1.
+    seed :
```
BalzaniEdoardo:
```suggestion
    random_key. :
```

tests/test_hmm_initialization.py
```diff
@@ -0,0 +1,545 @@
...
+        with pytest.raises(ValueError, match="must be >= 1"):
+            _resolve_dirichlet_priors(alphas, expected_shape)
+
+    def test_invalid_type_raises_type_error(self):
+        """Test that invalid types raise TypeError."""
+        alphas = "invalid"
+        expected_shape = (3,)
+
+        with pytest.raises(TypeError, match="Invalid type"):
+            _resolve_dirichlet_priors(alphas, expected_shape)
```
BalzaniEdoardo, line 734:
```python
  def test_custom_init_no_default_key():                                                                                                                                                                            
      def my_init(n_states, random_key):
          return jnp.ones((n_states,)) / n_states                                                                                                                                                                   
      setup_hmm_initialization(initial_proba_init=my_init)  # should not raise
```


add this test that should not raise

src/nemos/hmm/initialize_parameters.py
BalzaniEdoardo: also, should pass the random key to func, see test that I added

BalzaniEdoardo (commented):
fixes mostly related to docstrings

src/nemos/hmm/initialize_parameters.py
```diff
@@ -0,0 +1,595 @@
...
+        raise ValueError(
+            f"Invalid keyword argument for initialization function '{func.__name__}'. "
+            f"Unexpected keys: {sorted(bads)}. "
+            f"Expected parameters are: {sorted(sig.parameters.keys())}."
+        )
+    return kwargs
+
+
+def _validate_custom_init_func(
+    key: str, func: Callable, kwargs: Optional[dict] = None
```
BalzaniEdoardo:
```suggestion
    key: str, func: InitFunctionHMM, kwargs: Optional[dict] = None
```

BalzaniEdoardo:
As a general comment, I am leaning on keeping `X` and `y` in the protocol signature because they can be passed automatically by the class, the user doesn't have to remember to pass those down if needed, and, at least `y` is almost certainly needed for the initialization of the GLM-HMM (setting the intercept).

The only drawback is that the setup-time check which calls the initialization function won't have access to the actual `X` and `y` (only available when `fit` is called). However, the class can provide generated `X` and `y` with the correct shape, since it has access to `n_states` and knows if it is a single or population model. One known limitation of this approach: a custom function could pass the dummy-data check but fail at fit time with real data — for example if it computes statistics from `y` that require a minimum number of samples. What I would do is reintroduce `X` and `y` in the protocol, and do not do a set-time check of the parameter initialization, only a runtime check (call the functions and validate the output params).

## Questions / things to look out for:

**Is there other reasoning to re-introducing the typing for the initialization dictionary?**

I would keep it. The dict is for internal usage if I understand correctly your rationale, and should not be created by the user. If that's the case, adding the typed dict would make development easier because mistyped keys will be flagged by static analysis. The user doesn't need to know this complex type exists. If we do not care about complexity and we are trying to only make the development easy, we can use a [`TypedDict`](https://typing.python.org/en/latest/spec/typeddict.html). This would make the code a little harder to read, but it will make the  debugging easier. 


**I have not touched the `initialization_functions.py` in the glm_hmm module. Should it be removed? Or do we keep it to be updated later?**

I would drop the old functions in this PR — if we need the original implementation we know where to find it. 

**I assumed that in the HMM class initialization, the user would provide an integer seed for random initialization. Therefore, I have an integer `seed` as an input parameter to `generate_hmm_initial_params`, which I assume will be the seed stored in the class, which is then converted and split into random keys. We can always change this if it makes more sense to convert the seed to a random key inside the class.**

That sounds good. See how the class is implemented — supporting both a key and an int would not be hard, and if we need that we can always add it later.

src/nemos/hmm/initialize_parameters.py
sjvenditto: will add back in!

src/nemos/hmm/initialize_parameters.py
sjvenditto: I'm ok with taking out this flexibility, and we can wait for someone to come and tell us they need more control over specific seeds

sjvenditto:
## New changes / updates:
- `X` and `y` have been reintroduced into the protocol
    - because of this, the output check of custom functions has been removed from the setup
- `random_key` is no longer allowed in kwargs. Now all "required" parameters of the protocol ("n_states", "X", "y", "random_key") cannot be set in kwargs.
- ` random_key` is type hinted to be `int | jax.Array` in `generate_hmm_init_params`. this is because `jax.random.PRNG` isn't a data type, it is a function, and was throwing an error when resolving `int | jax.random.PRNG`
- **NEW:** a KMeans initialization method for both parameters has been introduced
    - The initializer is a class that fits KMeans via sklearn, which uses the output labels to estimate the initial probability and transition matrix, which is computed in the `initial_probability` and `transition_probability` methods
    - The initialization functions that are stored in the dictionary are outside the class, `kmeans_initial_proba_init` and `kmeans_transition_proba_init`. These functions call the methods of the initializer class, where there is an optional argument `initializer` for sharing a single instance. If the initializer is not provided, it creates one within the function call. 
    - Ideally, future models that want to use KMeans to initialize parameters can inherit this class and add a method for initializing the model parameters

## Questions
- Is the KMeans too much for now? I can split it into a separate PR if so
    - if not: I'm fitting KMeans on the combined predictors and observations. Is this the right approach? 
- is it worth having `is_new_session` also an input parameter for the initialization protocols?
- should I add a validation step in `generate_hmm_initial_params` for custom functions since it was removed from `setup_hmm_initialization`? I can add a flag so it bypasses the check for internal functions?

## Remaining to-dos
- [x] reintroduce initialization dictionary class
- [x] remove `glm_hmm/initialize_parameters.py` and remove parameter naming exception
- [x] add support for PopulationGLM observations

src/nemos/hmm/initialize_parameters.py
sjvenditto: I removed this check here. Should I add it back inside of `generate_hmm_initial_params`?

src/nemos/hmm/initialize_parameters.py
```diff
@@ -0,0 +1,818 @@
+"""Initialization functions and related utility functions for HMMs."""
+
+import inspect
+from typing import Any, Callable, Literal, Optional, Protocol, Tuple
+
+import jax
+import jax.numpy as jnp
+from numpy.typing import NDArray
+from sklearn.cluster import KMeans
```
BalzaniEdoardo:
```suggestion
import lazy_loader as lazy
```

src/nemos/hmm/initialize_parameters.py
```diff
@@ -0,0 +1,818 @@
...
+
+from ..type_casting import is_numpy_array_like
+from ..typing import DESIGN_INPUT_TYPE
+from ..validation import _suggest_keys
+from .utils import initialize_new_session
+
```
BalzaniEdoardo, line 15:
```suggestion

sklearn = lazy.load("sklearn")
```

src/nemos/hmm/initialize_parameters.py
```diff
@@ -0,0 +1,818 @@
...
+        is_new_session: Optional[jnp.ndarray] = None,
+        random_key: int | jax.Array = 0,
+    ):
+        if isinstance(random_key, jax.Array):
+            random_key = int(random_key[-1])
+
+        self.n_states = n_states
+        self.random_key = random_key
+        self.is_new_session = initialize_new_session(y.shape[0], is_new_session)
+        self.model = KMeans(n_clusters=n_states, random_state=random_key)
```
BalzaniEdoardo:
```suggestion
        self.model = sklearn.cluster.KMeans(n_clusters=n_states, random_state=random_key)
```

src/nemos/hmm/initialize_parameters.py
```diff
@@ -0,0 +1,818 @@
...
+        Compute transition probabilities based on KMeans assigned states.
+
+        This computes the transition probabilities by counting the transitions between states across time points,
+        excluding transitions that occur at the start of new sessions.
+        """
+        transition_matrix = (
+            self.states[:-1][~self.is_new_session[1:]].T
+            @ self.states[1:][~self.is_new_session[1:]]
+        )
+        return transition_matrix / transition_matrix.sum(axis=1, keepdims=True)
```
BalzaniEdoardo, line 375: same as the init proba, transitions should not be 0 exactly. let's make sure that we never divide for 0 also, which can happen if there is never a transition from a particular state to another.

src/nemos/hmm/initialize_parameters.py
```diff
@@ -0,0 +1,818 @@
...
+        self.states = jax.nn.one_hot(self.model.labels_, num_classes=n_states)
+
+    def initial_probability(self):
+        """
+        Compute initial state probabilities based on KMeans assigned states.
+
+        This takes the average occurrence of each state at the start of sessions to estimate the initial state
+        probabilities.
+        """
+        initial_probability = self.states[self.is_new_session].sum(axis=0)
```
BalzaniEdoardo, line 353: since we are storing log-proba, let's make sure that no state as exactly 0 proba. For example, if there is only a single session, this init proba will be [0,...,0,1,0,...,0], then log proba will have -inf or nans

src/nemos/hmm/initialize_parameters.py
```diff
@@ -0,0 +1,818 @@
...
+    def __init__(
+        self,
+        n_states: int,
+        X: DESIGN_INPUT_TYPE,
+        y: NDArray | jnp.ndarray,
+        is_new_session: Optional[jnp.ndarray] = None,
+        random_key: int | jax.Array = 0,
+    ):
+        if isinstance(random_key, jax.Array):
+            random_key = int(random_key[-1])
```
BalzaniEdoardo: unsure if we can assume we can index into this. 
```suggestion
            random_key = int(jax.random.randint(random_key, (), 0, 2**31 - 1))                                                                  
```

BalzaniEdoardo: 2**31 - 1 is the max of uint32

src/nemos/hmm/initialize_parameters.py
```diff
@@ -0,0 +1,818 @@
...
+INITIALIZATION_FN_DICT = dict[
+    Literal[
+        "initial_proba_init",
+        "initial_proba_init_kwargs",
+        "initial_proba_init_custom",
+        "transition_proba_init",
+        "transition_proba_init_kwargs",
+        "transition_proba_init_custom",
+    ],
+    InitFunctionHMM | dict[str, Any] | bool | InitFunctionHMM | dict[str, Any] | bool,
```
BalzaniEdoardo:
```suggestion
    InitFunctionHMM | dict[str, Any] | bool,
```

src/nemos/hmm/initialize_parameters.py
```diff
@@ -0,0 +1,818 @@
...
+                f"Custom transition probabilities must have rows that sum to 1, but got row sums={proba.sum(axis=1)}."
+            )
+
+
+def generate_hmm_initial_params(
+    n_states: int,
+    X: DESIGN_INPUT_TYPE,
+    y: NDArray | jnp.ndarray,
+    random_key: int | jax.Array = 123,
+    init_funcs: dict = {},
```
BalzaniEdoardo:
```suggestion
    init_funcs: Optional[dict] = None,
```

src/nemos/hmm/initialize_parameters.py
```diff
@@ -0,0 +1,818 @@
...
+    transition_matrix :
+        Transition probability matrix of shape (n_states, n_states) where each row sums to 1.
+
+    See Also
+    --------
+    :func:`nemos.hmm.setup_hmm_initialization`
+        Function to set up the initialization functions and their kwargs based on user input.
+    """
+    # check for unexpected/unknown keys in init_funcs if user provided dictionary not made by setup_hmm_initialization
+    init_funcs = _validate_init_funcs_keys(init_funcs)
```
BalzaniEdoardo, line 783:
```suggestion
    init_funcs = {} if init_funcs is None else init_funcs
    init_funcs = _validate_init_funcs_keys(init_funcs)
```

src/nemos/hmm/initialize_parameters.py
```diff
@@ -0,0 +1,818 @@
...
+    Initialize transition probabilities with uniform dynamics.
+
+    Creates a transition probability matrix that assign equal probability of
+    transitioning to any of the states.
+
+    Parameters
+    ----------
+    n_states :
+        Number of HMM states. Must be greater than 1.
+    random_key :
```
BalzaniEdoardo, line 138: add X and y

tests/test_hmm_initialization.py
```diff
@@ -0,0 +1,726 @@
...
+
+        initial_prob1 = method(n_states, random_key=jax.random.PRNGKey(124))
+        initial_prob2 = method(n_states, random_key=jax.random.PRNGKey(999))
+
+        # Should be identical regardless of random key
+        assert jnp.allclose(initial_prob1, initial_prob2)
+
+    @pytest.mark.parametrize("method_subset", [[random_initial_proba_init]])
+    def test_non_deterministic(self, method, method_subset, use_method_for_test):
+        """Test that output is deterministic (same across different calls)."""
```
BalzaniEdoardo:
```suggestion
        """Test that output is non-deterministic (different across different calls)."""
```

BalzaniEdoardo (approved):
A bunch of minor things but then LGTM

src/nemos/hmm/initialize_parameters.py
BalzaniEdoardo: Sklearn takes a while to import, so let's lazy import it

PR #546, opened 2026-04-06
Point process GLM log-likelihood 

macari216:
This PR introduces the core negative log-likelihood computation for the Poisson point process GLM. It also includes:

* Utils necessary to compute NLL 
* Parameter container `PPGLMParams` and wrapper `PPGLMParamsWithKey`, which holds the JAX random key for stochastic optimization
* Tests for `log_likelihood` and `utils` modules 

Terminology used across docstrings:
* `n_spikes` - length of y; `n_events` - length of X
* `n_channels` - generic first dimension, used when the array can be `X`, `y`, or `mc_samples` (so either 2 or 3)
* `n_time_points` - generic second dimension
* `n_features` - first dimension of the coefficients before reshaping; after reshaping it splits into `n_predictors` * `n_basis_funcs`

Notes:
* `random_key` is stored as float in `PPGLMParamsWithKey` and cast to `uint32` in compute_loss. This is a workaround for the Adam optimizer I use with PP-GLM, which initializes all buffers to the minimum dtype (so `uint32` in this case) and crashes. The random key itself is masked out and never updated by Adam but it still has to be part of the parameters.
* `random_key` is part of the parameter wrapper (rather than the state) because it needs to be updated at each gradient step and _also_ passed into `compute_loss`.
* `_log_likelihood_scan` handles both LL terms (via the `lam_tilde_function` and `log` arguments) and both model variants: population (default) and single-neuron, where coefficients are reshaped to include a leading dimension of size 1 corresponding to the single target.

Questions: 
* `PPGLMParamsWithKey` wraps `PPGLMParams`, but since `PPGLMParams` is identical to `GLMParams`, should it just wrap `GLMParams` directly instead?
* Currently, log likelihood is computed as a sum over all spike times (and `aggregate_sample_scores` is not implemented). Should the default be mean LL/spike instead?


src/nemos/pp_glm/params.py
```diff
@@ -0,0 +1,35 @@
+from typing import Callable, Tuple, Union
+
+import equinox as eqx
+import jax.numpy as jnp
+from numpy.typing import ArrayLike
+
+from ..typing import DESIGN_INPUT_TYPE
```
BalzaniEdoardo, line 7:
```suggestion
from ..params import ModelParams
from ..typing import DESIGN_INPUT_TYPE
```

src/nemos/pp_glm/params.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+class PPGLMParams(eqx.Module):
```
BalzaniEdoardo, line 10:
```suggestion
class PPGLMParams(ModelParams):
```

src/nemos/pp_glm/params.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+PPGLMUserParams = Tuple[Union[DESIGN_INPUT_TYPE, ArrayLike], ArrayLike]
+
+
+class PPGLMParamsWithKey(eqx.Module):
+    """Wrapper around PPGLMParams that contains a PRNG key."""
+
+    params: PPGLMParams
+    random_key: jnp.ndarray
```
BalzaniEdoardo, line 29: This is a discussion we had with @sjvenditto, If we want a flatter or a more nested param structure. In the GLMHMM we landed on adding a `log_scale` attribute in a flat param class that have also `coef` and `intercept`.  Below is an analogous implementation for your case, I am not saying you need to go for it, just for illustration:

```suggestion
class PPGLMParamsWithKey(PPGLMParams):
    """Wrapper around PPGLMParams that contains a PRNG key."""

    random_key: jnp.ndarray
```

We are not sure about the current choice, and I think it is probably more convenient to nest the parameter, so that extracting the sub-parameters becomes easier by accessing the attribute. 

@sjvenditto what do you think? Should we all move back to nest the parameter classes?

src/nemos/pp_glm/params.py
```diff
@@ -0,0 +1,35 @@
...
+
+    @staticmethod
+    def regularizable_subtrees() -> (
+        list[Callable[["PPGLMParamsWithKey"], jnp.ndarray | dict]]
+    ):
+        return [lambda p: p.params.coef]
```
BalzaniEdoardo, line 35: If you go for a flat structure, you won't need this re-imlpementation

src/nemos/pp_glm/test_ppglm_log_likelihood.py
```diff
@@ -0,0 +1,469 @@
+import jax
```
BalzaniEdoardo, line 1: the test files should be in `tests/`

src/nemos/pp_glm/test_ppglm_log_likelihood.py
```diff
@@ -0,0 +1,469 @@
...
+
+from nemos.basis import RaisedCosineLogEval
+from nemos.pp_glm import log_likelihood, utils
+from nemos.pp_glm.validation import to_pp_glm_params, to_pp_glm_params_with_key
+
+
+def create_basis(n_basis_funcs=4, history_window=0.01):
+    """Use nemos RC Eval basis and return the evaluate method"""
+    basis = RaisedCosineLogEval(n_basis_funcs, bounds=(0, history_window), fill_value=0)
+    return lambda pts: basis.evaluate(pts)
```
BalzaniEdoardo, line 15:
```suggestion
    return basis.evaluate
```

src/nemos/pp_glm/test_ppglm_log_likelihood.py
```diff
@@ -0,0 +1,469 @@
...
+        # when divisible, padding length is 0
+        times = jnp.ones((3, 8))
+        out, pad_val, pad_len = utils.reshape_input_for_scan(times, scan_size=2)
+        assert out.shape == (4, 2, 3)  # (n_scans, scan_size, n_channels)
+        assert pad_len == 0
+
+        # when not divisible, padding fills to next multiple
+        times = jnp.ones((3, 9))
+        out, pad_val, pad_len = utils.reshape_input_for_scan(times, scan_size=2)
+        assert out.shape == (5, 2, 3)  # (n_scans, scan_size, n_channels)
```
BalzaniEdoardo, line 168: What's in `out[-1]`? should you check the fill value for it?

src/nemos/pp_glm/test_ppglm_log_likelihood.py
```diff
@@ -0,0 +1,469 @@
...
+        assert pad_len == 1
+
+        # test that padding is the last spike and that it's all the same
+        times = jnp.stack(
+            [jnp.arange(5, dtype=jnp.float64), jnp.arange(5, dtype=jnp.float64)]
```
BalzaniEdoardo, line 173:
```suggestion
            [jnp.arange(5, dtype=float), jnp.arange(5, dtype=float)]
```

the config used will determine the type

src/nemos/pp_glm/test_ppglm_log_likelihood.py
```diff
@@ -0,0 +1,469 @@
...
+        np.testing.assert_array_equal(pad_val, times[:, -1])
+
+    @pytest.mark.requires_x64
+    def test_build_mc_sampling_grid(self):
+        """test that the grid is built correctly with multiple epochs"""
+        recording_time = IntervalSet(start=[0.0, 6.0], end=[4.0, 10.0])
+        grid = utils.build_mc_sampling_grid(recording_time, M_samples=100)
+
+        # assert grid size is exactly M_samples
+        assert grid.shape[0] == 100
```
BalzaniEdoardo, line 188: is the full expected shape known? if so do assert `grid.shape == (100,...rest of the shape)`

src/nemos/pp_glm/test_ppglm_log_likelihood.py
```diff
@@ -0,0 +1,469 @@
...
+
+        assert np.all(in_epoch)
+
+        # test that edge case when the number of samples is less than the number of epochs
+        # raises an error
+        starts = np.array([0.0, 2.0, 4.0, 6.0, 8.0])
+        ends = np.array([1.0, 3.0, 5.0, 7.0, 9.0])
+        recording_time = IntervalSet(start=starts, end=ends)
+
+        with pytest.raises(ValueError):
```
BalzaniEdoardo, line 207: add a `match="matching str"`

BalzaniEdoardo:
```suggestion
        with pytest.raises(ValueError, match="beginning of the error message"):
```

src/nemos/pp_glm/test_ppglm_log_likelihood.py
```diff
@@ -0,0 +1,469 @@
...
+    def test_adjust_indices_and_spike_times(self):
+        """Test that shapes and indices are shifted correctly and that padding
+        is outside the history window"""
+
+        # returns already preprocessed X and y
+        dataset = create_dataset()
+
+        # test X shape increase by history window
+        n_spk_original = dataset["y"].shape[1]  # unchanged
+        assert dataset["X"].shape[1] == n_spk_original + dataset["max_window"]
```
BalzaniEdoardo, line 220: This seems a testing condition for 

`X, y = utils.adjust_indices_and_spike_times(X, history_window, max_window, y)`

shouldn't this be moved to the TestUtils and do just that?

src/nemos/pp_glm/test_ppglm_log_likelihood.py
```diff
@@ -0,0 +1,469 @@
...
+                jnp.array([-dataset["history_window"], 0]), ref, events
+            )
+        )
+
+        assert max_window == 3
+
+        # multiple spikes, multiple events
+        ref = jnp.array([1.0, 3.0])
+        events = jnp.array(
+            [0.990, 0.993, 0.995, 2.990, 2.993, 2.995, 2.999]  # 3 events within ref 1
```
BalzaniEdoardo, line 268: what happen if you add a spike at exactly 3.0? I think we should test this other boundary too

src/nemos/pp_glm/test_ppglm_log_likelihood.py
```diff
@@ -0,0 +1,469 @@
...
+        padding = dataset["X"][:, : dataset["max_window"]]
+        assert np.all(padding[0] < bound)
+
+        first_spike = dataset["y"][:, 0]
+        dts = first_spike[0] - padding[0]
+        basis_at_dts = dataset["eval_function"](dts)
+
+        np.testing.assert_array_equal(basis_at_dts, 0)
+
+    def test_compute_max_window_size(self):
```
BalzaniEdoardo, line 236: This one can be parametrized
```suggestion
    @pytest.mark.parametrize(
        "ref, events, expected_max_window",
        [ 
           (jnp.array([1.0]), jnp.array([0.1, 0.5, 0.990, 0.993, 0.995]), 3),
           (jnp.array([1.0, 3.0]), jnp.array([0.990, 0.993, 0.995, 2.990, 2.993, 2.995, 2.999]), 4)
        ]
    )
    def test_compute_max_window_size(self, ref, events, expected_max_window):
         ...

             
```

This special case should be its own test:
```python
        max_window = int(
            utils.compute_max_window_size(
                jnp.array([-dataset["history_window"], 0]),
                dataset["X"][0],
                dataset["X"][0],
            )
        )
```

src/nemos/pp_glm/utils.py
```diff
@@ -0,0 +1,215 @@
...
+from pynapple import IntervalSet
+
+from ..typing import DESIGN_INPUT_TYPE
+
+
+### SCAN UTILS
+@partial(jax.jit, static_argnums=2)
+def slice_array(array: jnp.ndarray, i: int, window_size: int):
+    """
+    Select events within the history window.
```
BalzaniEdoardo, line 16: Jit-compilable equivalent of `array[:, i-window_size:i]`

BalzaniEdoardo: Add a second line with a description like that one, to make it easier to understand what is going on

src/nemos/pp_glm/utils.py
```diff
@@ -0,0 +1,215 @@
...
+import jax
+import jax.numpy as jnp
+from numpy.typing import ArrayLike
+from pynapple import IntervalSet
+
+from ..typing import DESIGN_INPUT_TYPE
+
+
+### SCAN UTILS
+@partial(jax.jit, static_argnums=2)
```
BalzaniEdoardo, line 13: unless you are using this outside a jit-compilable function, what I would do is jit compile only the callable that are used directly, and let the jax tracing compile everything downstream.

In particular, this is called by `_log_likelihood_scan` which is called by `_negative_log_likelihood`. If both this callable are called upstream on their own, then jit compile them, if only is called `_negative_log_likelihood` by your algorithm jit compile only that one.

You can mark `window_size` as static as part of that jit-compilation.

src/nemos/pp_glm/test_ppglm_log_likelihood.py
BalzaniEdoardo: note that in general `evaluate` ignores the `fill_value`, which is set by `_compute_features`. this works because the RaisedCosine clipping is actually doing the job already, see lines 150-159 in `basis/_raised_cosine_basis.py`. One one hand, this optimal for you, since there is no need to post-process the output of the callable and replace values with a `where`, on the other hand it might be an unexpected behavior of `evauate`.

I think the method should be probably made private, so that we can keep the current behavior, which is efficient, but for this pr we should make sure that the out-of bounds value are actually set to zero, in case we change the backend implementation of the basis and mess up your code.

src/nemos/pp_glm/utils.py
```diff
@@ -0,0 +1,215 @@
...
+        array,
+        (0, i - window_size),
+        (
+            n_channels,
+            window_size,
+        ),
+    )
+
+
+def reshape_coef_for_scan(weights: jnp.ndarray, n_basis_funcs: int):
```
BalzaniEdoardo, line 43: Why do we need this reshaping as part of the `_log_likelihood` call? can't we assume that the class reshapes the weights internally? We can also request the user to provide the weight of the correct shape already. 

Also, if more convenient, for a user we can request to provide `(n_predictors, n_basis_funcs, n_neurons)` or `(n_predictors, n_basis_funcs)` for pop. or single neu glms. Then the validator `to_model_params` reshape it to `(n_predictors * n_basis_funcs, n_neurons)` for you if that's more convenient for the computation

src/nemos/pp_glm/utils.py
```diff
@@ -0,0 +1,215 @@
...
+        n_target_neurons = weights.shape[1]
+        return weights.reshape(-1, n_basis_funcs, n_target_neurons)
+    else:
+        raise ValueError(
+            f"Weights must be either 1d or 2d array, the provided weights have shape {weights.shape}"
+        )
+
+
+@partial(jax.jit, static_argnums=1)
+def reshape_input_for_scan(times: jnp.ndarray, scan_size: int):
```
BalzaniEdoardo, line 73: Similar, the jitting may happen only in the highest level of the callable stack, so that we let the tracer figure out what to do here.

src/nemos/pp_glm/utils.py
```diff
@@ -0,0 +1,215 @@
...
+    padded_times_reshaped = padded_spikes.reshape(n_channels, scan_size, -1).transpose(
+        2, 1, 0
+    )
+
+    return padded_times_reshaped, padding_value, padding_len
+
+
+def build_mc_sampling_grid(recording_time: IntervalSet, M_samples: int):
+    """
+    Build a stratified sampling grid of bin midpoints for Monte Carlo integration
```
BalzaniEdoardo, line 108: likely needs a one liner description followed by a longer function description. something like,

```python
     Build a stratified sampling grid for Monte Carlo integration.
     
     Detailed description of the function....
```

Otherwise the linter probably will complain

src/nemos/pp_glm/utils.py
BalzaniEdoardo, line 73: add type annotation of the output
```python
def reshape_input_for_scan(times: jnp.ndarray, scan_size: int) -> Tuple[your-type-annotation]:
```

src/nemos/pp_glm/utils.py
```diff
@@ -0,0 +1,215 @@
...
+    padding = jnp.full((n_channels,) + (padding_len,), padding_value[:, None])
+    padded_spikes = jnp.hstack((times, padding))
+    padded_times_reshaped = padded_spikes.reshape(n_channels, scan_size, -1).transpose(
+        2, 1, 0
+    )
+
+    return padded_times_reshaped, padding_value, padding_len
+
+
+def build_mc_sampling_grid(recording_time: IntervalSet, M_samples: int):
```
BalzaniEdoardo, line 106: It spits out an array? can you type annotate the output?

src/nemos/pp_glm/utils.py
```diff
@@ -0,0 +1,215 @@
...
+
+    Subdivides each recording epoch into equal-width bins proportionally to its
+    length and ensures the total grid size equals M_samples exactly.
+
+    Parameters
+    ----------
+    recording_time :
+        pynapple IntervalSet defining the recording epochs.
+    M_samples :
+        Total number of Monte Carlo sample points.
```
BalzaniEdoardo, line 119: shape?

src/nemos/pp_glm/utils.py
```diff
@@ -0,0 +1,215 @@
...
+            f"The number of MC samples ({M_samples}) must be equal or greater than the number of recording "
+            f"epochs {len(recording_time.start)})."
+        )
+    dt = recording_time.tot_length() / M_samples
+    starts, ends = recording_time.start, recording_time.end
+    M_sub = jnp.floor((ends - starts) / dt).astype(int)
+    M_sub = M_sub.at[-1].set(M_samples - jnp.sum(M_sub[:-1]))
+    return jnp.concatenate(
+        [jnp.linspace(s + dt, e, m) - dt / 2 for s, e, m in zip(starts, ends, M_sub)]
+    )
```
BalzaniEdoardo, line 137: There is an argument for this to be numpy purely because imagine one is using a gpu and has 1K + sessions, it is 1K calls to the gpu, which is a crazy overhead. I think this function will be called once before the fit, so it is fine if you use pure numpy, and covert the output to jax after

src/nemos/pp_glm/utils.py
```diff
@@ -0,0 +1,215 @@
...
+
+
+### DATA PREPROCESSING UTILS
+@jax.jit
+def compute_max_window_size(
```
BalzaniEdoardo, line 142: If called only once at setup time, @jax.jit is counterproductive: the compilation 
overhead likely exceeds the runtime of the function itself.

src/nemos/pp_glm/utils.py
```diff
@@ -0,0 +1,215 @@
...
+    :
+        Maximum number of events within a history window.
+    """
+    idxs_plus = jnp.searchsorted(event_times, ref_spike_times + bounds[1])
+    idxs_minus = jnp.searchsorted(event_times, ref_spike_times + bounds[0])
+    within_windows = idxs_plus - idxs_minus
+    return jnp.max(within_windows)
+
+
+@partial(jax.jit, static_argnums=(1, 2))
```
BalzaniEdoardo, line 172: same thing with the jitting as the other functions. It is probably worth not compiling

src/nemos/pp_glm/utils.py
```diff
@@ -0,0 +1,215 @@
+from functools import partial
```
BalzaniEdoardo, line 1: for this file, it would be nice to have the type annotation for all function output, and whenever possible, the input and output shape in the docstrings. For example,

```python
def myfunc(x: jnp.ndarray) -> jnp.ndarray:
    """
    ...
    Parameters
    ------------
    x:
        Input array, shape ``(n_neurons, n_features)``.
    Returns
    --------
    :
        Output, shape ``(n_neurons,)` `.
    """
    ...
```

src/nemos/pp_glm/utils.py
```diff
@@ -0,0 +1,215 @@
...
+### SCAN UTILS
+@partial(jax.jit, static_argnums=2)
+def slice_array(array: jnp.ndarray, i: int, window_size: int):
+    """
+    Select events within the history window.
+
+    Parameters
+    ----------
+    array :
+        event time series. Shape (n_channels, n_events).
```
BalzaniEdoardo, line 21:
```suggestion
        event time series. Shape ``(n_channels, n_events)``.
```

BalzaniEdoardo: Similar wrapping on the other shapes. This is more important for function that will be in the documentation website (public API) because it will be rendered as an inline code segment. For internal function I would still do it for consistency, but not a big deal

src/nemos/pp_glm/validation.py
```diff
@@ -0,0 +1,21 @@
+"""Validation classes for PPGLM and PopulationPPGLM models."""
+
+import jax.numpy as jnp
```
BalzaniEdoardo, line 3: The parameter reshaping can happen at this level if the parameter shape is not expected to change during the computation which i think it's true

src/nemos/pp_glm/log_likelihood.py
```diff
@@ -0,0 +1,347 @@
...
+jax.config.update("jax_enable_x64", True)
+
+
+def _compute_lam_tilde_single(
+    dts: jnp.ndarray,
+    i: int,
+    event_ids: jnp.ndarray,
+    weights: Union[jnp.ndarray, Dict],
+    bias: jnp.ndarray,
+    eval_function,
```
BalzaniEdoardo, line 19:
```suggestion
    eval_function: Callable,
```

src/nemos/pp_glm/log_likelihood.py
```diff
@@ -0,0 +1,347 @@
...
+):
+    """
+    Evaluate the non-rectified firing rate (lambda tilde) for a single
```
BalzaniEdoardo, line 22:
```suggestion
    Evaluate the non-rectified firing rate.

    Evaluate the non-rectified firing rate (lambda tilde) for a single target neurons at 
    a single time point. This function selects the coefficients for the predictors present 
    in the history window, evaluates the basis functions at the lag times, and accumulates 
    their weighted sum plus the bias for the target neuron indexed by i.

```

I think the recommended style is a one liner description in imperative followed by a longer description.

src/nemos/pp_glm/log_likelihood.py
```diff
@@ -0,0 +1,347 @@
...
+from ..typing import DESIGN_INPUT_TYPE
+from . import utils
+from .params import PPGLMParams, PPGLMParamsWithKey
+
+jax.config.update("jax_enable_x64", True)
+
+
+def _compute_lam_tilde_single(
+    dts: jnp.ndarray,
+    i: int,
```
BalzaniEdoardo, line 15: instead of i, can we call this `neuron_id`?

src/nemos/pp_glm/log_likelihood.py
```diff
@@ -0,0 +1,347 @@
...
+    event_ids: jnp.ndarray,
+    weights: Union[jnp.ndarray, Dict],
```
BalzaniEdoardo, line 17: weights can now be any pytree in nemos.

src/nemos/pp_glm/log_likelihood.py
```diff
@@ -0,0 +1,347 @@
...
+
+    Returns
+    -------
+    :
+        Scalar non-rectified firing rate for the target neuron.
+    """
+    w = weights[event_ids, :, i]  # shape (max_window, n_basis_funcs)
+    fx = eval_function(dts)  # shape (max_window, n_basis_funcs)
+
+    return jnp.sum(fx * w) + bias[i]
```
BalzaniEdoardo, line 51: is that just `fx@w + bias[i]`

src/nemos/pp_glm/log_likelihood.py
```diff
@@ -0,0 +1,347 @@
...
+
+
+def _compute_lam_tilde_all(
+    dts: jnp.ndarray,
+    i: int,
```
BalzaniEdoardo, line 56: same with the name

src/nemos/pp_glm/log_likelihood.py
```diff
@@ -0,0 +1,347 @@
...
+        An array of predictor ids corresponding to all events in the history window.
+    weights :
+        Reshaped model coefficients. Shape (n_predictors, n_basis_funcs, n_neurons).
+    bias :
+        Intercept for each target neuron. Shape (n_neurons,).
+
+    Returns
+    -------
+    :
+        Scalar non-rectified firing rate for the target neuron.
```
BalzaniEdoardo, line 46:
```suggestion
        Scalar non-rectified firing rate for the target neuron. Shape ``(1,)``.
```

src/nemos/pp_glm/log_likelihood.py
```diff
@@ -0,0 +1,347 @@
...
+    return jnp.sum(fx * w) + bias[i]
+
+
+def _compute_lam_tilde_all(
+    dts: jnp.ndarray,
+    i: int,
+    event_ids: jnp.ndarray,
+    weights: Union[jnp.ndarray, Dict],
+    bias: jnp.ndarray,
+    eval_function,
```
BalzaniEdoardo, line 60:
```suggestion
    eval_function: Callable,
```

src/nemos/pp_glm/log_likelihood.py
```diff
@@ -0,0 +1,347 @@
...
+    fx = eval_function(dts)  # shape (max_window, n_basis_funcs)
+
+    return jnp.sum(fx * w) + bias[i]
+
+
+def _compute_lam_tilde_all(
+    dts: jnp.ndarray,
+    i: int,
+    event_ids: jnp.ndarray,
+    weights: Union[jnp.ndarray, Dict],
```
BalzaniEdoardo, line 58: same as above

src/nemos/pp_glm/log_likelihood.py
```diff
@@ -0,0 +1,347 @@
...
+    bias: jnp.ndarray,
+    eval_function,
+):
+    """
+    Evaluate the non-rectified firing rate (lambda tilde) for all target neurons
```
BalzaniEdoardo, line 63: split with one liner + description

src/nemos/pp_glm/log_likelihood.py
```diff
@@ -0,0 +1,347 @@
...
+
+
+def _compute_lam_tilde_all(
+    dts: jnp.ndarray,
+    i: int,
+    event_ids: jnp.ndarray,
+    weights: Union[jnp.ndarray, Dict],
+    bias: jnp.ndarray,
+    eval_function,
+):
```
BalzaniEdoardo, line 61: return annotation

src/nemos/pp_glm/utils.py
BalzaniEdoardo: Also, with `n_channels` what do you mean? is it equal to 2, `(spike time, spike id)` ? we should describe what this channel dimension is

src/nemos/pp_glm/log_likelihood.py
```diff
@@ -0,0 +1,347 @@
...
+    Iterates over time_points in y, selects the recent history events form X, evaluates
+    the linear combination of predictors via lam_tilde_function, and accumulates the (log-)firing
+    rates. Padding added by reshape_input_for_scan is subtracted out at the end.
+
+    Parameters
+    ----------
+    X :
+        Padded event time series. Shape (2, n_events).
+    eval_pts :
+        Observed spike time series or MC sample points. Shape (n_channels, n_time_points).
```
BalzaniEdoardo, line 159: IF this is always (2, n_time_points) I would write that. I would also add, `eval_pts[0]` are the spike times, `eval_pts[1]` are the spike ids marking which neuron spiked.

BalzaniEdoardo: Also, this packing indices and time into an array is it more efficient than keeping the array separated an operate two dyamic slices?

src/nemos/pp_glm/log_likelihood.py
```diff
@@ -0,0 +1,347 @@
...
+        random_key, shape=(M_samples,), minval=0.0, maxval=dt
+    )
+    tau_m = M_grid + epsilon_m
+    tau_m_idx = jnp.searchsorted(X[0], tau_m)
+    mc_spikes = jnp.vstack((tau_m, tau_m_idx))
+
+    return mc_spikes
+
+
+def _log_likelihood_scan(
```
BalzaniEdoardo, line 134: Two potential simplifications worth considering:

1. **Unify `_compute_lam_tilde_single` and `_compute_lam_tilde_all`**: both reduce to the same einsum,
```python
   fx = eval_function(dts)          # (max_window, n_basis_funcs)
   w = weights[event_ids]           # (max_window, n_basis_funcs, n_neurons)
   return jnp.einsum("ij,ij...->...", fx, w) + bias
```

   The two cases can then be handled by two thin scan bodies that differ only in whether they pre-slice `weights` and `bias` to the target neuron index before calling a shared compute function. This removes the dummy unused `i` argument in `_compute_lam_tilde_all` and makes the einsum the single computation.

2. **Store times and ids as separate arrays from the start**: since ids are only ever used for indexing (never arithmetic), they can be kept as `int32` throughout. Passing a tuple to the scan instead of a single float64 array would eliminate all `.astype(int)` conversions inside the scan body:
```python
   jax.lax.scan(scan_fn, init, (pts_times, pts_ids))  # ids already int32
```

This requires  to split `X`  into two arrays. The input pre-processing of the class can take care of that. We may still ask the user to provide an array of two channels, but we can pre-processing and split and convert the channels. the function for drawing samples will return the arrays as well. I attached a little benchmarking script which compares a single slice and then conversion to two slices, no conversion. On my cpu the two slice operation seems faster, also, on the gpu (even though the gpu scan is super slow). 


[benchmark_two_slices.py](https://github.com/user-attachments/files/26521019/benchmark_two_slices.py)

src/nemos/pp_glm/log_likelihood.py
```diff
@@ -0,0 +1,347 @@
...
+    padding_contrib = scan_fn(jnp.array(0.0), padding_val)[0] * padding_len
+
+    return jnp.sum(out) - padding_contrib
+
+
+def _negative_log_likelihood(
+    X: DESIGN_INPUT_TYPE,
+    y: jnp.ndarray,
+    params: PPGLMParams,
+    random_key: jnp.ndarray,
```
BalzaniEdoardo, line 214: We currently have moved the `params` as the first argument of the compute function, see for example the `em_hmm` function in hmm/expectation_maximization.py

src/nemos/pp_glm/log_likelihood.py
```diff
@@ -0,0 +1,347 @@
...
+    inverse_link_function,
+    M_samples,
+    M_grid,
+    recording_time,
+    n_basis_funcs,
+    scan_size,
+    max_window,
+    eval_function,
```
BalzaniEdoardo, line 222: here the type annotation are missing

BalzaniEdoardo (changes requested):
I think the algo are looking good, we need to clean stuff up a little, but this is looking good

BalzaniEdoardo:
@macari216 is this stale? what is the difference with the new one?

BalzaniEdoardo:
@macari216 should we close this one since we have #616 ?

PR #552, opened 2026-04-16
Initialization of `BaseHMM` class

sjvenditto:
## Summary
This PR adds the base components of the `BaseHMM` abstract class and tests initialization and validation using a mock class. This class inherits `BaseRegressor` using HMM-generic parameter types `HMMModelParamsT` and `HMMUserProvidedParamsT` (introduced in #540). In this way, classes that inherit `BaseHMM` do not also need to inherit `BaseRegressor`, and will specify typing through `BaseHMM`, e.g. `MockHMM(BaseHMM[MockHMMParams, MockHMMUserParams])`

### New class attributes:
- **`hmm_initialization_funcs`**: This is included in the init signature to have a sklearn compatible way of copying or cross validating over initialization methods. However, the user should not be defining the functions here, and instead should be going through the new `setup` method (detailed below).
- **`_validator_class`**: similar to `GLM` and `PopulationGLM`, this attribute is defined outside of `__init__` in the inheriting class so that the validator can be initialized correctly when setting `n_states`

### New/modified class methods:
- **`setup`**: An optional method to run after model instantiation if the user wants to update what initialization method is used for initial probabilities and transition probabilities (with initialization methods introduced in #542). This is essentially a wrapper around `setup_hmm_initialization`, which updates the internal `_hmm_initialization_funcs` dictionary. 
- **`_check_hmm_is_fit`**: called by `_check_is_fit`, checks HMM-specific parameters (initial prob and transition prob)
- **`_check_model_is_fit`**: a new _abstract_ method, also called by `_check_is_fit`. to be implemented by inheriting class to check model-specific parameters.
- **`_check_is_fit`**: in an attempt to generalize core functions, this function is being handled by the base class. it calls the two functions listed above.
- **`_hmm_params_initialization`**: private function for calling `generate_hmm_initial_params` and indicating whether or not the output needs to be validated (i.e. it was generated from a custom function). called by `_model_specific_initialization` (see below)
- **`_model_params_initialization`**: a new _abstract_ method to mirror `_hmm_params_initialization` except for model parameters. needs to be implemented by inheriting class. called by `_model_specific_initialization` (see below)
- **`_model_specific_initialization`**: similar to `_check_is_fit`, this function is being handled by the base class. it calls the two previously mentioned functions to separately initialize hmm and model params. it will validate all parameters if at least one is being generated by a custom function. _This called for new assumptions/restrictions on the validator class for this to work simply (see below)_
- **`_validate_and_prepare_inputs`**: this function was pulled and modified from GLM-HMM and adds the option to pass a user-defined `is_new_session`. this parameter is validated and recast to a jax boolean array (if it isn't already), and allow some flexibility on what `is_new_session` can be passed as. _This called for updates to the validator class and utility functions (detailed below)_

### Class attributes pulled from GLM-HMM:
- `n_states`, `dirichlet_prior_alphas_init_prob`, `dirichlet_prior_alphas_transition`, `regularizer`, `regularizer_strength`, `solver_name`, `solver_kwargs`, `maxiter`, `tol`, `seed`
- Validation of these parameters, initialization of `BaseRegressor`, and initialization of the validator were all moved to `BaseHMM`.

### Changes to `HMMValidator`
1. In order to easily concatenate generated HMM and model parameters in `_validate_and_prepare_inputs`, I went back to the assumption that model parameters come first, and HMM parameters are always the last two. This removed the `initial_prob_ind` and `transition_prob_ind` properties of the validator.
2. I removed the field `expected_param_dims` from and moved `check_array_dimensions` back to the `GLMHMMValidator`. This is because checking the array dimensions is redundant with shape checking, and I believe we should remove this step altogether. However, to limit updates to `GLMHMMValidator` for now, I moved this function instead of removing this step. 
3. `validate_and_cast_is_new_session` is a new validation check specifically for `is_new_session`. It calls the utils function **`initialize_is_new_session`**, which allows `is_new_session` to be `None`, a boolean array, an integer array, or an interval set. Error checking in handled inside this function. Before returning the validated `is_new_session`, it calls the utils function **`shift_nan_is_new_session`** to make sure all new-session markers are moved off of NaN values in X and/or y.

## Outstanding questions / concerns
- is it OK that we reintroduce the requirement that HMM params must always come last in the inheriting model's user params? it's the only way to generically and simply concatenate all user params in `_model_specific_initialization` to pass through the parameter validation sequence. we can't use the indices of the HMM params, because for jax compatibility, we can't insert them in-place, and recreating the array also assumes knowledge of the index (and number) of the model params, which is model dependent. if we want to avoid this assumption, we can remove `_model_specific_initialization` from `BaseHMM` and have the inheriting class handle model-specific initialization. However, I don't find the assumption to be too limiting in general.
- Similar to how it was implemented in `GLMHMM`, `_model_specific_initialization` will validate the generated parameters if they were created by custom functions. However, this is redundant with checks I had in place in `generate_hmm_initial_params`. Should I remove the shape and output sum checks from `generate_hmm_initial_params`?
- I think we should rename and/or combine the dirichlet prior parameters. Should we do that in this PR or make a small one after?
- The way NaN's in the middle of sessions are handled now is different than before. Before, you were adding new session indicators after any NaN break, where now session indicators are only _moved_ if they fall during a NaN, not added. This is negligible if we don't allow NaN's to fall in the middle of the session. But if we want to allow that behavior in the future, we will want to think about how to handle it. Some thoughts:
    - I don't know if we can make any judgements on whether to treat a trial with a NaN value as if it never happened, or as a new session. If this occurs for some reason other than convolution, i.e. an animal has a violation trial, it's up to the experimenter on how they want to deal with this, not us. They should be the ones responsible for dropping those trials or updating a new session boundary as they see fit.
    - If the NaNs are occurring at session edges from convolution: if this is being done using pynapple time supports to segment the data, then theoretically the current approach should be sufficient to handle it IF the user doesn't pass anything for `is_new_session`. ideally, the time support used for the convolution will be the same that is used to identify session boundaries. they will each get shifted off any NaN values at the start of the sessions, and any NaN values at the ends of sessions can be dropped without worry. But if they do pass another value for `is_new_session`? I don't know
- the checks inside `initialize_is_new_session` makes it no longer jittable. Therefore, I had to remove calls to this function in the EM functions. This makes it such that `is_new_session` will still default to a single session if `None`, but no checks are performed (i.e. the first index is no longer ensured to be 1) if the user passes something. I think this is OK because this will be enforced within the model when validating `is_new_session`, and anyone using the EM algorithms outside of the model already does so at their own risk

### DIFF
--- a/src/nemos/hmm/hmm.py
+++ b/src/nemos/hmm/hmm.py
@@ -3,14 +3,24 @@
 from __future__ import annotations
 
 import abc
+import warnings
 from numbers import Number
-from typing import Any, Callable, Optional, Tuple, Union
+from typing import Any, Callable, Literal, Optional, Tuple, Union
 
 import jax
 import jax.numpy as jnp
+import pynapple as nap
+from numpy.typing import ArrayLike, NDArray
 
+from .. import tree_utils
 from ..base_regressor import BaseRegressor
+from ..hmm.expectation_maximization import (
+    forward_backward,
+    forward_pass,
+    max_sum,
+)
 from ..regularizer import Regularizer
+from ..type_casting import support_pynapple
 from ..typing import (
     DESIGN_INPUT_TYPE,
 )
@@ -22,6 +32,7 @@ from .initialize_parameters import (
     setup_hmm_initialization,
 )
 from .params import HMMModelParamsT, HMMUserParams, HMMUserProvidedParamsT
+from .utils import _check_state_format
 from .validation import HMMValidator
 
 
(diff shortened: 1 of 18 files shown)


BalzaniEdoardo:
> ## Outstanding questions / concerns
> 
>     * is it OK that we reintroduce the requirement that HMM params must always come last in the inheriting model's user params? it's the only way to generically and simply concatenate all user params in `_model_specific_initialization` to pass through the parameter validation sequence. we can't use the indices of the HMM params, because for jax compatibility, we can't insert them in-place, and recreating the array also assumes knowledge of the index (and number) of the model params, which is model dependent. if we want to avoid this assumption, we can remove `_model_specific_initialization` from `BaseHMM` and have the inheriting class handle model-specific initialization. However, I don't find the assumption to be too limiting in general.

As long as it's obvious that something will fail if we implement subclasses that do not respect this parameter convention, we are good. I think any validation would fail if we implement parameter classes violating this assumption. we can make this explicit by adding a general `BaseHMM` subclass test that check that the parameters are ordered correctly, if we know the number of parameters we can create 

```python
user_par = [False] * (n_params - 2) + [True, True]
validator.to_model_params(user_par)
# check that the hmm_params are true and true and the rest is false
```

> 
>     * Similar to how it was implemented in `GLMHMM`, `_model_specific_initialization` will validate the generated parameters if they were created by custom functions. However, this is redundant with checks I had in place in `generate_hmm_initial_params`. Should I remove the shape and output sum checks from `generate_hmm_initial_params`?

yes, I would remove redundant checks.

 
>     * I think we should rename and/or combine the dirichlet prior parameters. Should we do that in this PR or make a small one after?

I agree, I was thinking about that just by reading the param name in the description. What about `dirichlet_init`, `dirichlet_transition`? or `dirichlet_alpha_init/transition`?

>     * The way NaN's in the middle of sessions are handled now is different than before. Before, you were adding new session indicators after any NaN break, where now session indicators are only _moved_ if they fall during a NaN, not added. This is negligible if we don't allow NaN's to fall in the middle of the session. But if we want to allow that behavior in the future, we will want to think about how to handle it. Some thoughts:
>       
>       * I don't know if we can make any judgements on whether to treat a trial with a NaN value as if it never happened, or as a new session. If this occurs for some reason other than convolution, i.e. an animal has a violation trial, it's up to the experimenter on how they want to deal with this, not us. They should be the ones responsible for dropping those trials or updating a new session boundary as they see fit.
>       * If the NaNs are occurring at session edges from convolution: if this is being done using pynapple time supports to segment the data, then theoretically the current approach should be sufficient to handle it IF the user doesn't pass anything for `is_new_session`. ideally, the time support used for the convolution will be the same that is used to identify session boundaries. they will each get shifted off any NaN values at the start of the sessions, and any NaN values at the ends of sessions can be dropped without worry. But if they do pass another value for `is_new_session`? I don't know

If they pass other values that do not match the time support, I think we should treat as boundaries the union of the epochs in the time support and the new indicators session indicators (is_new_session) provided by the user. then we can apply the same logic; will this work?

>     * the checks inside `initialize_is_new_session` makes it no longer jittable. Therefore, I had to remove calls to this function in the EM functions. This makes it such that `is_new_session` will still default to a single session if `None`, but no checks are performed (i.e. the first index is no longer ensured to be 1) if the user passes something. I think this is OK because this will be enforced within the model when validating `is_new_session`, and anyone using the EM algorithms outside of the model already does so at their own risk

I think it is ok too. The algorithms should be only invoked internally, and we can make sure that we pass `is_new_session` correctly.

src/nemos/hmm/hmm.py
```diff
@@ -0,0 +1,386 @@
...
+    Parameters
+    ----------
+    n_states :
+        The number of hidden states in the HMM. Must be a positive integer.
+    dirichlet_prior_alphas_init_prob :
+        Alpha parameters for the Dirichlet prior over the initial state probabilities.
+        Shape ``(n_states,)``. If None, a flat (uninformative) prior is assumed.
+    dirichlet_prior_alphas_transition :
+        Alpha parameters for the Dirichlet prior over the transition probabilities.
+        Shape ``(n_states, n_states)``. If None, a flat (uninformative) prior is assumed.
```
BalzaniEdoardo, line 44: Another idea would could grouping those into a single `dirichlet_prior` which is a tuple of the two arrays?

src/nemos/hmm/hmm.py
```diff
@@ -0,0 +1,386 @@
...
+        self.dirichlet_prior_alphas_init_prob = dirichlet_prior_alphas_init_prob
+        self.dirichlet_prior_alphas_transition = dirichlet_prior_alphas_transition
+
+        self.seed = seed
+        self.maxiter = maxiter
+        self.tol = tol
+
+        # fit attributes
+        self.transition_prob_: jnp.ndarray | None = None
+        self.initial_prob_: jnp.ndarray | None = None
```
BalzaniEdoardo, line 114:
```suggestion
        self.transition_prob_: Optional[jnp.ndarray] = None
        self.initial_prob_: Optional[jnp.ndarray] = None
```

src/nemos/hmm/hmm.py
```diff
@@ -0,0 +1,386 @@
...
+        initial_proba_init_kwargs: Optional[dict] = None,
+        transition_proba_init: Optional[str | Callable] = None,
+        transition_proba_init_kwargs: Optional[dict] = None,
+    ):
+        """
+        Set up the HMM model with specified initialization functions for the initial and transition probabilities.
+
+        An optional initialization step that allows for users to specify initialization functions other than the
+        defaulst for initial and transition probabilities. The user can specify other built-in initialization functions
+        or provide custom ones. If no initialization functions are provided, default initialization will be used.
```
BalzaniEdoardo, line 130: we should mention that custom initialization function must match a specific signature. It would be nice to have an example section showing how to:

- set a custom init function
- set one of the available non-default initialization using a string.

src/nemos/hmm/hmm.py
```diff
@@ -0,0 +1,386 @@
...
+
+    @property
+    def hmm_initialization_funcs(self):
+        """Dictionary of initialization functions for HMM parameters."""
+        return self._hmm_initialization_funcs
+
+    @hmm_initialization_funcs.setter
+    def hmm_initialization_funcs(self, value: INITIALIZATION_FN_DICT | dict):
+        """Validate and set the dictionary of initialization functions for HMM parameters."""
+        self._hmm_initialization_funcs = _validate_init_funcs_keys(value)
```
BalzaniEdoardo, line 289: Does this validate the signature as well?

src/nemos/hmm/initialize_parameters.py
BalzaniEdoardo, line 596: it looks like it is returning a tuple of two elements, however the return tuple for string values is 

`VAILABLE_INIT_FUNCTIONS[key][value], kwargs, False`

and the same for the other branch because `_validate_custom_init_func` returns `func, kwargs, True`.

So I think all typing should be updated 


```suggestion
) -> Tuple[InitFunctionHMM, dict, bool]:
```

BalzaniEdoardo, line 671: same issue with the return type (there is an extra bool returned)

src/nemos/hmm/utils.py
```diff
@@ -2,12 +2,20 @@
 
 import jax
 import jax.numpy as jnp
-from numpy.typing import NDArray
+import pynapple as nap
```
BalzaniEdoardo, line 5: I think we should be able to importing lazily pynapple in the whole module. Most uses are for typing, and there is an `isinstance` that can be replaced with some  checks on the attributes to check if something is an interval set. 

You can check how I did the typing for the `label_encoder.py`:
- used TYPE_CHECKING for importing pynapple conditionally 
- you can use the string based typing `"nap.IntervalSet"` instead of directly `nap.IntervalSet`. 


in #500, there is an implementation description that tells you what I did for keeping the imports lazy and why.


```suggestion
import lazy_loader as lazy
```

then below add 

`nap = lazy.load("pynapple")`

BalzaniEdoardo: Once this is done you can add to the tests for lazy imports, I am changing that now because I was only testing module level imports, which are lazy independently because we are doing `_lazy.attach_stub(__name__, __file__)`

src/nemos/hmm/utils.py
```diff
@@ -35,17 +52,160 @@ def initialize_new_session(n_samples, is_new_session):
...
+    start = is_new_session.start
+    is_new_session = (
+        jax.numpy.zeros_like(time, dtype=bool)
+        .at[jax.numpy.searchsorted(time, start)]
+        .set(True)
+    )
     return is_new_session
+
+
+def shift_nan_is_new_session(
```
BalzaniEdoardo, line 155: the scan non-compiled is a bit slow, but. if we compile it we pay the jit compilation cost. I think a good compromise (not compilable) is this

```python
def shift_searchsorted(is_new_session: jnp.ndarray, is_nan: jnp.ndarray) -> jnp.ndarray:
    n = is_new_session.shape[0]
    valid = jnp.arange(n)[~is_nan]                               # non-NaN indices
    invalid_session_idx = jnp.arange(n)[is_nan & is_new_session] # misplaced markers

    # Append sentinel n: searchsorted returning len(valid) (no valid sample after
    # the markebuwr) then maps to n, which mode="drop" discards. The append copies
    # valid but avoids any clip/where logic and handles empty valid correctly.
    pos = jnp.searchsorted(valid, invalid_session_idx, side="left")
    shifted_idx = jnp.append(valid, n)[pos]

    return is_new_session.at[invalid_session_idx].set(False).at[shifted_idx].set(True, mode="drop")
```

if we don't use the shift in the algorithm but only at initialization, this one scales decently with problem size and doesn't require compilation

BalzaniEdoardo (commented):
Reviewed everything except the tests. looks very good. The only annoying change is the lazy loading of pynapple


I think it is worth it because if people do want to use nemos with array only they can avoid pynapple entirely.

tests/test_hmm_initialization.py
BalzaniEdoardo, line 1: We should add tests for: 
- has_nans_only_at_border
- shift_nan_is_new_session

tests/test_hmm_base_class.py
```diff
@@ -0,0 +1,1087 @@
...
+        BaseHMM.setup(
+            self,
+            initial_proba_init=initial_proba_init,
+            initial_proba_init_kwargs=initial_proba_init_kwargs,
+            transition_proba_init=transition_proba_init,
+            transition_proba_init_kwargs=transition_proba_init_kwargs,
+        )
+
+    def _check_model_is_fit(self):
+        BaseHMM._check_is_fit(self)
```
BalzaniEdoardo: wouldn't this create an infinite recursion since `BaseHMM._check_is_fit` calls `self._check_model_is_fit`?

tests/test_hmm_base_class.py
```diff
@@ -0,0 +1,1087 @@
...
+            jnp.ndarray, None
+        ] = None,  # (n_state, )
+        dirichlet_prior_alphas_transition: Union[
+            jnp.ndarray | None
+        ] = None,  # (n_state, n_state):
+        maxiter: int = 1000,
+        tol: float = 1e-8,
+        seed=jax.random.PRNGKey(123),
+        hmm_initialization_funcs: INITIALIZATION_FN_DICT = {},
+        model_initialization_funcs: INITIALIZATION_FN_DICT = {},
```
BalzaniEdoardo: I don't know if this is very important for a mock, but we are initializing kwargs with mutables here, better setting it to None, and add an 
```python
if model_initialization_funcs is None:
    model_initialization_funcs = {}
```

BalzaniEdoardo: for both parameters

src/nemos/hmm/utils.py
```diff
@@ -17,10 +25,19 @@ def initialize_new_session(n_samples, is_new_session):
...
+        Input data/design matrix, shape ``(n_samples, n_features)``. Used to infer
+        session boundaries if is_new_session is a pynapple.IntervalSet and y is not
+        a pynapple Tsd or TsdFrame.
+    y :
+        Output data/observations, shape ``(n_samples, n_observations)``. Used to get
+        n_samples and to infer session boundaries if is_new_session is a pynapple.IntervalSet.
     is_new_session :
-        Boolean array indicating session starts, shape ``(n_samples,)``.
+        Optional array indicating user-provided session boundaries. Can be:
+        - a boolean array indicating session starts, shape ``(n_samples,)``
```
BalzaniEdoardo: Maybe add that integer arrays of 0s and 1s of shape ``n_samples`` will be treated as boolean masks too.

tests/test_hmm_base_class.py
```diff
@@ -0,0 +1,1087 @@
...
+                np.ones((3,)),
+                nap.IntervalSet([0, 3]),
+                pytest.raises(
+                    TypeError,
+                    match="X or y must be a pynapple",
+                ),
+            ),
+        ],
+    )
+    def test_initialize_and_compute_new_session_errors(
```
BalzaniEdoardo, line 1194: add a integer array that repeats the same valid value twice. The code should work as is, but it's a good edge case

BalzaniEdoardo (changes requested):
Checked all the tests.

src/nemos/hmm/hmm.py
sjvenditto: As we discussed offline, we're not doing any checks here because this should only be set at init for compatibility with sklearn clone. I'll add in more disclaimers to make this clear to the user

src/nemos/hmm/hmm.py
sjvenditto: I added an example function in the doc string, but I can't add an examples section since this is an abstract base class that can't be initialized

src/nemos/hmm/utils.py
sjvenditto: updated the first bullet to indicate boolean and integer arrays of 1s and 0s

src/nemos/hmm/hmm.py
sjvenditto: As we discussed offline, I've shortened the names to reduce code changes

src/nemos/hmm/initialize_parameters.py
sjvenditto: good catch, this was an oversight after updating the function. I've fixed this

BalzaniEdoardo (approved):
LGTM

PR #553, opened 2026-04-17
Add common inference methods into `BaseHMM`

sjvenditto:
## Summary
This PR reintroduces the inference methods `score`, `smooth_proba`, `filter_proba`, and `decode_state` (and their private counterparts) into `BaseHMM`. These functions are very similar to the original GLM-HMM versions with a few changes:
- `is_new_session` has been added as an optional input argument
- examples were removed (they should be re-added by the inheriting model)
- calls to EM functions were updated to match the new signatures

The (probably) most important change: A new abstract method for `_log_likelihood` was added, which needs to be implemented in the inheriting class. This is what gets passed as the `log_likelihood_func` to the EM algorithms. For the GLM-HMM, this just means storing the output of `prepare_estep_log_likelihood` into this method

### DIFF
--- a/src/nemos/hmm/hmm.py
+++ b/src/nemos/hmm/hmm.py
@@ -3,14 +3,24 @@
 from __future__ import annotations
 
 import abc
+import warnings
 from numbers import Number
-from typing import Any, Callable, Optional, Tuple, Union
+from typing import Any, Callable, Literal, Optional, Tuple, Union
 
 import jax
 import jax.numpy as jnp
+import pynapple as nap
+from numpy.typing import ArrayLike, NDArray
 
+from .. import tree_utils
 from ..base_regressor import BaseRegressor
+from ..hmm.expectation_maximization import (
+    forward_backward,
+    forward_pass,
+    max_sum,
+)
 from ..regularizer import Regularizer
+from ..type_casting import support_pynapple
 from ..typing import (
     DESIGN_INPUT_TYPE,
 )
@@ -22,6 +32,7 @@ from .initialize_parameters import (
     setup_hmm_initialization,
 )
 from .params import HMMModelParamsT, HMMUserParams, HMMUserProvidedParamsT
+from .utils import _check_state_format
 from .validation import HMMValidator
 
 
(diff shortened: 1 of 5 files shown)


src/nemos/hmm/hmm.py
```diff
@@ -384,3 +395,347 @@ def _validate_and_prepare_inputs(self, X, y, is_new_session=None):
...
+        self,
+        X: Union[DESIGN_INPUT_TYPE, ArrayLike],
+        y: ArrayLike,
+        is_new_session: Optional[ArrayLike] = None,
+        score_type: Literal[
+            "log-likelihood", "pseudo-r2-McFadden", "pseudo-r2-Cohen"
+        ] = "log-likelihood",
+        null_model: Optional[Literal["constant", "glm"]] = None,
+    ) -> jnp.ndarray:
+        """Compute the model score."""
```
BalzaniEdoardo: should we have a complete docstring for this one?

src/nemos/hmm/hmm.py
```diff
@@ -384,3 +395,347 @@ def _validate_and_prepare_inputs(self, X, y, is_new_session=None):
...
+        The smoothing posteriors answer: "Given all observations, what is the probability
+        that the system was in state k at time t?"
+
+        Parameters
+        ----------
+        X
+            Predictors, shape ``(n_time_points, n_features)``.
+        y
+            Observations, shape ``(n_time_points,)`` for single observation or
+            ``(n_time_points, n_observations)`` for population.
```
BalzaniEdoardo, line 544: missing is new session

src/nemos/hmm/hmm.py
```diff
@@ -384,3 +395,347 @@ def _validate_and_prepare_inputs(self, X, y, is_new_session=None):
...
+
+        The filtering posteriors answer: "Given observations up to time t, what is the
+        probability that the system is in state k at time t?"
+
+        Parameters
+        ----------
+        X
+            Predictors, shape ``(n_time_points, n_features)``.
+        y
+            Observations, shape ``(n_time_points,)`` for single observation or
```
BalzaniEdoardo, line 639: missing is new session

src/nemos/hmm/hmm.py
```diff
@@ -384,3 +395,347 @@ def _validate_and_prepare_inputs(self, X, y, is_new_session=None):
...
+        The decoded states answer: "What is the most likely sequence of states that
+        generated the observed data?"
+
+        Parameters
+        ----------
+        X
+            Predictors, shape ``(n_time_points, n_features)``.
+        y
+            Observations, shape ``(n_time_points,)`` for single observation or
+            ``(n_time_points, n_observations)`` for population.
```
BalzaniEdoardo, line 749: missing is new session

src/nemos/hmm/hmm.py
```diff
@@ -384,3 +395,347 @@ def _validate_and_prepare_inputs(self, X, y, is_new_session=None):
...
+        ValueError
+            If the model has not been fit (``fit()`` must be called first).
+        ValueError
+            If inputs contain NaN values in the middle of epochs (only boundary NaNs allowed).
+        ValueError
+            If X and y have inconsistent shapes or features.
+
+        See Also
+        --------
+        smooth_proba : Compute smoothing posteriors (conditioned on all observations).
```
BalzaniEdoardo: maybe we should cross-ref these?

src/nemos/hmm/hmm.py
```diff
@@ -384,3 +395,347 @@ def _validate_and_prepare_inputs(self, X, y, is_new_session=None):
...
+        is_new_session = is_new_session.at[0].set(True)
+
+        # smooth with forward backward
+        _, log_norm = forward_pass(
+            params=params,
+            X=data,
+            y=y,
+            is_new_session=is_new_session,
+            # do we store this in the model during initialization?
+            log_likelihood_func=self._log_likelihood,
```
BalzaniEdoardo, line 429: as long as we make sure that it is updated any time one changes properties of the model that affect the function. 
What are the properties depends on the model: 
 - for HMMs I think it is only when the obs model changes. 
 - for GLM-HMMs it is that and the link function.


I think it is worth caching the function (instead of generating it on the fly all the time); generating it on the fly is safer (we can just have a property instead of an attribute that sets it up and returns it), but I am think that if we do that we will re-compile the `forward_pass` all the time.

BalzaniEdoardo (commented):
left a few comments

src/nemos/hmm/hmm.py
```diff
@@ -384,3 +395,347 @@ def _validate_and_prepare_inputs(self, X, y, is_new_session=None):
             X=X, y=y, is_new_session=is_new_session
         )
         return params, X, y, is_new_session
+
+    @abc.abstractmethod
+    def _log_likelihood(self, params: HMMModelParamsT, X, y):
```
BalzaniEdoardo: the signature doesn't match the expectation maximization one, I think this signature is the way to go, in the algorithm we have: `log_likelihood(X, y, params)` but everywhere else we start with the parameters

BalzaniEdoardo: so I would keep this as is change the call in the em

tests/test_hmm_base_class.py
```diff
@@ -1085,3 +1091,408 @@ def test_initialize_and_compute_new_session_errors(
...
+        # Add NaNs at specific indices
+        nan_indices = [0, 1, 2]
+        X_with_nan[nan_indices] = np.nan
+
+        # This should work - NaNs get filtered internally
+        posteriors = getattr(model, method_name)(X_with_nan, y_with_nan)
+
+        # Check that we get valid output (NaN rows filtered)
+        assert posteriors.shape[1] == model.n_states
+        # After filtering NaNs, shape[0] should be reduced
```
BalzaniEdoardo: this is not true, i think it is a leftover 
```suggestion
```

BalzaniEdoardo (changes requested):
I think it looks good, with some minor changes, once you address them, I think it can be merged

src/nemos/hmm/hmm.py
```diff
@@ -384,3 +395,347 @@ def _validate_and_prepare_inputs(self, X, y, is_new_session=None):
...
+        Notes
+        -----
+        - Viterbi decoding finds the globally optimal state sequence, not the sequence
+          of individually most likely states from ``smooth_proba()``
+        - This is a hard assignment (single best path) unlike probabilistic posteriors
+        - The algorithm properly handles session boundaries and NaN values at epoch borders
+        - Decoding is useful for segmenting continuous data into discrete behavioral states
+        - For uncertainty estimates about states, use ``smooth_proba()`` instead
+        """
+        params, X, y, is_new_session = self._validate_and_prepare_inputs(X, y)
```
BalzaniEdoardo: I think we must pass is_new_session down, otherwise it may be auto generated no matter what

src/nemos/hmm/hmm.py
```diff
@@ -384,3 +395,347 @@ def _validate_and_prepare_inputs(self, X, y, is_new_session=None):
...
+
+        Notes
+        -----
+        - Filtering provides causal state estimates suitable for online/real-time applications
+        - Smoothing provides better estimates but requires all data (non-causal)
+        - The algorithm properly handles session boundaries and NaN values at epoch borders
+        - NaN values are removed before inference, but session markers are preserved
+        - For pynapple inputs, the output TsdFrame has columns named "state_0", "state_1", etc.
+        """
+        params, X, y, is_new_session = self._validate_and_prepare_inputs(X, y)
```
BalzaniEdoardo: we should pass is_new_session

src/nemos/hmm/hmm.py
```diff
@@ -384,3 +395,347 @@ def _validate_and_prepare_inputs(self, X, y, is_new_session=None):
...
+        --------
+        filter_proba : Compute filtering posteriors (conditioned on past observations only).
+        decode_state : Compute most likely state sequence (Viterbi decoding).
+
+        Notes
+        -----
+        - Smoothing provides better state estimates than filtering because it uses all data
+        - The algorithm properly handles session boundaries and NaN values at epoch borders
+        """
+        params, X, y, is_new_session = self._validate_and_prepare_inputs(X, y)
```
BalzaniEdoardo: we should pass is_new_session

src/nemos/hmm/hmm.py
```diff
@@ -384,3 +395,347 @@ def _validate_and_prepare_inputs(self, X, y, is_new_session=None):
...
+            warnings.warn(
+                "The null model is not used for the log-likelihood computation.",
+                UserWarning,
+                stacklevel=2,
+            )
+        if score_type != "log-likelihood":
+            raise NotImplementedError(
+                f"score of type {score_type} not implemented yet!"
+            )
+        params, X, y, is_new_session = self._validate_and_prepare_inputs(X, y)
```
BalzaniEdoardo: we should pass is_new_session, this tells me that we are missing a test case showing that the provided is_new_session is actually used

BalzaniEdoardo (changes requested):
Add a test making sure that is_new_session, when provided it is actually used.

src/nemos/hmm/hmm.py
sjvenditto: I agree, this is something to keep in mind in the subclasses

PR #557, opened 2026-04-27
Glm hmm ref init

BalzaniEdoardo:
## Summary

Adds the GLM-HMM parameter initialization layer (`glm_hmm/initialize_parameters.py`), modeled after the HMM initialization design in the base PR. Provides GLM-specific initialization for coefficients, intercept, and scale; extends `KMeansInitializer` with a GLM-aware subclass; wires up a unified `setup_glm_hmm_initialization` / `generate_glm_hmm_initial_params` API; ships accompanying validation and tests. Several HMM-base-class utilities were extended to support GLM-specific protocols and key sets.

## Motivation

The HMM initialization infrastructure (base PR) provides `setup_hmm_initialization` and `generate_hmm_initial_params` for initial/transition probabilities, but GLM-HMM also needs per-state GLM parameters (coef, intercept) and scale. Reusing as much of the HMM machinery as possible (key registry, kwargs validation, KMeans initializer) avoids duplicating logic and keeps the two layers consistent.

### New file: `src/nemos/glm_hmm/initialize_parameters.py`

| Component | Description |
|---|---|
| `InitFunctionGLM` | Protocol for GLM init functions; extends `InitFunctionHMM` signature with `inverse_link_function` (4th arg, required) and `is_new_session` (5th arg, optional). |
| `random_glm_params_init` | Random small coefficients + intercept matching the mean firing rate. Decorated with `@cast_to_jax(dtype=None)` to preserve integer PRNG keys. |
| `KMeansInitializerGLM` | Inherits `KMeansInitializer`; adds `glm_params()` (fits per-state GLM after KMeans assignment) and `scale()` (extracts fitted scale, short-circuiting for fixed-scale observation models via `has_fixed_scale`). Supports pytree `X` and population `y`. |
| `kmeans_glm_params_init` | Instantiates `KMeansInitializerGLM` if no `initializer` is supplied, then returns `glm_params()`. |
| `kmeans_scale_init` | Same pattern for scale, reusing the same `KMeansInitializerGLM` instance when the caller supplies one. |
| `constant_scale_init` | Fills scale to a user-specified constant (default 1.0). Default scale initializer. |
| `GLM_INIT_FUNCS` | Registry dict (keys: `glm_params_init`, `scale_init`, plus `_kwargs` and `_custom` flags) mirroring the HMM dict layout. |
| `AVAIL_INIT_FUNCTIONS_GLM` | Available string aliases: `"random"`, `"kmeans"` for params; `"constant"`, `"kmeans"` for scale. |
| `_validate_custom_glm_init_func` | Validates a custom callable against `InitFunctionGLM`. Required because `_validate_custom_init_func` in the HMM module is hardcoded to `InitFunctionHMM`, which lacks `inverse_link_function`. |
| `_glm_resolve_init_funcs` | Dispatches to `_validate_custom_glm_init_func` for callables; passes `protocol=InitFunctionGLM` to `_validate_init_funcs_kwargs` for string lookups. Mirrors `_resolve_init_funcs` but for GLM keys. |
| `setup_glm_hmm_initialization` | Validates and stores GLM-specific init functions, delegating HMM-probability setup to `setup_hmm_initialization`. |
| `generate_glm_hmm_initial_params` | Splits the random key into three independent subkeys (GLM params, scale, HMM probs) and calls each init function in sequence. Returns `(coef, intercept, scale, initial_probs, transition_matrix)`. |

### Changes to `src/nemos/hmm/initialize_parameters.py`

- `_validate_init_funcs_keys`: accepts optional `default_init_dict` so the GLM-HMM layer can validate its own key set without coupling to `DEFAULT_INIT_FUNCTIONS`.
- `_resolve_init_funcs`: accepts `available_init_funcs` for the same reason.
- `_validate_init_funcs_kwargs`: accepts a `protocol` argument so GLM init functions are validated against `InitFunctionGLM` rather than `InitFunctionHMM`.
- `generate_hmm_initial_params`: return type annotated as `HMMUserParams`; random key split changed from 2 to 1 (see Open Question 1); added `_validate_custom_init_output` call for custom `transition_proba_init` functions.
- `InitFunctionHMM`: `random_key` annotation changed from `jax.random.PRNGKey` to `jax.Array`.

### Changes to `src/nemos/hmm/hmm.py`

`_hmm_params_initialization` updated to unpack the return of `generate_hmm_initial_params` as `hmm_params, _` (see Open Question 2).

### Changes to `src/nemos/glm_hmm/algorithm_configs.py`

Extracted `has_fixed_scale(observation_model)` helper; replaced the two inline `type(observation_model) in _NO_SCALE` checks with calls to it. Used in `KMeansInitializerGLM.scale()` to short-circuit fitting for fixed-scale observation models.

### Changes to `src/nemos/type_casting.py`

`cast_to_jax` extended to a decorator factory: `@cast_to_jax` (plain, `dtype=float`) and `@cast_to_jax(dtype=None)` (keep original dtypes). The `dtype=None` form is needed by `random_glm_params_init` and `constant_scale_init` so that integer PRNG keys are not promoted to float.

### Changes to `src/nemos/validation.py`

`_suggest_keys` type hint widened from `List[str]` to `Collection[str]` to accept dict key views.

### Changes to `scripts/check_parameter_naming.py`

Added valid name pairs introduced by this module: `glm_params_init`/`glm_params_init_kwargs`, `scale_init_kwargs`/`solver_init_kwargs`, and `_func`/`func`/`funcs`.

### New test file: `tests/test_glm_hmm_initialization.py`

Covers:
- `random_glm_params_init`: output shapes (single/population, varying states), output types, randomization across seeds, `std_dev` parameter, intercept-to-mean-rate correctness.
- `KMeansInitializerGLM` and `kmeans_glm_params_init`/`kmeans_scale_init`: shape correctness, reuse of a shared `initializer`, fixed-scale short-circuit.
- `setup_glm_hmm_initialization` and `generate_glm_hmm_initial_params`: valid/invalid key detection, reserved-param rejection, custom function validation (signature check against `InitFunctionGLM`), all four init-function slots (individually, in pairs, all at once, mixed).
- Mock templates use `create_autospec` with 6 mandatory params for GLM functions (`n_states, X, y, inverse_link_function, is_new_session, random_key`) and 5 for HMM functions. Parametrized at class level over function name.

- **`_validate_custom_glm_init_func` / `_glm_resolve_init_funcs` duplication**: These are local GLM-side copies of `_validate_custom_init_func` / `_resolve_init_funcs` from the HMM module, with `InitFunctionGLM` substituted. The cleaner alternative would be to make the originals accept `protocol` and `available_init_funcs` arguments (as `_validate_init_funcs_kwargs` and `_resolve_init_funcs` already do in part) and call them directly. Worth unifying?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,7 +25,7 @@ on:
 
 env:
   CODECOV_OS: ubuntu-latest
-  CODECOV_PY: '3.13'
+  CODECOV_PY: '3.14'
 
 jobs:
   tox_tests_not_solver_related:
@@ -33,7 +33,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
-        python-version: ['3.10', '3.11', '3.12', '3.13']
+        python-version: ['3.12', '3.13', '3.14']
       fail-fast: false
     runs-on: ${{ matrix.os }}
     steps:
@@ -67,7 +67,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
-        python-version: ['3.10', '3.11', '3.12', '3.13']
+        python-version: ['3.12', '3.13', '3.14']
       fail-fast: false
     runs-on: ${{ matrix.os }}
     steps:
@@ -101,7 +101,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
-        python-version: ['3.10', '3.11', '3.12', '3.13']
+        python-version: ['3.12', '3.13', '3.14']
       fail-fast: false
     runs-on: ${{ matrix.os }}
     steps:
@@ -142,7 +142,7 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405 #v6.2.0
         with:
-          python-version: '3.13'
+          python-version: '3.14'
 
       - name: Set up uv
         uses: astral-sh/setup-uv@5a095e7a2014a4212f075830d4f7277575a9d098 #v7.3.1
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -8,7 +8,7 @@ version: 2
 build:
   os: ubuntu-22.04
   tools:
-    python: "3.11"
+    python: "3.12"
     ruby: "3.3"
   jobs:
     pre_build:
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -64,6 +64,7 @@ dev = [
     "hmmlearn",                     # Test algorithmic implementations of HMM
     "tox-uv",                       # Run tox with uv for faster installs
     "fsspec",
+    "tqdm",
 ]
 docs = [
     "pynapple>=0.11",               # Required for nap.compute_event_triggered_average(...) in a tutorial
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -8,17 +8,16 @@ dynamic = ["version"]
 authors = [{name = "nemos authors"}]
 description = "NEural MOdelS, a statistical modeling framework for neuroscience."
 readme = "README.md"
-requires-python = ">=3.10,<3.14"
+requires-python = ">=3.12,<3.15"
 keywords = ["neuroscience", "Poisson-GLM"]
 license = { file = "LICENSE" }
 classifiers = [
     "Intended Audience :: Science/Research",
     "Natural Language :: English",
     "Programming Language :: Python :: 3",
-    "Programming Language :: Python :: 3.10",
-    "Programming Language :: Python :: 3.11",
     "Programming Language :: Python :: 3.12",
-    "Programming Language :: Python :: 3.13"
+    "Programming Language :: Python :: 3.13",
+    "Programming Language :: Python :: 3.14"
 ]
 
 # Define dependencies for the project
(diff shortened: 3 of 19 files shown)


sjvenditto:
Looking great! Some changes I added in (that we discussed offline):
- In `BaseHMM`:
     - Re-split `initialization_funcs` into `hmm_initialization_funcs` and `model_initialization_funcs` to keep separate logic (add added in their getters and setters)
     - Implemented private method `_setup_hmm()` that gets called instead of `setup()` when setting `hmm_initialization_funcs`. This methods also updates the internal flag `_hmm_use_kmeans`, which is a dictionary tracking any parameters set to be initialized with kmeans
     - Added abstract method for `_setup_model()` that gets called when setting `model_initialization_funcs`. This method needs to update the internal flag `_model_use_kmeans` in the child class. 
     - Changed `setup()` to call both `_setup_hmm()` and `_setup_model()`. Works as a default case but should be redefined in child class for better docstrings
     - Updated kmeans initializer to pull extra kwargs from `model_initialization_funcs` only (hmm parameter methods don't have kwargs for setting up the initializer). new/updated methods/properties: `_use_kmeans`, `_kmeans_resolve_model_kwargs`, `_kmeans_setup_initializer`
- in `hmm/initialize_parameters.py`
     - removed `minimum_prob` as an init parameter of `KMeansInitializer` and instead a parameter of the init_prob and transition_prob methods themselves
     - updated `_resolve_init_funcs` to take function protocol to pass to validation funcs
     - updated `_validate_custom_init_func` to take function protocol
- in `glm_hmm/initialize_parameters.py`
     - removed `_validate_custom_glm_init_func` and `_glm_resolve_init_funcs` to use generic versions updated above
     - removed initialization of hmm parameters in `setup_glm_hmm_initialization`
- also added new tests for kmeans initializer setup and making sure sklearn is only called once

An additional test you could add is verifying that the glm fits are called the appropriate amount of times for kmeans initialization of glm parameters and scale. Otherwise I think this is good to go!

BalzaniEdoardo:
this LGTM, I'll merge after the tests pass, thanks @sjvenditto

PR #558, opened 2026-05-03
Newton solvers

wulfdewolf:
# Summary

This PR introduces a native Newton solver including support for analytic & autodiff Hessians.

The implementation adds a fully integrated second-order optimization path for `GLM` and `PopulationGLM` as well as `ClassifierGLM` and `PopulationClassifierGLM`.

# Main additions

## Newton solver

Added a new `Newton` solver implementation in `solvers/_newton.py`.

Features include:
- Newton updates using explicit Hessian solves
- optional analytic Hessians
- fallback to autodiff Hessians when no analytic Hessian is provided
- backtracking line search via `optax`
- support for both JIT and non-JIT execution

The solver internally uses:
- Cholesky solves for positive definite systems
- SVG solves for positive semi definite systems
- LU solves for general systems

## Hessian tagging system

Added `solvers/_hess.py`, introducing metadata tags describing Hessian structure and matrix properties.

### Supported matrix structures
- `Diagonal`
- `BlockDiagonal`
- `Full`

### Supported matrix properties
- `General`
- `Symmetric`
- `PositiveDefinite`
- `PositiveSemiDefinite`
- `NegativeDefinite`

This metadata is used to:
- select appropriate linear solvers
- exploit Hessian structure
- combine structure information across objectives (e.g. loss + regularizer)

The PR also adds utilities for:
- combining Hessian structures
- combining matrix properties
- combining full Hessian tags for additive objectives

# Analytic GLM Hessians

Added analytic Fisher-scoring Hessians for:
- `GLM`
- `PopulationGLM`

The Hessians are computed analytically using:
- the inverse link derivative
- observation variance functions
- optional ridge regularization contributions

This avoids expensive autodiff Hessian evaluations and enables efficient Newton updates.

`ClassifierGLM` is also supported, but uses autodiff by default.
For the population version we avoid the full hessian by creating blocks of the autodiff'd individual ones.

## GLM Hessian structure

`GLM` Hessians are tagged as:
- `Full`
- `PositiveDefinite`

`ClassifierGLM` Hessians are tagged as:
- `Full`
- `PositiveSemiDefinite`

## PopulationGLM Hessian structure

`PopulationGLM` Hessians are implemented as block-diagonal systems, with one block per neuron.

These are tagged as:
- `BlockDiagonal`
- `PositiveDefinite`

`ClassifierPopulationGLM` Hessians are tagged as:
- `BlockDiagonal`
- `PositiveSemiDefinite`

The Newton solver detects this structure and performs per-block solves efficiently using `vmap`.

# Regularizer integration

Added Hessian metadata to regularizers:
- `UnRegularized`
- `Ridge`

Unregularized just uses the model property.
Ridge contributes:
- diagonal structure
- positive definite curvature

This information is automatically combined with model Hessian structure when constructing the full penalized objective Hessian.

# Automatic Hessian reconfiguration

`BaseRegressor` now includes:
- `_get_hess_fn`
- `_invalidate_solver`

When the regularizer/regularizer strength/observation model is updated, the solver is invalidated.
Calling `update` at that point will throw a `RuntimeError`.
To fix the state, either `fit` or `initialize_optimizer_and_state` resets the hessian.

# Solver registry updates

Added `Newton` to:
- the public solver API
- the solver registry
- shared solver test infrastructure

`Newton` is now an available solver for:
- `UnRegularized`
- `Ridge`

# Tree utilities

Added `ravel_pytree_nest` to `tree_utils.py`:
- supports batched pytrees with batch-last semantics
- supports reconstruction of flattened parameter vectors back into pytree form

This is used internally by the Newton solver for structured parameter flattening.

# Observation model support

Added variance-function mappings for:
- Poisson
- Gamma
- Gaussian
- Bernoulli

These are used to construct analytic Fisher-scoring Hessians.

# Tests

Added extensive Newton solver tests covering:
- convergence on linear regression problems
- convergence on ridge regression problems
- initialization behavior
- solver state behavior
- GLM integration
- solver kwarg propagation
- analytic Hessian updates
- observation-model Hessian updates
- regularizer Hessian updates
- regularizer-strength Hessian updates
- block-diagonal `PopulationGLM` behavior
- equality of autodiff and anayltic hessians
- correct regularization in the hessians

Also added Newton to the shared solver compatibility tests.

# Questions for reviewers
Should we set Newton as the default solver for `GLM`/`PopulationGLM`?
The benchmarks (see #551) suggest that it is faster for almost everything.


src/nemos/solvers/_newton.py
```diff
@@ -0,0 +1,443 @@
...
+import jax
+import jax.numpy as jnp
+from jax.flatten_util import ravel_pytree
+
+import lineax
+import optax
+from ._abstract_solver import AbstractSolver, OptimizationInfo, SolverAdapterState
+
+
+class NewtonState(NamedTuple):
```
BalzaniEdoardo: I prefer an equinox module for a state, just because the repr is slightly better, what do you think?

src/nemos/solvers/_newton.py
```diff
@@ -0,0 +1,443 @@
...
+        :attr:`linear_solver`.  Subclasses may override this to enforce
+        structural properties (e.g. symmetry) before factorisation.
+        """
+        operator = lineax.MatrixLinearOperator(H)
+        solution = lineax.linear_solve(operator, -g_flat, solver=self.linear_solver)
+        return solution.value
+
+    def run(self, init_params, *args) -> tuple[Any, NewtonState]:
+        func = self.func
+        val_and_grad = jax.value_and_grad(func)
```
BalzaniEdoardo: why don't we do this at initialization and store val_and_grad?

src/nemos/solvers/_newton.py
```diff
@@ -0,0 +1,443 @@
...
+            ls_init_state,
+            jnp.array(0),
+            jnp.array(False),
+            jnp.linalg.norm(g0_flat),
+        )
+
+        # ------------------------------------------------------------------
+        # Run loop (optionally JIT-compiled)
+        # ------------------------------------------------------------------
+        while_loop = jax.lax.while_loop
```
BalzaniEdoardo: consider using https://github.com/patrick-kidger/equinox/blob/main/equinox/internal/_loop/loop.py

```python
import equinox
eqx.internal.while_loop
```

For compiled loops, the default kind should be  None. Check the FISTA implementation, I think we expose the while loop kind.

src/nemos/solvers/_newton.py
```diff
@@ -0,0 +1,443 @@
...
+
+    linear_solver: lineax.AbstractLinearSolver = lineax.LU()
+
+
+# ---------------------------------------------------------------------------
+# GLM-specific analytic Hessian factory
+# ---------------------------------------------------------------------------
+
+
+def _elementwise_derivative(f: Callable) -> Callable:
```
BalzaniEdoardo: Unsure if this belongs here. We probably should have a utility script for, like autodiff utilities.

src/nemos/solvers/_newton.py
```diff
@@ -0,0 +1,443 @@
...
+
+    @wraps(f)
+    def df(x):
+        _, grad = jax.jvp(f, (x,), (jnp.ones_like(x),))
+        return grad
+
+    return df
+
+
+def _var_func_of_mu(model) -> Callable:
```
BalzaniEdoardo: this is also to be moved in a script in the glm module

src/nemos/solvers/_newton.py
```diff
@@ -0,0 +1,443 @@
...
+        "GammaObservations": lambda mu: mu**2,
+        "GaussianObservations": lambda mu: jnp.ones_like(mu),
+        "BernoulliObservations": lambda mu: mu * (1.0 - mu),
+    }
+    if obs_name not in var_funcs:
+        raise NotImplementedError(f"No variance function defined for {obs_name!r}")
+    return var_funcs[obs_name]
+
+
+def define_hess(model, regularizer_strength: float = 0.0) -> Callable:
```
BalzaniEdoardo: same, this is in a glm module

src/nemos/solvers/_newton.py
```diff
@@ -0,0 +1,443 @@
...
+    def init_state(self, init_params, *args) -> NewtonState:
+        return NewtonState(iter_num=0, converged=False, grad_norm=float("inf"))
+
+    def run(self, init_params, *args) -> tuple[Any, NewtonState, None]:
+        params, state = self._solver.run(init_params, *args)
+        return params, state, None
+
+    def update(self, params, state, *args):
+        raise NotImplementedError(
+            "NewtonCholeskydoes not support incremental updates; use run()."
```
BalzaniEdoardo:
```suggestion
            "NewtonCholesky does not support incremental updates; use run()."
```

BalzaniEdoardo: why is that? we should totally support a single step. It is actually needed in some optimization procedures. Sometimes you need to take a full newton step + linesearch.

BalzaniEdoardo:
**On linesearch exposure:** I don't think we should expose the linesearch. 99% of users would never touch it, and the surface area for breakage is large. There are other options worth considering (e.g. zoom line search), but I'd keep it internal for now. Would be curious what @bagibence thinks.

**On Hessian placement and the broader solver design:**

This is the trickiest point. All GLM-specific Hessian machinery should live in the GLM module — the solver class itself shouldn't own it. My proposal:

- Remove `hess` from the solver's `__init__` parameters entirely. Instead, add a `setup_hessian(hess_fn, hess_tag)` method on the solver that gets called *after* initialization.
- If we want to let users force autodiff, add a `force_autodiff_hessian: bool = False` flag to the solver init. Default behavior: use whatever the model provides; fall back to autodiff if the model returns `None`.

**On unifying the Newton variants:**

If the only thing that differentiates the Newton solver classes is the linear solve type, we should collapse them into a single `Newton` class and treat the solve variant the same way we treat the Hessian — as something the model configures, not the user. No need to expose this choice at the API level.

My suggestion is a class attribute on `BaseRegressor` (e.g. `_hessian_tag`, default `None`) that holds the lineax tag directly when applicable. `_instantiate_solver` then uses this to wire everything up:

```python
# Concrete method — provides an analytic Hessian callable for models that have one
# (e.g. GLM), or returns None to fall back to autodiff.
def _hess_func(self, *args, **kwargs):
    return None

def _instantiate_solver(self, ...):
    ... # standard instantiation machinery
    if isinstance(solver, NewtonSolverProtocol):
        solver.setup_hessian(self._hess_func, self._hessian_tag)
```

For the `isinstance` check to work properly, we should introduce a `NewtonSolverProtocol` that extends `SolverProtocol` rather than adding `setup_hessian` to the general protocol. Adding it to `SolverProtocol` would force every custom solver implementation to define a method that's only meaningful for Newton-type solvers, which violates interface segregation. The extension would look like:

```python
@runtime_checkable
class NewtonSolverProtocol(SolverProtocol[SolverState], Protocol, Generic[SolverState]):
    def setup_hessian(
        self,
        hess_fn: Callable | None,
        hess_tag: lineax.AbstractLinearOperator | None,
    ) -> None: ...
```

This is cleaner than a `hasattr` check: it's explicit, statically analyzable, and makes the contract clear to anyone implementing a custom Newton-type solver.

On the GLM side: at init, and whenever `regularizer_strength` or the observation model changes, we call a method that checks whether the active solver is a `NewtonSolverProtocol` instance and, if so, updates the Hessian accordingly. This keeps all GLM-specific logic inside the GLM and gives a clean extension point for other model classes down the line.


**Hessian Tag and Regulatizer: Extra considerations**

The hessian type is dependent on the model **and** the regularizer **and** how the regularizer acts on the loss.

Currently, all our regularizers are additive: `loss + alpha * regularizer` is always the case, so the hessian of the loss and the regularizer adds up. There is a clean way to resolve the hessian tag of a model with a regularizer in this scenario:

If I am not mistaken here is how the tags are related:


```
unit_diagonal → diagonal → symmetric
                diagonal → lower_triangular
                diagonal → upper_triangular
positive_semidefinite → symmetric
negative_semidefinite → symmetric
```

If this is right, we can have an operator that gets two tags (one from the model and one from the regularizer) and spits out the appropriate tag. This operator can be used by the `_instantiate_solver` to resolve the hessian type on the fly. Below a sketch of this operator:

```python
IMPLIES = {
    lx.unit_diagonal_tag:        {lx.diagonal_tag, lx.symmetric_tag, lx.lower_triangular_tag, lx.upper_triangular_tag},
    lx.diagonal_tag:             {lx.symmetric_tag, lx.lower_triangular_tag, lx.upper_triangular_tag},
    lx.positive_semidefinite_tag:{lx.symmetric_tag},
    lx.negative_semidefinite_tag:{lx.symmetric_tag},
    lx.symmetric_tag:            set(),
    lx.lower_triangular_tag:     set(),
    lx.upper_triangular_tag:     set(),
}
NOT_ADDITIVE = {lx.unit_diagonal_tag}

def _expand(tag):
    return ({tag} | IMPLIES.get(tag, set())) - NOT_ADDITIVE

def combine_hessian_tags(tag1, tag2):
    if tag1 is None or tag2 is None:
        return None
    combined = _expand(tag1) & _expand(tag2)
    # drop tags implied by something more specific already in the set
    return frozenset(t for t in combined if not any(t in IMPLIES.get(s, set()) for s in combined))
```

BalzaniEdoardo (changes requested):
Looking good, a few things to discuss but the solver is nice and clean

src/nemos/solvers/_newton.py
```diff
@@ -0,0 +1,482 @@
...
+        """
+        self._hess_fn = hess_fn
+        self._hess_tag = hess_tag
+
+        # Select linear solver based on Hessian structure tag
+        if hess_tag is lineax.positive_semidefinite_tag:
+            self._linear_solver = lineax.Cholesky()
+        else:
+            # Default to LU for general or unspecified Hessians
+            self._linear_solver = lineax.LU()
```
BalzaniEdoardo: One thing worth thinking about before this solidifies: the current tag-based dispatch works for GLM, but will need revisiting for `PopulationGLM` where the Hessian is block-diagonal (one block per neuron). In that case we'd want to vmap a per-block solve over a `(n_neurons, n_features, n_features)` tensor rather than assembling and solving a full `(n_neurons * n_features, n_neurons * n_features)` system.

The flat tag approach struggles here because it can't express nesting — we need to say not just *"block diagonal"* but *"block diagonal whose blocks are positive semidefinite"* (i.e. use Cholesky per block). A small recursive structure type handles this naturally:

```python
class HessianStructure: pass

@dataclass(frozen=True)
class PositiveDefinite(HessianStructure): pass

@dataclass(frozen=True)
class BlockDiagonal(HessianStructure):
    block_structure: HessianStructure
```

Models declare their structure:
```python
class GLM(...):
    _hessian_structure = PositiveDefinite()

class PopulationGLM(...):
    _hessian_structure = BlockDiagonal(block_structure=PositiveDefinite())
```

And the solver dispatches recursively:
```python
def newton_solve(H, g, structure):
    match structure:
        case BlockDiagonal(block_structure=inner):
            return jax.vmap(lambda h, gi: newton_solve(h, gi, inner))(H, g)
        case PositiveDefinite():
            return lx.linear_solve(lx.MatrixLinearOperator(H), g, lx.Cholesky()).value
        case _:
            return lx.linear_solve(lx.MatrixLinearOperator(H), g, lx.LU()).value
```

Not necessarily a blocker for this PR if `PopulationGLM` support is out of scope here, but worth keeping in mind so the tag machinery doesn't get load-bearing before we have a chance to replace it.

BalzaniEdoardo (commented):
another idea for the netwon solve

BalzaniEdoardo:
Just posting here since I do not know where to keep track of this. A sketch of the tagging system that includes the structure (block diag vs diag vs full... we could add stuff here, like block tri-diag if it ever happens), and the property of the hessian: sym vs pos/neg semidef. 

Then we can create an operator to resolve how this combines under the assumption of adding two hessians.


```python
from dataclasses import dataclass


# --- Properties ---

class MatrixProperty:
    pass

class General(MatrixProperty):
    pass

class Symmetric(MatrixProperty):
    pass

class PositiveDefinite(MatrixProperty):
    pass

class NegativeDefinite(MatrixProperty):
    pass


PROPERTY_IMPLIES: dict[type, set[type]] = {
    PositiveDefinite: {Symmetric, General},
    NegativeDefinite: {Symmetric, General},
    Symmetric:        {General},
    General:          set(),
}


def _expand_property(p) -> set[type]:
    cls = p if isinstance(p, type) else type(p)
    return {cls} | PROPERTY_IMPLIES.get(cls, set())


def combine_property(p1, p2) -> type | None:
    common = _expand_property(p1) & _expand_property(p2)
    if not common:
        return None
    most_specific = [p for p in common if not any(p in PROPERTY_IMPLIES.get(q, set()) for q in common)]
    return most_specific[0] if len(most_specific) == 1 else None


# --- Structures ---

class MatrixStructure:
    pass

class Full(MatrixStructure):
    pass

class BlockDiagonal(MatrixStructure):
    pass

class Diagonal(MatrixStructure):
    pass


_STRUCTURE_GENERALITY: dict[type, int] = {
    Diagonal:      0,
    BlockDiagonal: 1,
    Full:          2,
}


def combine_structure(s1, s2) -> type:
    c1 = s1 if isinstance(s1, type) else type(s1)
    c2 = s2 if isinstance(s2, type) else type(s2)
    return max(c1, c2, key=lambda t: _STRUCTURE_GENERALITY[t])


# --- Combined tag ---

@dataclass(frozen=True)
class HessianTag:
    structure: type  # a MatrixStructure subclass
    property: type   # a MatrixProperty subclass


def combine_hessian_tags(t1: HessianTag | None, t2: HessianTag | None) -> HessianTag | None:
    """Combine tags assuming additivity.

    Valid when the total objective is a sum of two functions (e.g. loss + regularizer),
    since the Hessian of a sum is the sum of the Hessians.
    """
    if t1 is None or t2 is None:
        return None
    prop = combine_property(t1.property, t2.property)
    if prop is None:
        return None
    return HessianTag(
        structure=combine_structure(t1.structure, t2.structure),
        property=prop,
    )
```

BalzaniEdoardo:
Hi Wolf, this is looking great. I have a single concern that I think must be addressed before we merge the PR. Fixing it will solve the ClassifierGLM case too.


## Summary

`_setup_hessian()` is invoked from four places. Only one of them — inside `_instantiate_solver` — is reachable on a path that produces a coherent solver state. The other three (the `regularizer`, `regularizer_strength`, and `observation_model` setters) are de facto dead, and when they *do* fire post-init they leave the solver in an inconsistent state where the Hessian reflects new configuration but the loss does not.

Removing them frees the `_get_hess_fn` signature to accept `params`, which is what `ClassifierPopulationGLM` (and any future autodiff-based block Hessian) needs in order to close over the regularizer correctly.

## Where `_setup_hessian` is invoked today

| # | Site | Path | Notes |
|---|------|------|-------|
| 1 | `base_regressor.py:422` | `_instantiate_solver` | Direct `solver.setup_hessian(...)`. Always runs pre-fit when a `NewtonSolverProtocol` is built. |
| 2 | `base_regressor.py:258` | `regularizer.setter` | Calls `_setup_hessian()`. |
| 3 | `base_regressor.py:268` | `regularizer_strength.setter` | Calls `_setup_hessian()`. |
| 4 | `glm.py:384`, `glm.py:391` | `observation_model.setter` | Calls `_setup_hessian()`. |

`_setup_hessian()` itself short-circuits unless `self.solver` already exists and is a `NewtonSolverProtocol`. During `__init__`, all three setters fire before `self.solver` exists, so the calls no-op. The only scenario in which they do work is a setter mutation *after* a successful `initialize_optimizer_and_state`/`fit`.

## Why the post-init setter branches are incoherent

`_instantiate_solver` bakes the regularizer, regularizer strength, and unregularized loss into `self.fun` at solver construction:

```python
loss_fn = regularizer.penalized_loss(
    unregularized_loss,
    params=init_params,
    strength=regularizer_strength,
)
```

After construction, mutating `regularizer`, `regularizer_strength`, or `observation_model` does **not** rebuild `self.fun`. The setter-driven `_setup_hessian()` refreshes only `_hess_fn` / `_hess_tag` on the existing solver. A subsequent `update()` would therefore optimize a mismatched objective: new Hessian, stale loss.

A setter cannot rebuild the solver itself — `_instantiate_solver` needs `init_params` (and downstream, `X`/`y` for `init_state`), none of which the setter has access to. The only thing a setter can soundly do is **invalidate the cached solver** so the next operation is forced to re-run `initialize_optimizer_and_state` from scratch.

Note that the cached solver is only load-bearing on the `update()` path. `fit()` always calls `initialize_optimizer_and_state` and rebuilds the solver from scratch, so mutating a setter between two `fit()` calls is already safe. The staleness only leaks when a caller threads a setter mutation between `initialize_optimizer_and_state` and a subsequent `update()`.

## Existing tests pin the partial-refresh contract

`tests/test_newton.py` has three tests that assert the partial-refresh behavior:

- `test_newton_glm_set_regularizer_update_hessian` (lines 165–183)
- `test_newton_glm_set_regularizer_strength_update_hessian` (lines 187–207)
- `test_newton_glm_set_observation_model_update_hessian` (lines 229–248)

Each calls `initialize_optimizer_and_state`, mutates a setter, and then asserts that `solver._solver._hess_fn(params, X, y)` matches `glm._get_hess_fn()(params, X, y)`. They never run a fit after the mutation, so they never observe the loss/Hessian mismatch. They lock in the contract that is *exactly the bug*: "Hessian gets refreshed, loss does not."

These three tests should be rewritten in two parts: (a) assert that after a setter mutation the cached solver has been invalidated (i.e. `update()` raises or `self._solver` is `None`), and (b) assert that a fresh `initialize_optimizer_and_state` after the mutation produces a coherent solver (both `self.fun` and `_hess_fn` reflect the new configuration).

## Proposed API change

Change the signature of `_get_hess_fn` to accept `params`:

```python
# base_regressor.py
def _get_hess_fn(self, params) -> Callable | None:
    """Analytic Hessian callable for models that have one.

    `params` is the initial parameter pytree — passed so subclasses can close over regularizer state (mask shape, structured strength) that depends on the parameter structure. Returns None to fall back to autodiff.
    """
    return None
```

`_instantiate_solver` becomes the single entry point:

```python
# base_regressor.py, inside _instantiate_solver
if isinstance(solver, NewtonSolverProtocol):
    solver.setup_hessian(
        self._get_hess_fn(init_params),
        self._hess_tag,
        self.regularizer,
    )
```

The `regularizer`, `regularizer_strength`, and `observation_model` setters drop their `_setup_hessian()` calls and instead invalidate the cached solver:

```python
def _invalidate_solver(self):
    self._solver = None
    self._optimizer_init_state = None
    self._optimizer_update = None
    self._optimizer_run = None
    self._solver_loss_fun = None
```

(exact attribute set TBD — match whatever `_instantiate_solver` writes). Each setter calls `_invalidate_solver()` once the new value has been validated. `_setup_hessian` itself can be inlined into `_instantiate_solver` or kept as a private helper called only from there.

### Behavior change to flag

Users who currently mutate `regularizer`, `regularizer_strength`, or `observation_model` on an already-initialized model and then call `update()` directly — without re-running `initialize_optimizer_and_state` — will now get an explicit "solver not initialized" error instead of silently optimizing a mismatched objective. `fit()`-only workflows are unaffected, since `fit()` always rebuilds the solver.

## Example: block-diagonal autodiff Hessian for `ClassifierPopulationGLM`

With `_get_hess_fn(params)` available, the per-neuron penalized loss closure can be built from the real init params instead of going through `self._validator.get_empty_params()`. The structure of `init_params` already encodes everything `regularizer.penalized_loss` needs (coefficient shape for mask init in `GroupLasso`, per-leaf strength broadcasting, etc.).

```python
# src/nemos/glm/classifier_glm.py — ClassifierPopulationGLM

def _get_hess_fn(self, params):
    # extract a single-neuron slice of the init params; this is the shape the per-neuron penalized loss closure will see at call time
    coef_neu = jax.tree_util.tree_map(lambda x: x[:, 0], params.coef)
    params_neu_template = self._validator.to_model_params(
        [coef_neu, params.intercept[0]]
    )

    # build the per-neuron penalized loss once; the regularizer captures filter_kwargs (mask, structured strength) from params_neu_template
    penalized_single = self.regularizer.penalized_loss(
        super()._compute_loss,
        params_neu_template,
        strength=self.regularizer_strength,
    )
    hess_single = jax.hessian(penalized_single)

    def hess(params, *args):
        X = args[0]
        n_neurons = params.intercept.shape[0]
        coef = params.coef
        if self._feature_mask is not None:
            coef = coef * self._feature_mask

        blocks = []
        for neuron_idx in range(n_neurons):
            if self._feature_mask is not None:
                mask = self._feature_mask[:, neuron_idx]
                X_neuron = X * mask[None, :]
            else:
                X_neuron = X

            coef_neu = jax.tree_util.tree_map(lambda x: x[:, neuron_idx], coef)
            params_neu = self._validator.to_model_params(
                [coef_neu, params.intercept[neuron_idx]]
            )
            blocks.append(hess_single(params_neu, X_neuron, *args[1:]))

        return jnp.stack(blocks)

    return hess
```

The single-neuron closure builds the penalty filter kwargs once, from the real param structure. Per-neuron blocks reuse it with sliced coefficients and a masked design matrix, matching the prediction function exactly. The result is a `(n_neurons, d, d)` stack that the Newton solver can consume via the existing `BlockDiagonal` Hessian tag.

src/nemos/tree_utils.py
```diff
@@ -227,3 +228,51 @@ def drop_nans(*trees):
...
+        per_sample = x_flat.reshape(batch_size, -1)
+
+        samples = jax.vmap(unravel)(per_sample)
+
+        return jax.tree_util.tree_map(
+            lambda x: jnp.moveaxis(x, 0, -1) if x.ndim > 0 else x,
+            samples,
+        )
+
+    return flat, unravel_nest
```
BalzaniEdoardo:
```suggestion

def ravel_pytree_nest(pytree):
    leaves = jax.tree.leaves(pytree)
    batch_dims = [x.shape[-1] for x in leaves if x.ndim > 0]
    if not batch_dims or not all(b == batch_dims[0] for b in batch_dims):
        return ravel_pytree(pytree)

    N = batch_dims[0]
    in_axes = jax.tree.map(lambda x: -1 if x.ndim > 0 else None, pytree)
    sample0 = jax.tree.map(lambda x: jnp.take(x, 0, axis=-1) if x.ndim > 0 else x, pytree)
    _, unravel_one = ravel_pytree(sample0)

    flat = jax.vmap(lambda t: ravel_pytree(t)[0], in_axes=(in_axes,))(pytree).reshape(-1)

    out_axes = jax.tree.map(lambda x: -1 if x.ndim > 0 else 0, sample0)

    def unravel(x_flat):
        return jax.vmap(unravel_one, out_axes=out_axes)(x_flat.reshape(N, -1))

    return flat, unravel
```

I think this may work, it vmaps the default `ravel_pytree`.

src/nemos/solvers/_newton.py
```diff
@@ -0,0 +1,331 @@
...
+        self.tol = tol
+
+        self._hess_tag: HessianTag | None = None
+        self._hess_fn: Callable | None = None
+
+        self._line_search = optax.scale_by_backtracking_linesearch(
+            max_backtracking_steps=30
+        )
+
+        self._linear_solver = lx.LU()
```
BalzaniEdoardo:
```suggestion
```

BalzaniEdoardo:
Three findings on `_Newton.update` and `_Newton.run` that are independent of the `_get_hess_fn` signature change and the `ravel_pytree_nest` rewrite. They sit in the same file and probably want to land in adjacent PRs.

## 1. Lift `value_and_grad` to a cached attribute

`solvers/_newton.py:121-124`:

```python
params_flat, unravel, f = self._flat(params, args)
fval = f(params_flat)
g_flat = jax.grad(f)(params_flat)
```

Two issues per `update` call:

- **Two forward passes.** `f(params_flat)` then `jax.grad(f)(params_flat)` — the latter retraces the forward pass to seed the backward. `jax.value_and_grad(f)` shares primals → one fwd + one bwd, single trace.
- **`f` is rebuilt every call.** `_flat` closes over `args`, so `f` is a fresh closure every iteration. Under jit it traces once and caches, but the cache key includes the function identity, so any structural change to `args` forces recompilation that should not be necessary.

### Where to do the setup

`_Newton.__init__` doesn't know the param structure — `unravel` needs `init_params`. The protocol's first method that does see `init_params` is `init_state`, which is always called before any `update` or `run` (the existing `_hess_tag` default already piggybacks there). That's the natural hook. Guard with `if self._unravel is None` so it's idempotent.

```python
class _Newton:
    def __init__(self, fun, fun_with_aux, tol=1e-6):
        self.fun = fun
        self.fun_with_aux = fun_with_aux
        self.tol = tol
        self._hess_tag = None
        self._hess_fn = None
        self._line_search = optax.scale_by_backtracking_linesearch(max_backtracking_steps=30)
        # populated on first init_state call
        self._unravel = None
        self._fun_flat = None
        self._vag_flat_with_aux = None
        self._hessian_flat = None

    def _build_cache(self, init_params):
        _, self._unravel = ravel_pytree_nest(init_params)
        self._fun_flat          = lambda x, *a: self.fun(self._unravel(x), *a)
        self._fun_flat_with_aux = lambda x, *a: self.fun_with_aux(self._unravel(x), *a)
        self._vag_flat_with_aux = jax.value_and_grad(self._fun_flat_with_aux, has_aux=True)
        self._hessian_flat      = jax.hessian(self._fun_flat)

    def init_state(self, init_params, *args):
        if self._unravel is None:
            self._build_cache(init_params)
        if self._hess_tag is None:
            self._hess_tag = HessianTag(structure=Full, property=General)
        params_flat, _ = ravel_pytree_nest(init_params)
        return NewtonState(
            grad_norm=jnp.inf,
            stats=OptimizationInfo(...),
            ls_state=self._line_search.init(params_flat),
        )

    def update(self, params, state, *args, maxiter, force_autodiff_hessian):
        params_flat, _ = ravel_pytree_nest(params)
        (fval, aux), g_flat = self._vag_flat_with_aux(params_flat, *args)
        ...
        H = (
            self._hess_fn(params, *args)
            if (self._hess_fn is not None and not force_autodiff_hessian)
            else self._hessian_flat(params_flat, *args)
        )
        ...
        return new_params, new_state, aux
```

Two things this fix does:

- **Stops rebuilding `f` every call.** `self._fun_flat` is a stable bound-method closure over `self._unravel`. `args` is now a runtime argument, not closure-captured, so the function identity (and the jit cache key) stays constant across `update` calls.
- **Collapses three forward passes into one.** `_vag_flat_with_aux` returns `(value, aux), grad` in a single fwd+bwd. The outer `Newton.update` loses its post-step `fun_with_aux(params, *args)` (`_newton.py:294-303`) — `_Newton.update` already returns `aux`. Same change applies to `Newton.run` (line 314); once-per-fit so lower priority.

### The autodiff Hessian path needs the same treatment, not a free ride

`jax.hessian(self._fun_flat)` is an independent compiled trace from `jax.value_and_grad(self._fun_flat_with_aux)` — caching the gradient transform does not automatically improve the autodiff Hessian path. Both transformations want the same underlying fix (a stable `_fun_flat` reference) but each must be cached as its own attribute. The sketch above caches `self._hessian_flat` alongside `self._vag_flat_with_aux` for exactly this reason; they're parallel fixes that land in the same PR.


## 2. Pull the run loop into a `solvers/loop` abstraction

`solvers/_newton.py:195-241`:

```python
if jit:
    def cond(carry): ...
    def body(carry): ...
    params, state = eqx.internal.while_loop(cond, body, (params, state), kind="lax")
else:
    for _ in range(maxiter):
        params, state = self.update(...)
        if state.stats.converged:
            break
```

Two loop bodies for the same logic. The convergence/maxiter check is duplicated; future early-stopping solvers (LBFGS adapter, SVRG) will duplicate it again.

Suggest a `solvers/while_loop.py` exposing a single signature:

```python
def while_loop(cond_fn, body_fn, init_carry, *, jit: bool):
    ...
```

Backed by `eqx.internal.while_loop` for jit (with `kind="lax"`; `"bounded"` / `"checkpointed"` available if we ever need backward through the loop) and a thin Python `while` otherwise. Equinox does not ship a pure-Python form with this signature — we add it. `_Newton.run` collapses to a single `while_loop(...)` call.

## 3. Replace in-method closure composition with methods + the optax line-search piece

`solvers/_newton.py:112-193` — `update` mixes the Newton direction, the line search, the `descent` accept/reject branch, and the state-update accounting in one body of nested closures (`step`, `accept`, `reject`, `no_step`).

Refactor:

- Newton direction as a method: `_newton_direction(self, g, H, tag) -> step`.
- Accept/reject as a method: `_apply_or_reject(self, params, step, grad, state, *args) -> (params, ls_state)`.
- Line search stays as `self._line_search`.
- `update` becomes a flat assembly point with no per-call closures.

Lines up with point 1 (no closures rebuilt per call) and makes variants like damped Newton or trust-region a one-method override.

## Smaller cleanups

- **`Newton.__init__` blows up opaquely if `init_params=None`.** `regularizer.penalized_loss(..., params=init_params, ...)` reaches `_get_filter_kwargs`, which for `GroupLasso` walks the param shape. Guard with a clear error or document that `init_params` is mandatory.

BalzaniEdoardo:
We are close to be done here! 

I added a few tests for the correctness of the hessian, it failed only for gaussian, we were missing a 0.5 factor, I fixed that.

A few comments, no great solutions:

## Batch-dim contract

Right now we assume the Hessian batch dimension is the last one (this is what `ravel_pytree_nest` infers from the trailing shape of the leaves). That heuristic can misfire when unrelated dimensions coincide — `n_features == n_neurons`, or a single-feature/single-class model where `coef` and `intercept` both end in a `1`. It would be safer to make the contract explicit: have the parameter class carry an attribute like `batched_dims` (possibly more than one) that fully specifies the batching, instead of guessing it from shapes.

## Vmapped Hessian for the analytic path

If we know the batched dims, we can `vmap` the Hessian so we never expand to the full, gigantic `(D, D)` matrix — not even on the analytic path:

```python
hess_fun = jax.vmap(jax.hessian(per_block_loss), in_axes=batched_dims)
blocks = hess_fun(params, X, y)   # (B..., d, d), the full (D, D) is never formed
```

Today the autodiff path materializes the full dense `(D, D)` and `PopulationGLM._get_hess_fn` builds its blocks with a Python `for` loop over neurons; with `batched_dims` declared, both collapse to the same `vmap` and feed the existing block-diagonal solve, scaling linearly in the number of blocks.

## Tree-native solve without ravel/unravel

For the analytic path (maybe not worth trying): we could avoid raveling/unraveling the trees altogether and solve in tree space with `lineax.PyTreeLinearOperator`. The Hessian from `jax.hessian` is already a pytree-of-pytrees mapping params to params, so it can be handed to the operator directly:

```python
H = jax.hessian(value_fn)(params)                  # GLMParams-of-GLMParams
neg_g = jax.tree_util.tree_map(lambda x: -x, g_tree)
operator = lx.PyTreeLinearOperator(
    H,
    output_structure=jax.eval_shape(lambda: g_tree),  # = gradient/param structure
    tags=lx.positive_semidefinite_tag,                # NB: tag is the 3rd arg, not output_structure
)
step_tree = lx.linear_solve(operator, neg_g, lx.Cholesky()).value  # step is a GLMParams tree
```

Two gotchas: the structural tag goes in `tags=` (passing it positionally lands it in `output_structure` and lineax errors), and the gradient must stay a tree — flattening `g` defeats the point.

BalzaniEdoardo:
I opened an issue for this vmap proposal, I think it should be on a subsequent pr. Let me know if i can fully review this one last time (test completed etc).

src/nemos/base_regressor.py
```diff
@@ -407,6 +434,14 @@ def _instantiate_solver(
             **solver_kwargs,
         )
 
+        if isinstance(solver, NewtonSolverProtocol):
+            solver.setup_hessian(
+                self._get_hess_fn(init_params, autodiff=solver.autodiff),
```
BalzaniEdoardo, line 441: add autofiff to the solver protocol, otherwise this attribute may be not defined

src/nemos/solvers/_newton.py
```diff
@@ -0,0 +1,368 @@
...
+            while cond(carry):
+                carry = body(carry)
+            final_params, final_state = carry
+
+        _, aux = self.fun_with_aux(final_params, *args)
+        return final_params, final_state, aux
+
+
+class Newton(NewtonSolverProtocol[NewtonState]):
+    def __init__(
```
BalzaniEdoardo: the init signature is not compatible with the protocol

src/nemos/base_regressor.py
```diff
@@ -335,6 +352,16 @@ def _check_solver_kwargs(solver_class: Type, solver_kwargs: dict[str, Any]) -> N
                 f"kwargs {undefined_kwargs} in solver_kwargs not a kwarg for {solver_class.__name__}!"
             )
 
+    def _get_hess_fn(self, params, solver, autodiff: bool) -> Callable | None:
```
BalzaniEdoardo: this signature has a `solver` positional argument but the glm override differs. we should resolver this

src/nemos/glm/glm.py
```diff
@@ -50,6 +58,47 @@
...
+            ncols = x_leaf.shape[1]
+            lam_leaf = jnp.asarray(lam_leaf)
+
+            if lam_leaf.ndim == 0:
+                return jnp.full(ncols, lam_leaf)
+
+            return jnp.ravel(lam_leaf)
+
+        lam = jnp.concatenate(
+            jax.tree_util.tree_leaves(jax.tree_util.tree_map(expand_lam, X, lam.coef))
```
BalzaniEdoardo, line 95: remove the default lam=0.0, here lam has to be a glmparam, otherwise glm.coef will fail

src/nemos/base_regressor.py
```diff
@@ -335,6 +352,16 @@ def _check_solver_kwargs(solver_class: Type, solver_kwargs: dict[str, Any]) -> N
...
+        return None
+
+    def _invalidate_solver(self):
```
BalzaniEdoardo, line 360: shouldn't we invalidate the sovler also when the setter of solver_name is called? if so, add a test that checks that this happen

src/nemos/solvers/_hess.py
```diff
@@ -0,0 +1,133 @@
...
+) -> HessianTag | None:
+    """Combine tags assuming additivity.
+
+    Valid when the total objective is a sum of two functions (e.g. loss + regularizer),
+    since the Hessian of a sum is the sum of the Hessians.
+    """
+    if t1 is None or t2 is None:
+        return None
+    prop = combine_property(t1.property, t2.property)
+    if prop is None:
```
BalzaniEdoardo: this is a dead branch, none is never returned

tests/test_newton.py
```diff
@@ -0,0 +1,728 @@
...
+
+    # Test not UnRegularized
+    assert not np.allclose(H, regularized_hess)
+    # Test Ridge
+    diag = jnp.stack([jnp.diag(jnp.array(d).ravel()) for d in diag], axis=0)
+    H = H.at[:, :-3, :-3].add(diag)
+    np.testing.assert_allclose(desired=H, actual=regularized_hess)
+
+
+@pytest.mark.parametrize("regularizer ", ["UnRegularized", "Ridge"])
```
BalzaniEdoardo:
```suggestion
@pytest.mark.parametrize("regularizer", ["UnRegularized", "Ridge"])
```

src/nemos/regularizer.py
```diff
@@ -178,6 +180,32 @@ class Regularizer(Base, abc.ABC):
...
+        if self._hess_tag is None:
+            return None
+        if hasattr(params, "regularizable_subtrees"):
+            regularized = {
+                id(leaf)
+                for where in params.regularizable_subtrees()
+                for leaf in jax.tree_util.tree_leaves(where(params))
+            }
+            all_regularized = regularized == {
+                id(leaf) for leaf in jax.tree_util.tree_leaves(params)
```
BalzaniEdoardo, line 202: since we are using the id, we should make sure that `regularizable_subtrees` does not copy the params. I think the most robust test would be checking all ModelParams subclasses and checks. I'll add it.

src/nemos/glm/glm.py
```diff
@@ -1252,6 +1421,36 @@ def save_params(self, filename: Union[str, Path]):
         self._save_params(filename, fit_attrs, string_attrs)
 
 
+def _var_func_of_mu(model) -> Callable:
```
BalzaniEdoardo, line 1424: This function that uses func name to pull the correct var function is a bit brittle, if a user, for some weird reason subclass poisson obs and adds some methods, all of a sudden it would not be able to use newton. 
For now it is ok, but i think we should just use autodiff to get the hessian the more I think about it.

BalzaniEdoardo (approved):
This generally looks good, except a few things I spotted here and there.

My only high level doubt is how general is the solver and how many choices are somewhat hard-coded. For example, currently 

```
jax.vmap(solve)(H, g_flat.reshape(H.shape[0], -1))
```

Have an hard-coded assumption on `g_flat`. It would be neat if we could do everything in pytree terms, using `lineax.PytreeLinearOperator` preventing all raveling and un-raveling. 

In that case we would need to structure the hessian the same way as autodiff would, and at this point we should directly use autodiff instead of constructing the hessian ad hoc. The tagging mechanism would still be useful to detect the structure and property of the problem.

I think this PR is good as is, but we could follow up with more polishing and simplification.

src/nemos/glm/glm.py
wulfdewolf: Agreed! That's for the next PR then, eager to delete so many lines

src/nemos/regularizer.py
wulfdewolf: So this is a different PR as well then?

BalzaniEdoardo (approved):
LGTM, thans @wulfdewolf that was a lot of work

PR #564, opened 2026-05-11
Compute optimal SVRG params if inv link is given as "softplus"

bagibence:
Step size and batch size can be optimized for GLMs fitted with SVRG if the observation model is Poisson and the inverse link is softplus.

This was only done if the inverse link was specified as `jax.nn.softplus`. If the user specified the same with `"softplus"`, that got resolved to the wrapped `inverse_link_function_utils.softplus` and this calculation was skipped.

This fix just adds that function to the if statement.


BalzaniEdoardo (approved):
LGTM

PR #566, opened 2026-05-19
Add GLMHMM class init

BalzaniEdoardo:
## Summary

Adds a `GLMHMM` class that plugs into the post-#557 `BaseHMM` initialization API. The class is **scaffolding**: constructor, attribute validation setters, `setup()`, and the model-side initialization plumbing are implemented and tested; `fit`, `score`, `simulate`, `decode_state`, `smooth_proba`, `filter_proba`, `update`, `_log_likelihood`, `save_params`, and the optimization-init hooks are intentionally `pass` and will be filled in subsequent PRs.

Secondary changes hoist a small introspection helper, fix a stale parameter name flagged by `scripts/check_parameter_naming.py`, and remove the abstract `predict` hook from `BaseRegressor`.

## Motivation

`BaseHMM` (merged via #557) splits initialization into two pipelines:

| pipeline   | property                       | setup hook         | merger function                  |
| ---------- | ------------------------------ | ------------------ | -------------------------------- |
| HMM-side   | `hmm_initialization_funcs`     | `_hmm_setup`       | `setup_hmm_initialization`       |
| model-side | `model_initialization_funcs`   | `_model_setup`     | (model-specific)                 |

`GLMHMM` is the first concrete model to consume this contract. It needs:

- a model-side merger (`setup_glm_hmm_initialization`) wired into `_model_setup`,
- a model-specific default dict (`DEFAULT_INIT_FUNCTIONS_GLMHMM`),
- a GLM-aware KMeans initializer subclass (`KMeansInitializerGLM`),
- correct sklearn-compatible introspection so `get_params` / `set_params` round-trip through the property setters.

### New: `src/nemos/glm_hmm/glm_hmm.py`

`GLMHMM(BaseHMM[GLMHMMUserParams, GLMHMMParams, GLMHMM_INITIALIZATION_FN_DICT])`:

- Class attributes: `_validator_class = GLMHMMValidator`, `_model_default_init_dict = DEFAULT_INIT_FUNCTIONS_GLMHMM`, `_kmeans_init_class = KMeansInitializerGLM`.
- Constructor exposes `hmm_initialization_funcs` and `model_initialization_funcs` (same name as the inherited property, so sklearn `get_params` round-trips).
- `observation_model` and `inverse_link_function` setters validate / resolve string shortcuts and store the resolved instance.
- `setup()` accepts the HMM-side keys and the GLM-side keys (`glm_params_init`, `scale_init`) and forwards them; HMM-side keys go to `_hmm_setup`, model-side keys go to `_model_setup` via `**kwargs`.
- `_model_setup` invokes `setup_glm_hmm_initialization` and then derives `_model_use_kmeans` from **function identity** against `kmeans_glm_params_init` / `kmeans_scale_init`. This means the flag is correct whether the user passes the string `"kmeans"`, the kmeans callable directly via `model_initialization_funcs={...}`, or the property setter — fixing a silent disable that string-equality detection would otherwise cause.

### Public API: `Literal` annotations on `setup()`

Both `BaseHMM.setup` / `BaseHMM._hmm_setup` and `GLMHMM.setup` now type each string-or-callable parameter as `Optional[Literal[...] | Callable]`, enumerating every built-in string alias. New tests in both test files use a shared `extract_literal_options` helper to assert the literals match the corresponding registry; any future addition or rename of a built-in initializer that misses the signature will fail CI loudly.

### Hoisted: `nemos._inspect_utils.extract_literal_options`

Generic introspection helper that walks `Optional[Literal[...] | Callable]` annotation trees and returns the literal set. Lives in `_inspect_utils` next to the other inspect helpers; consumed by both `tests/test_hmm_base_class.py` and `tests/test_glm_hmm_class_init.py`.

### Tests

`tests/test_glm_hmm_class_init.py` (new, ~395 lines) covers:

- inherited setter routing (parametrized over `n_states`, `maxiter`, `tol`, `seed`, Dirichlet priors),
- `observation_model` / `inverse_link_function` setters (strings, instances, invalid input),
- both init-funcs pipelines: defaults, custom callable threading through to the respective merger,
- `_model_use_kmeans` flag for the three entry points (default off, set via `"kmeans"` string, set via kmeans callable in `model_initialization_funcs`),
- `setup()` arg routing per pipeline, partial-args, and cross-key accumulation (after configuring one key, configuring a different key in the same pipeline preserves the first via threaded `init_funcs`),
- `Literal` annotation drift check against the registry.

`tests/test_hmm_base_class.py` gains the matching `Literal` drift check.

### Adjacent fixes

| change                                                          | why                                                                                                                                |
| --------------------------------------------------------------- | ---------------------------------------------------------------------------------------------------------------------------------- |
| `BaseRegressor.predict` removed                                 | `GLMHMM` exposes `score` / `simulate` / `decode_state` etc., but not a single `predict` — see open question 2.                     |
| `src/nemos/io/io.py` arg `parameter_name` → `param_name`        | `scripts/check_parameter_naming.py` flagged a clash with the new `extract_literal_options` arg; `param_name` matches the codebase. |
| `scripts/check_parameter_naming.py` `VALID_PAIRS` updated       | declare `{"hmm_initialization_funcs", "model_initialization_funcs"}` as a legitimate co-occurrence in `GLMHMM.__init__`.            |


src/nemos/glm_hmm/glm_hmm.py
```diff
@@ -0,0 +1,609 @@
...
+    KMeansInitializerGLM,
+    kmeans_glm_params_init,
+    kmeans_scale_init,
+    setup_glm_hmm_initialization,
+)
+from .params import GLMHMMParams, GLMHMMUserParams
+from .validation import GLMHMMValidator
+
+
+def _check_state_format(state_format: str) -> None:
```
sjvenditto: I previously moved this function to hmm/utils.py since it is being used by the BaseHMM class. We can probably remove it here

src/nemos/glm_hmm/glm_hmm.py
```diff
@@ -0,0 +1,609 @@
...
+        dirichlet_initial_proba: Union[jnp.ndarray, None] = None,  # (n_state, )
+        dirichlet_transition_proba: Union[
+            jnp.ndarray | None
+        ] = None,  # (n_state, n_state)
+        solver_name: str = None,
+        solver_kwargs: Optional[dict] = None,
+        maxiter: int = 1000,
+        tol: float = 1e-8,
+        seed=jax.random.PRNGKey(123),
+        hmm_initialization_funcs: Optional[INITIALIZATION_FN_DICT] = None,
```
sjvenditto: should we rename this dictionary type to `HMM_INITIALIZATION_FN_DICT` to match other conventions?

src/nemos/hmm/hmm.py
```diff
@@ -133,17 +133,21 @@ def __init__(
         self.transition_prob_: Optional[jnp.ndarray] = None
         self.initial_prob_: Optional[jnp.ndarray] = None
 
-        self.hmm_initialization_funcs = hmm_initialization_funcs
-
         # flags for kmeans initialization
         self._hmm_use_kmeans = None
```
sjvenditto: I think we can remove these entirely since they're getting set in  `_hmm_setup` and `_model_setup` during the dictionary setters

src/nemos/glm_hmm/glm_hmm.py
```diff
@@ -0,0 +1,609 @@
...
+                p for p, missing in zip(param_labels, is_missing) if missing
+            ]
+            raise ValueError(
+                "This GLMHMM instance is not fitted yet. The following attributes are not set:"
+                f" {missing_params}.\nPlease fit the GLM-HMM model first or "
+                "set the missing attributes."
+            )
+
+    def _kmeans_extra_kwargs(self) -> dict:
+        return {"inverse_link_function": self.inverse_link_function}
```
sjvenditto:
```suggestion
        return {"inverse_link_function": self.inverse_link_function, "observation_model": self.observation_model}
```

observation model needs to be added here for kmeans

src/nemos/hmm/hmm.py
sjvenditto: What is the logic behind including the initialization dictionary type like this? Is it to be clear to child classes that a new initialization dictionary must be made?

sjvenditto (approved):
I have some minor changes to address, otherwise it's looking great!

I only had one question about why the dictionary typing was being added in the class header -- it's not that I disagree with including it there, I was more curious about why it was added in

sjvenditto:
Some last changes that I could not put at specific lines since they were in non-edited files:
- From the `GLMHMMValidator`, I think we can remove the validation step `"check_array_dimensions"` since it's redundant with the shape checks. We can also remove, then, the `expected_param_dims` attribute
- For the `GLMHMM_INITIALIZATION_FN_DICT` type definition in `glm_hmm/initialize_parameters.py`, we can remove `InitFunctionHMM` from the type hint since it no longer contains the hmm parameters (I forgot to do this when making the update before)

src/nemos/hmm/hmm.py
BalzaniEdoardo: that works but the static checker complains that the attribute is set outside the init, I think we can ignore the checker

src/nemos/hmm/hmm.py
BalzaniEdoardo: when the dictionary of the glm hmm contained both hmm and glm init funcs, that was a nice way to state that model object is defined by the parameter structure and the initializaitons.  I don't think it makes as much sense now

BalzaniEdoardo: My solution is now use the model specific initialization param dict as the generic type, this makes sense because any new model that inherits form basehmm will need specific initialization functions

PR #567, opened 2026-05-26
Complete GLMHMM class

BalzaniEdoardo:
## Summary
- Introduces the `GLMHMM` public class with a complete fit/update/inference API: `fit`, `update`, `simulate`, `smooth_proba`, `filter_proba`, `decode_state`, and `save_params`. The class builds on the existing `BaseHMM`.
- Extends `BaseHMM` with `session_starts` propagation through the initialization path, `_kmeans_kwargs_equal` for correct deduplication of kmeans initializers across slots, and `_kmeans_init_func_kwargs` for passing model-specific kwargs to kmeans-based init functions.
- Adds `solver_param_subtree` to `GLMHMMParams` so the numerical solver and GroupLasso regularizer operate on the GLM-level sub-pytree (`model_params`) rather than the full HMM parameter tree.
- Adds `_resolve_existing_slots` in `hmm/initialize_parameters.py` to support the IO round-trip: `save_params` serialises built-in callables as strings; `load_model` resolves them back and sets `*_custom = False` before the setup path runs.
- Registers `GLMHMM` in the IO layer and improves `mapping_dict` handling for nested `*_initialization_funcs` dicts so that per-slot overrides flow correctly through `load_model`.
- Drops the pseudo-r2 `score` method: there is no saturated model for an HMM, so a deviance-based score is not meaningful.
- Removes the stale `_validate` method that was a leftover from an old `BaseRegressor` API.
- Comprehensive, self-contained docstrings for all public methods — no method defers to "see base class for full documentation".
- Removes implementation-detail prose from the `save_params` docstring; replaces it with user-facing guidance on recovering custom initializers via `mapping_dict`.


src/nemos/glm_hmm/glm_hmm.py
```diff
@@ -270,19 +423,23 @@ def setup(
         - ``glm_params_init``: ``"random"`` (default), ``"kmeans"``.
         - ``scale_init``: ``"constant"`` (default), ``"kmeans"``.
 
-        Custom callables must follow one of two protocols depending on the parameter
-        they initialize:
+        Custom callables must satisfy one of two ``typing.Protocol`` classes:
```
sjvenditto: I wonder if it's useful to move these details to a separate section (something like "Notes") down below so that the details aren't too overwhelming for users that just want basic use cases

src/nemos/glm_hmm/glm_hmm.py
```diff
@@ -435,92 +644,813 @@ def _kmeans_extra_kwargs(self) -> dict:
...
+
+        self._validator.feature_mask_consistency(
+            getattr(self, "_feature_mask", None), init_params
+        )
+
+        # filter for non-nans, grab data if needed
+        data, y, session_starts = self._preprocess_inputs(X, y, session_starts)
+
+        # make sure session_starts starts with a 1
+        session_starts = session_starts.at[0].set(True)
```
sjvenditto: This should already be enforced in `validate_and_cast_session_starts`. Instead of this redundancy, we can update the tests for initializing sessions starts to also ensure that the first value is 1 after  `_preprocess_inputs` is called (see suggestions in  `test_hmm_base_class.py`)

tests/test_hmm_base_class.py
```diff
@@ -935,16 +939,16 @@ class TestHMMNewSession:
...
+    def test_initialize_new_session(self, X, y, session_starts, expected_new_session):
         """Test that session boundaries are correctly initialized."""
         model = MockHMM(n_states=3)
-        is_new_session = model._validator.validate_and_cast_is_new_session(
-            X, y, is_new_session
+        session_starts = model._validator.validate_and_cast_session_starts(
+            X, y, session_starts
         )
-        assert jnp.all(is_new_session == expected_new_session)
+        assert jnp.all(session_starts == expected_new_session)
```
sjvenditto, line 789:
```suggestion
        assert jnp.all(session_starts == expected_new_session)
        _, _, session_starts = model._preprocess_inputs(X, y, session_starts)
        assert sessions_starts[0] == 1
```

tests/test_hmm_base_class.py
```diff
@@ -1009,19 +1013,19 @@ def test_initialize_new_session(self, X, y, is_new_session, expected_new_session
...
     ):
         """Test that session boundaries are correctly moved when there are NaN values."""
         model = MockHMM(n_states=3)
-        is_new_session = model._validator.validate_and_cast_is_new_session(
-            X, y, is_new_session
+        session_starts = model._validator.validate_and_cast_session_starts(
+            X, y, session_starts
         )
-        assert jnp.all(is_new_session == expected_new_session)
+        assert jnp.all(session_starts == expected_new_session)
```
sjvenditto, line 866:
```suggestion
        assert jnp.all(session_starts == expected_new_session)
        _, _, session_starts = model._preprocess_inputs(X, y, session_starts)
        assert sessions_starts[0] == 1
```

src/nemos/glm_hmm/glm_hmm.py
```diff
@@ -435,92 +644,813 @@ def _kmeans_extra_kwargs(self) -> dict:
...
+        """
+        # Convert a pytree to a design-matrix with pytrees
+        X = jnp.hstack(jax.tree_util.tree_leaves(X))
+
+        if n_samples is None:
+            n_samples = X.shape[0]
+        else:
+            if not isinstance(n_samples, int):
+                raise TypeError(
+                    "`n_samples` must either `None` or of type `int`. Type {type(n_sample)} provided "
```
sjvenditto:
```suggestion
                    f"`n_samples` must either `None` or of type `int`. Type {type(n_sample)} provided "
```

src/nemos/glm_hmm/glm_hmm.py
```diff
@@ -435,92 +644,813 @@ def _kmeans_extra_kwargs(self) -> dict:
...
+            self._validate_and_prepare_inputs(feedforward_input, None, session_starts)
+        )
+
+        # preprocess inputs (drop nans, extract data)
+        data, _, session_starts = self._preprocess_inputs(
+            feedforward_input, None, session_starts
+        )
+
+        # ensure first time point is a session start
+        session_starts = session_starts.at[0].set(True)
```
sjvenditto: also redundant

src/nemos/glm_hmm/glm_hmm.py
```diff
@@ -435,92 +644,813 @@ def _kmeans_extra_kwargs(self) -> dict:
...
+        >>> states_onehot.shape
+        (100, 3)
+        >>> bool(np.all(states_onehot.sum(axis=1) == 1))
+        True
+        """
+        return super().decode_state(
+            X, y, session_starts=session_starts, state_format=state_format
+        )
+
+    def _simulate(
```
sjvenditto: very minor, but can you move this function to come before `simulate` up above?

src/nemos/glm_hmm/glm_hmm.py
```diff
@@ -550,42 +1480,158 @@ def _set_model_params(self, params: GLMHMMParams):
...
+        self._validator.validate_inputs(X=X, y=y)
+        session_starts = self._validator.validate_and_cast_session_starts(
+            X, y, session_starts=session_starts
+        )
+
+        # drop nans and pull pytree data
+        data, y, session_starts = self._preprocess_inputs(X, y, session_starts)
+
+        # ensure first sample is a session start
+        session_starts = session_starts.at[0].set(True)
```
sjvenditto: also redundant

src/nemos/hmm/hmm.py
```diff
@@ -496,47 +538,82 @@ def _kmeans_resolve_model_kwargs(self, use_kmeans, init_funcs) -> dict:
...
         if self._model_use_kmeans is not None:
             kmeans_kwargs = self._kmeans_resolve_model_kwargs(
                 self._model_use_kmeans, self._model_initialization_funcs
             )
         else:
             kmeans_kwargs = {}
 
-        # setup initializer
+        model_kwargs = self._kmeans_extra_kwargs()
+        kmeans_kwargs = self._kmeans_reconcile_user_kwargs(kmeans_kwargs, model_kwargs)
```
sjvenditto: Instead of this extra checking of whether or not a user provided observaiton model matches the model, isn't it easier to just add `observation_model` in the `InitFunctionGLM` protocol? That will prevent a user from setting it, and thus bypassing the need of having `_kmeans_reconcile_user_kwargs` and `_kmeans_init_func_kwargs`. 

Also, the difference between `_kmeans_extra_kwargs` and `_kmeans_init_func_kwargs` isn't immediately obvious, especially given that there is overlapping but different sets for the GLM-HMM. I worry that this will be a point of confusion in future HMM subclasses

src/nemos/glm_hmm/validation.py
sjvenditto, line 72: This validation step is redundant with shape checks later on

sjvenditto, line 50: we can remove this attribute if we remove the "check_array_dimensions" validation step

tests/test_glm_hmm_validator.py
```diff
@@ -0,0 +1,591 @@
...
+            validator.check_init_and_transition_prob_sum_to_1(params)
+
+
+# ---------------------------------------------------------------------------
+# validate_inputs
+# Moved from test_glm_hmm_class.py::TestGLMHMM::test_nan_between_epochs
+# ---------------------------------------------------------------------------
+
+
+class TestValidateInputs:
```
sjvenditto: a couple of these are redundant with tests in `test_hmm_base_class.py::TestHMMValidator`

tests/test_glm_hmm_validator.py
```diff
@@ -0,0 +1,591 @@
...
+        with pytest.raises(TypeError):
+            validator.validate_and_cast_params(params)
+
+
+# ---------------------------------------------------------------------------
+# check_init_and_transition_prob_shape
+# ---------------------------------------------------------------------------
+
+
+class TestCheckInitAndTransitionProbShape:
```
sjvenditto, line 125: There are a number of tests here in `TestCheckInitAndTransitionProbShape`, `TestCheckInitAndTransitionProbSumTo1`, `TestValidateInputs`, `TestValidateAndCastIsNewSession` that are not specific to the GLM-HMM, or are overlapping with tests in `test_hmm_base_class.py` (e.g. in `TestHMMValidator` and `TestHMMNewSession`). The tests are fast so it might not be worth it, but if we want to reduce overlap we can remove redundant tests and maybe move all of these to a `test_hmm_validator.py` script to acknowledge that it's testing thing specific to the `HMMValidator`

sjvenditto (changes requested):
Looking great! I've made a few suggestions throughout the code. Also as a more high-level comment: In a lot of the docstrings we describe things related to `y` using terms like `neurons`, `n_neurons`, `neural_activity`, etc. I know this terminology is inherited from how things are described in the GLM docstrings, but I wonder if we want to move to more "general" language? Or even a combination of both, e.g. saying something like "neuron or observed variable". If you think this is worth doing, I can help comb through the docstrings to make these updates.

src/nemos/glm_hmm/validation.py
sjvenditto: nevermind, it is not completely redundant

sjvenditto: ignore

sjvenditto (approved):

PR #568, opened 2026-05-27
Add efgp basis functions

pgree:
### Adds equispaced Fourier Gaussian process basis functions, only with squared-exponential kernel in 1D. 

The main addition is a function that chooses frequencies and weights (multiplicative scalings) of sin/cos 
basis functions. Weights and frequencies are chosen so that the Gaussian proccess defined by 
putting N(0, 1) priors on coefficients in the expansion is an approximate squared-exponential Gaussian 
process. 

- The code is primarily in `src/nemos/basis/_fourier_basis.py` 
- There are also test functions in `tests/test_efgp.py`
- There's a simple example of usage in `scripts/demo_fourier_se_basis.py`, though not sure this should be added
- Also, a more complete tutorial at `scripts/efgp_tutorial.ipynb`


BalzaniEdoardo (commented):
Sorry it took me so much time to get back to this. I think we do not need an extra class for this, we can basically re-use the old class with minor additions. 


Below an example subclass, that shows this may look. I worked on the 1d case only so far, just to show how this may look.

I added a class method that takes care of defining the weights and bounds, so that the object is actually configured in the right way, and I added a little extra computation in the `evaluate` method. 

```python

import jax

jax.config.update("jax_enable_x64", True)  # match the GP accuracy regime

import numpy as np

from nemos.basis import FourierEval, FourierSEBasis
from nemos.basis._fourier_basis import _se_quadrature


class WeightedFourierEval(FourierEval):
    """FourierEval with an optional per-frequency column weight.

    ``weights[j]`` scales BOTH the cosine and sine column of frequency ``j``
    (one spectral amplitude per frequency, shared by its cos/sin pair). This
    is exactly what the GP construction needs. ``weights=None`` -> all ones,
    i.e. identical to a plain FourierEval.
    """

    def __init__(self, *args, weights=None, **kwargs):
        # FourierEval defaults label="FourierEval", which the label validator
        # rejects on a subclass; fall back to this class's name instead.
        kwargs.setdefault("label", None)
        super().__init__(*args, **kwargs)
        self._weights = None if weights is None else np.asarray(weights, dtype=float)
        if self._weights is not None:
            n_freq = self._freq_combinations.shape[-1]
            if self._weights.shape != (n_freq,):
                raise ValueError(
                    f"``weights`` must have one entry per frequency ({n_freq},); "
                    f"got {self._weights.shape}."
                )

    def evaluate(self, *sample_pts):
        out = super().evaluate(*sample_pts)
        if self._weights is None:
            return out
        # Layout from FourierBasis.evaluate is [cos block | sin block], where
        # the cos block has one column per frequency and the sin block drops
        # the DC term. Build the matching per-column weight vector and scale.
        has_dc = self._has_zero_phase
        col_weights = np.concatenate([self._weights, self._weights[has_dc:]])
        return out * col_weights

    @classmethod
    def squared_exponential(cls, lengthscale, domain, eps, variance=1.0):
        """Build the SE-GP Fourier basis from kernel hyperparameters."""
        t0, t1 = float(domain[0]), float(domain[1])
        L = t1 - t0
        xcen = 0.5 * (t0 + t1)
        _, weights, h, m = _se_quadrature(lengthscale, variance, eps, L)
        # _se_quadrature returns [cos weights | sin weights]; the first m+1 are
        # the per-frequency amplitudes [w_0, sqrt2*w_1, ..., sqrt2*w_m].
        per_freq = np.asarray(weights[: m + 1])
        # bounds encode the physical frequency spacing: rescaling x to
        # (x - xcen) / (1/h) = h*(x - xcen) turns integer frequencies j into
        # the physical grid xi_j = j*h.
        return cls(
            frequencies=np.arange(m + 1),
            frequency_mask="all",
            bounds=(xcen, xcen + 1.0 / h),
            weights=per_freq,
        )



if __name__ == "__main__":
    # A few hyperparameter settings to compare against.
    cases = [
        dict(lengthscale=0.1, domain=(0.0, 1.0), eps=1e-4, variance=1.0),
        dict(lengthscale=0.5, domain=(-2.0, 3.0), eps=1e-6, variance=2.0),
        dict(lengthscale=0.05, domain=(0.0, 1.0), eps=1e-2, variance=0.5),
    ]

    for kw in cases:
        ref = FourierSEBasis(**kw)
        new = WeightedFourierEval.squared_exponential(**kw)

        x = np.linspace(kw["domain"][0], kw["domain"][1], 50)
        Phi_ref = np.asarray(ref.evaluate(x))
        Phi_new = np.asarray(new.evaluate(x))

        same_shape = Phi_ref.shape == Phi_new.shape
        max_abs_err = np.max(np.abs(Phi_ref - Phi_new)) if same_shape else np.inf
        ok = same_shape and np.allclose(Phi_ref, Phi_new, atol=1e-10, rtol=0)

        print(
            f"l={kw['lengthscale']:<5} var={kw['variance']:<4} eps={kw['eps']:<6} "
            f"domain={kw['domain']}  ->  n_cols ref/new = "
            f"{Phi_ref.shape[1]}/{Phi_new.shape[1] if same_shape else '?'}  "
            f"max|Δ|={max_abs_err:.2e}  {'OK' if ok else 'MISMATCH'}"
        )
        assert ok, "subclass does not reproduce FourierSEBasis"

    # Sanity check: weights=None reproduces a plain FourierEval exactly.
    plain = FourierEval(frequencies=np.arange(6), frequency_mask="all")
    wrapped = WeightedFourierEval(frequencies=np.arange(6), frequency_mask="all")
    x = np.linspace(0.0, 1.0, 17)
    assert np.allclose(plain.evaluate(x), wrapped.evaluate(x))
    print("\ndefault (weights=None) matches plain FourierEval: OK")

```

BalzaniEdoardo:
Closed to avoid confusion, see #581 for the correct pr

PR #570, opened 2026-05-29
Remove public `compute_loss`

sjvenditto:
Instead of making a custom override of `compute_loss` for HMM models, we're opting to remove the public method `compute_loss` completely. Any advanced need for access to this can use the private `_compute_loss` instead.


BalzaniEdoardo:
@sjvenditto are you still on this one?

PR #576, opened 2026-06-02
fix sklean compat

BalzaniEdoardo:
After release of scikit-learn `v1.1.9`, the internal machinery of sklearn sets a new attribute as part of a pipleline.

```python
sub_estimator._parent_callback_ctx = self
```
Our transformer basis only allows a selected number of attributes to be set. This new attribute needs to be added to the settable once.

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -18,9 +18,12 @@ build:
       # Skip files referencing to sklearn docs, which uses Text fragment URLs (#:~:text=)
       # that are not recognized as valid anchors by htmlproofer yet, see issue below
       # https://github.com/gjtorikian/html-proofer/issues/849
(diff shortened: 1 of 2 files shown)


BalzaniEdoardo:
# htmlproofer: `--allow-missing-href`
The `post_build` htmlproofer invocations in `.readthedocs.yaml` now pass `--allow-missing-href`.

scikit-learn >= 1.9 changed the estimator HTML repr (`BaseEstimator._repr_html_`) to always render the `?` documentation-link anchor, `<a class="estimator_doc_link">?</a>`, even for estimators whose doc URL it cannot resolve. nemos models subclass `BaseEstimator`, so every doc page that displays one (e.g. `custom_solvers`, `finegrained_regularization`, `plot_03_population_glm`, `save_and_load`, `plot_01_current_injection`, `plot_04_v1_cells`, `plot_05_place_cells`, `variable_selection_zero_basis`) now carries an `<a>` with no `href`. htmlproofer reports each as `'a' tag is missing a reference`, which broke the doc build with 119 failures.

Under scikit-learn <= 1.8 that anchor is not rendered when the doc URL is empty, so htmlproofer had nothing to flag and the build passed. The failure is therefore driven by the scikit-learn 1.9 release reaching Read the Docs (the `docs` extra leaves `scikit-learn` unpinned), not by any docs change on this branch.

`--allow-missing-href` is scoped to the missing-`href` rule only: every real link, hash, image, and script is still checked. The flag requires html-proofer >= 5.0.9, already pinned in the `pre_build` step.

Issue #556, opened 2026-04-23
Multi-Dimensional Fourier Basis


ahwillia:
Reading through the documentation on the [Fourier basis](https://nemos.readthedocs.io/en/latest/background/basis/ndimensional_foureir_basis.html), I'm realizing there is an error for the multi-dimensional case (I think probably my fault, given prototype code I've sent in the past).

The form of the basis function is correct:

$$a_{\mathbf{n}}(\mathbf{x}) = \cos \left ( 2 \pi  \sum_{d=1}^{D} \frac{n_d}{P_d} x_d \right ) + i \sin \left ( 2 \pi  \sum_{d=1}^{D} \frac{n_d}{P_d} x_d \right )$$

But we are using $\mathbf{n} = (n_1, \dots, n_d) \in \mathbb{N}^d$ as the multi-index. In $d=1$ dimension we can restrict to natural numbers (since a frequency of $-n$ is equivalent to a frequency of $+n$), but in $d > 1$ dimensions we need to be a bit careful. The basis function indexed by $(1, 1)$ is different than the basis function indexed by $(1, -1)$, even though they both have the same frequency.

So I think we are currently missing some basis functions. Assuming $K$ is the maximum frequency, I think we want to initially set $\mathbf{n} = (n_1, \dots, n_d) \in \{-K, \dots, K\}$ **_and then remove all cases where the first nonzero element of n is negative_**. This removes all the redundant basis functions, like the 1D case, but it includes all the other basis functions we're currently missing.

The intuition is that, even in $d$-dimensions, we always have $cos(2 \pi \mathbf{n}^\top \mathbf{x}) = cos(2 \pi (-\mathbf{n})^\top \mathbf{x})$ and likewise $sin(2 \pi \mathbf{n}^\top \mathbf{x}) = -1 \cdot sin(2 \pi (-\mathbf{n})^\top \mathbf{x})$. So by forcing the first nonzero entry of $\mathbf{n}$ to be positive, we all the remove redundant basis functions.

ahwillia:
Something like this would be the prototype for what I think needs to be done.

```python
def make_fourier_freqs(d: int, K: int) -> jnp.ndarray:
    """Basis functions for dimension `d` with max frequency `K`"""

    all_freqs = itertools.product(range(-K, K + 1), repeat=d)

    def in_half_space(m):
        # Keep m if its first nonzero component is positive.
        for mi in m:
            if mi > 0:
                return True
            if mi < 0:
                return False

[...]

PR #581, opened 2026-06-10
add 1d squared exponential fourier basis functions, demo, tests

pgree:
The main addition is a function that chooses frequencies and weights (multiplicative scalings) of sin/cos 
basis functions. Weights and frequencies are chosen so that the Gaussian proccess defined by 
putting N(0, 1) priors on coefficients in the expansion is an approximate squared-exponential Gaussian 
process. 

- The code is primarily in `src/nemos/basis/_fourier_basis.py` 
- There are also test functions in `tests/test_efgp.py`
- There's a simple example of usage in `scripts/demo_fourier_se_basis.py`


src/nemos/basis/_fourier_basis.py
```diff
@@ -713,3 +714,240 @@ def set_params(self, **params: Any):
 
     def __repr__(self):
         return format_repr(self, exclude_keys=["fill_value"])
+
+
+def _se_quadrature(
```
sjvenditto: "se" is uninformative as a function name, I suggest something like
```suggestion
def _squared_exponential_quadrature(
```

BalzaniEdoardo:
Sorry it took me so much time to get back to this. I think we do not need an extra class for this, we can basically re-use the old class with minor additions. 


Below an example subclass that shows this may look, but this edits should be done to the `FourierBasis` class directly. Note that this works in the 1d case only so far, so that's a sketch.

What we need to add to the `FourierBasis` class is:
- a class method that takes care of defining the weights and bounds etc. (basically the `squared_exponential` below) that instantiate the object with the correct configurations.
- A little extra computation in the `evaluate` method that scales up the bases. 
- A new `_weights` attributes defaulted to None
- Maybe a property that returns the weights with a setter that checks that the weights have the correct shape.

```python

import jax

jax.config.update("jax_enable_x64", True)  # match the GP accuracy regime

import numpy as np

from nemos.basis import FourierEval, FourierSEBasis
from nemos.basis._fourier_basis import _se_quadrature


class WeightedFourierEval(FourierEval):
    """FourierEval with an optional per-frequency column weight.

    ``weights[j]`` scales BOTH the cosine and sine column of frequency ``j``
    (one spectral amplitude per frequency, shared by its cos/sin pair). This
    is exactly what the GP construction needs. ``weights=None`` -> all ones,
    i.e. identical to a plain FourierEval.
    """

    def __init__(self, *args, weights=None, **kwargs):
        # FourierEval defaults label="FourierEval", which the label validator
        # rejects on a subclass; fall back to this class's name instead.
        kwargs.setdefault("label", None)
        super().__init__(*args, **kwargs)
        self._weights = None if weights is None else np.asarray(weights, dtype=float)
        if self._weights is not None:
            n_freq = self._freq_combinations.shape[-1]
            if self._weights.shape != (n_freq,):
                raise ValueError(
                    f"``weights`` must have one entry per frequency ({n_freq},); "
                    f"got {self._weights.shape}."
                )

    def evaluate(self, *sample_pts):
        out = super().evaluate(*sample_pts)
        if self._weights is None:
            return out
        # Layout from FourierBasis.evaluate is [cos block | sin block], where
        # the cos block has one column per frequency and the sin block drops
        # the DC term. Build the matching per-column weight vector and scale.
        has_dc = self._has_zero_phase
        col_weights = np.concatenate([self._weights, self._weights[has_dc:]])
        return out * col_weights

    @classmethod
    def squared_exponential(cls, lengthscale, domain, eps, variance=1.0):
        """Build the SE-GP Fourier basis from kernel hyperparameters."""
        t0, t1 = float(domain[0]), float(domain[1])
        L = t1 - t0
        xcen = 0.5 * (t0 + t1)
        _, weights, h, m = _se_quadrature(lengthscale, variance, eps, L)
        # _se_quadrature returns [cos weights | sin weights]; the first m+1 are
        # the per-frequency amplitudes [w_0, sqrt2*w_1, ..., sqrt2*w_m].
        per_freq = np.asarray(weights[: m + 1])
        # bounds encode the physical frequency spacing: rescaling x to
        # (x - xcen) / (1/h) = h*(x - xcen) turns integer frequencies j into
        # the physical grid xi_j = j*h.
        return cls(
            frequencies=np.arange(m + 1),
            frequency_mask="all",
            bounds=(xcen, xcen + 1.0 / h),
            weights=per_freq,
        )



if __name__ == "__main__":
    # A few hyperparameter settings to compare against.
    cases = [
        dict(lengthscale=0.1, domain=(0.0, 1.0), eps=1e-4, variance=1.0),
        dict(lengthscale=0.5, domain=(-2.0, 3.0), eps=1e-6, variance=2.0),
        dict(lengthscale=0.05, domain=(0.0, 1.0), eps=1e-2, variance=0.5),
    ]

    for kw in cases:
        ref = FourierSEBasis(**kw)
        new = WeightedFourierEval.squared_exponential(**kw)

        x = np.linspace(kw["domain"][0], kw["domain"][1], 50)
        Phi_ref = np.asarray(ref.evaluate(x))
        Phi_new = np.asarray(new.evaluate(x))

        same_shape = Phi_ref.shape == Phi_new.shape
        max_abs_err = np.max(np.abs(Phi_ref - Phi_new)) if same_shape else np.inf
        ok = same_shape and np.allclose(Phi_ref, Phi_new, atol=1e-10, rtol=0)

        print(
            f"l={kw['lengthscale']:<5} var={kw['variance']:<4} eps={kw['eps']:<6} "
            f"domain={kw['domain']}  ->  n_cols ref/new = "
            f"{Phi_ref.shape[1]}/{Phi_new.shape[1] if same_shape else '?'}  "
            f"max|Δ|={max_abs_err:.2e}  {'OK' if ok else 'MISMATCH'}"
        )
        assert ok, "subclass does not reproduce FourierSEBasis"

    # Sanity check: weights=None reproduces a plain FourierEval exactly.
    plain = FourierEval(frequencies=np.arange(6), frequency_mask="all")
    wrapped = WeightedFourierEval(frequencies=np.arange(6), frequency_mask="all")
    x = np.linspace(0.0, 1.0, 17)
    assert np.allclose(plain.evaluate(x), wrapped.evaluate(x))
    print("\ndefault (weights=None) matches plain FourierEval: OK")

```

src/nemos/basis/_fourier_basis.py
```diff
@@ -713,3 +714,240 @@ def set_params(self, **params: Any):
...
+    weights :
+        weights of shape ``(2 * m + 1,)`` corresponding to the columns
+        ``[cos columns, sin columns]``: cos columns first
+        (``j = 0, 1, ..., m``), then sin columns (``j = 1, ..., m``).
+    h :
+        spacing between frequencies
+    m :
+        number of positive frequencies
+    """
+    l = float(lengthscale)
```
sjvenditto: one-letter variable names will not pass linting, I would suggest renaming all of these using a descriptive name.

for this one, you can just overwrite the input parameter

```suggestion
    lengthscale = float(lengthscale)
```

src/nemos/basis/_fourier_basis.py
```diff
@@ -713,3 +714,240 @@ def set_params(self, **params: Any):
...
+
+
+def _se_quadrature(
+    lengthscale: float, variance: float, eps: float, L: float
+) -> Tuple[jnp.ndarray, jnp.ndarray, float, int]:
+    """Finds the nodes of the equispaced quadrature in Fourier domain
+    (discretized inverse Fourier transform) for a 1-D squared-exponential
+    kernel. This involves finding the spacing between nodes ``h`` and the
+    number of nodes ``2m + 1``. This is done from a formula in --
+    https://www.sciencedirect.com/science/article/pii/S1063520324000174
```
sjvenditto: You can actually have a references section (for an example, see `MSplineBasis` in `_spline_basis.py`, which will allow you to include the full citation

tests/test_efgp.py
```diff
@@ -0,0 +1,38 @@
...
+    basis = FourierSEBasis(
+        lengthscale = length_scale, domain = domain, eps = eps, variance = variance
+    )
+    x = jnp.linspace(domain[0], domain[1], 20)
+    Phi = basis.evaluate(x)
+    K_approx = Phi @ Phi.T
+    
+    x1, x2 = jnp.meshgrid(x, x)
+    K_true = se_kernel(x1, x2, length_scale, variance)
+    max_err = jnp.max(jnp.abs(K_approx - K_true))
```
sjvenditto: you can also use `np.testing.assert_array_almost_equal` and specify the precision in decimals. e.g.
```
np.testing.assert_array_almost_equal(K_approx, K_true, decimal=6) 
```
would be the same as `eps=1E-6`

sjvenditto (changes requested):
Here are some minor comments on the `_se_quadrature` function and tests. Once we incorporate `FourierSEBasis` into `FourierBasis`, we'll want to make sure that the new attributes are properly tested, e.g. tests that attributes are set to valid values and that errors are appropriately raised when they're not.

Additionally, the scripts folder is more for testing and formatting checks, not for demos. I would suggest incorporating the demo script into a how-to guide notebook

sjvenditto:
- add weights to FourierBasis init
      - set private attribute in init, make public weights property method that access private attribute
      - update evaluate to use weights
      
- new class FourierGPKernel that inherits FourierBasis
      - length scale and everything else is here
      - start from your existing class but inherit instead and remove redundant methods

BalzaniEdoardo:
@pgree, for the FourierGPKernel basis should we make the `bounds` a required parameter? Same for the `lengthscale`? 

I would do that in case there isn't a reasonable default for these parameters.

src/nemos/basis/_fourier_basis.py
```diff
@@ -713,3 +749,248 @@ def set_params(self, **params: Any):
...
+        t0, t1 = float(domain[0]), float(domain[1])
+        if not t0 < t1:
+            raise ValueError(f"``domain`` must satisfy ``t0 < t1``; got ({t0}, {t1}).")
+
+        xis, weights, _, m = _get_nodes_weights(lengthscale, variance, eps, t1 - t0)
+
+        self._lengthscale = lengthscale
+        self._variance = variance
+        self._eps = eps
+        self._domain = (t0, t1)
```
BalzaniEdoardo: This could be redundant: `self._bounds`

src/nemos/basis/_fourier_basis.py
```diff
@@ -713,3 +749,248 @@ def set_params(self, **params: Any):
...
+
+    @property
+    def eps(self) -> float:
+        """Kernel error approximation tolerance."""
+        return self._eps
+
+    @property
+    def domain(self) -> Tuple[float, float]:
+        """Construction domain ``(t0, t1)``."""
+        return self._domain
```
BalzaniEdoardo: I know the terminology is a bit different, but I think we should only keep `bounds` which is already a property of the eval basis.

src/nemos/basis/_fourier_basis.py
```diff
@@ -713,3 +749,248 @@ def set_params(self, **params: Any):
...
+        self._variance = variance
+        self._eps = eps
+        self._domain = (t0, t1)
+        self._xis = xis
+
+        # Build on integer harmonics ``j = 0, ..., m`` (``frequency_mask="all"``
+        # keeps the ``j = 0`` DC term). ``_shift_angles`` rescales the angle so
+        # the effective frequency of harmonic ``j`` becomes ``j * h = xi_j``,
+        # and ``weights`` injects the SE spectral density.
+        FourierBasis.__init__(
```
BalzaniEdoardo, line 1053: Just a question for @pgree, do we allow only 1D? is there a way to extend to ND?

src/nemos/basis/_fourier_basis.py
```diff
@@ -713,3 +749,248 @@ def set_params(self, **params: Any):
...
+        """Construction domain ``(t0, t1)``."""
+        return self._domain
+
+    @property
+    def xis(self) -> jnp.ndarray:
+        """Non-negative frequencies ``j * h`` for ``j = 0, ..., m``."""
+        return self._xis
+
+    @property
+    def h(self) -> float:
```
BalzaniEdoardo: can we call this property `frequency_spacing`?

src/nemos/basis/_fourier_basis.py
```diff
@@ -713,3 +749,248 @@ def set_params(self, **params: Any):
...
+        """Frequency spacing, recovered from :attr:`xis`."""
+        return _grid_params_from_nodes(self._xis)[0]
+
+    @property
+    def m(self) -> int:
```
BalzaniEdoardo: can this be named something liken `n_frequencies`? note that the frequencies are also stored in `self.frequencies[0]`

So an alternative to this could be something like:

```python
@property
def n_frequencies(self) -> int:
    return len(self.frequencies[0])
```

src/nemos/basis/_fourier_basis.py
```diff
@@ -713,3 +749,248 @@ def set_params(self, **params: Any):
...
+    """
+
+    # Fourier basis is defined over the entire real line; out-of-bounds
+    # samples should not be filled with a sentinel value.
+    _apply_bounds_fill = False
+
+    def __init__(
+        self,
+        lengthscale: float,
+        domain: Tuple[float, float],
```
BalzaniEdoardo: we can call this `bounds` directly to be consistent with the other bases. Domain is probably more correct here, but probably it is not worth having a new attribute and a new property. 

What do you think @sjvenditto ?

BalzaniEdoardo (commented):
For the main code, I left some very minor comments mostly on variable naming. This all looks good to me, thanks! 

The notebook seems to be unfinished, right? I added a couple of commits so that the notebook runs. My only question is: will it be possible to extend this basis to multiple dimensions? 

Great job with the pr.

src/nemos/basis/_fourier_basis.py
pgree: there's a nice way to extend beyond 1D. you just do a tensor product of the 1D complex exponentials

pgree:
Yes, definitely possible to extend this to multiple dimensions. do you prefer to do that in this PR or a new one?

BalzaniEdoardo:
> Yes, definitely possible to extend this to multiple dimensions. do you prefer to do that in this PR or a new one?

I  think we can start merging the 1D bases, because I am working on fixing the current ND Fourier basis in this branch:

https://github.com/flatironinstitute/nemos/tree/fix-fourier

It basically fixes what Alex described in #556. It generates the full n-dim fourier basis but it doesn't use directly use the complex exp, but relies on  real sin/cos pairs. It is probably equivalent to what you are describing. 

You can take a look at the branch to see if that n-dim Fourier basis implementation is what we need.

sjvenditto:
It's shaping up very well! I think the separation between the classes makes sense so far.

Some things that need to be addressed before merging:
- there are currently no tests of the class itself, only for the underlying algorithms.
     - also we'll want to add tests for the new `weights` attribute in the original fourier basis class
- related to the testing issue, because the `FourierGP` class exists in `_fourier_basis.py` and not in `basis.py`, it is not being tested as a concrete class, and is therefore getting missed in the existing test suites for basis functions. @BalzaniEdoardo how should we handle this? Do we want to move `FourierGP` to `basis.py`? Or do we want the intermediate `FourierGP` in `_fourier_basis.py`, and then define the concrete class `FourierGPEval`?
- there are no docstring examples of how to initialize and use the `FourierGP` basis
- the docs page `efgp.md` is running, but there is no supportive text, nor is it using the `FourierGP` class in its demonstrations. also given the section titles, I wonder if this would be better as a Background section? 

I think if we want to move this PR along, we can move the documentation page to a separate PR, what do you think @BalzaniEdoardo?

EDIT: I typed this up yesterday before Argha's commit, so some of this might be outdated

docs/background/efgp.md
```diff
@@ -0,0 +1,316 @@
...
+
+With such a high-quality approximation to the kernel, the EFGP approximate predictions are practically indistinguishable from the exact GP prediction.
+
+```{code-cell} ipython3
+def compute_efgp_mean(pred_stimuli, obs_stimuli, obs_firing_rate, noise_sigma):
+    obs_eval = basis.evaluate(obs_stimuli)
+    n_weights = obs_eval.shape[1]
+    augmented_K_obs = (obs_eval.T @ obs_eval) + (noise_sigma**2) * jnp.eye(n_weights)
+    L = jnp.linalg.cholesky(augmented_K_obs)
+    beta = jnp.linalg.solve(L.T, jnp.linalg.solve(L, obs_eval.T @ obs_firing_rate))
```
BalzaniEdoardo: we can use directly `jax.scipy.linalg.cho_solve` here

docs/background/efgp.md
```diff
@@ -0,0 +1,316 @@
...
+Once we specify our prior beliefs by choosing a kernel, finding the predictive mean of the GP is remarkably simple!
+When we try it out on the synthetic dataset from above, we can see that the GP is doing a fairly good job at capturing the relationship between the stimulus strength and the neural firing rate.
+
+```{code-cell} ipython3
+def compute_predictive_mean(pred_stimuli, obs_stimuli, obs_firing_rate, noise_sigma):
+    n_obs = obs_stimuli.shape[0]
+    K_obs  = se_kernel(obs_stimuli, obs_stimuli) + (noise_sigma**2) * jnp.eye(n_obs)
+    K_cross = se_kernel(pred_stimuli, obs_stimuli)
+    L = jnp.linalg.cholesky(K_obs)
+    w = jnp.linalg.solve(L.T, jnp.linalg.solve(L, obs_firing_rate))
```
BalzaniEdoardo: use `jax.scipy.linalg.cho_solve` as a simplification

docs/background/efgp.md
```diff
@@ -0,0 +1,316 @@
...
+axes[2].set_xlabel('stimulus')
+```
+
+Once we specify our prior beliefs by choosing a kernel, finding the predictive mean of the GP is remarkably simple!
+When we try it out on the synthetic dataset from above, we can see that the GP is doing a fairly good job at capturing the relationship between the stimulus strength and the neural firing rate.
+
+```{code-cell} ipython3
+def compute_predictive_mean(pred_stimuli, obs_stimuli, obs_firing_rate, noise_sigma):
+    n_obs = obs_stimuli.shape[0]
+    K_obs  = se_kernel(obs_stimuli, obs_stimuli) + (noise_sigma**2) * jnp.eye(n_obs)
```
BalzaniEdoardo: try to split into multilne for readability

docs/background/efgp.md
```diff
@@ -0,0 +1,316 @@
...
+Our goal is to learn this neuron's **tuning curve**: a function which defines the neural firing rate in response to any stimulus strength.
+In this example, we've defined the ground truth tuning curve as a cubic function of the stimulus value and generated synthetic measurements corrupted by Gaussian noise.
+
+```{code-cell} ipython3
+def evaluate_tuning(stimulus):
+    return 0.2 * (stimulus ** 3) - 2.5 * (stimulus ** 2) + 6 * stimulus
+
+def generate_observations(key, n_obs, stimulus_min, stimulus_max, noise_sigma):
+    stimulus_key, subkey = jr.split(key)
+    obs_stimuli = jr.uniform(stimulus_key, shape = (n_obs,), minval = stimulus_min, maxval = stimulus_max)
```
BalzaniEdoardo:
```suggestion
    obs_stimuli = jr.uniform(
        stimulus_key, 
        shape=(n_obs,), 
        minval=stimulus_min, 
        maxval=stimulus_max,
    )
```

docs/background/efgp.md
```diff
@@ -0,0 +1,316 @@
...
+    true_firing_rate = evaluate_tuning(obs_stimuli)
+    firing_rate_key, subkey = jr.split(subkey)
+    obs_firing_rate = true_firing_rate + jr.normal(firing_rate_key, shape = (n_obs,)) * noise_sigma
```
BalzaniEdoardo:
```suggestion
    obs_firing_rate = (
        true_firing_rate +
        jr.normal(firing_rate_key, shape = (n_obs,)) * noise_sigma
    )
```

docs/background/efgp.md
```diff
@@ -0,0 +1,316 @@
...
+    return obs_stimuli, obs_firing_rate, true_firing_rate
+
+key = jr.key(seed = 0)
+n_obs = 100
+stimulus_min = 0
+stimulus_max = 10
+noise_sigma = 1
+obs_stimuli, obs_firing_rate, true_firing_rate = generate_observations(key, n_obs, stimulus_min, stimulus_max, noise_sigma)
```
BalzaniEdoardo: split this one on a multiline too for readability

docs/background/efgp.md
```diff
@@ -0,0 +1,316 @@
...
+    diff = stim_1[:, None] - stim_2[None, :]
+    return output_scale * jnp.exp(-0.5 * diff**2 / length_scale**2)
+
+# odd count so the symmetric grid contains an exact 0 at index n_test_stimuli // 2;
+# the kernel-approximation cell uses that index as the zero-difference reference point.
+n_test_stimuli = 101
+test_stimuli_diffs = jnp.linspace(-0.5*stimulus_max, 0.5*stimulus_max, n_test_stimuli)
+kernel_evals = se_kernel(test_stimuli_diffs)
+smaller_length_scale_kernel_evals = se_kernel(test_stimuli_diffs, length_scale = 0.4)
+larger_length_scale_kernel_evals = se_kernel(test_stimuli_diffs, length_scale = 1.0)
```
BalzaniEdoardo: maybe drop the `kernel` from the name to keep it shorter. I am attempting to shorten the lines so that they fit better in codeblocks. The ideal would be having minimal cropping on the rendered docpage. 

Below a screenshot of what I mean, for some code lines you need to scroll right

Image

docs/background/efgp.md
```diff
@@ -0,0 +1,316 @@
...
+    obs_stimuli = jr.uniform(stimulus_key, shape = (n_obs,), minval = stimulus_min, maxval = stimulus_max)
+    true_firing_rate = evaluate_tuning(obs_stimuli)
+    firing_rate_key, subkey = jr.split(subkey)
+    obs_firing_rate = true_firing_rate + jr.normal(firing_rate_key, shape = (n_obs,)) * noise_sigma
+    return obs_stimuli, obs_firing_rate, true_firing_rate
+
+key = jr.key(seed = 0)
+n_obs = 100
+stimulus_min = 0
+stimulus_max = 10
```
BalzaniEdoardo: This variables can be renamed to `max_stim` or `stim_max` and same with min. I am trying to shorten the lines so that all the code fits on screen.

docs/background/efgp.md
```diff
@@ -0,0 +1,316 @@
...
+
+n_pred = 300
+pred_stimuli = jnp.linspace(stimulus_min, stimulus_max, n_pred)
+```
+
+```{code-cell} ipython3
+fig, axes = plt.subplots(1, 1)
+axes.scatter(obs_stimuli, obs_firing_rate, s = 10, c = 'k', label = 'observations', zorder = 5)
+sorted_indices = jnp.argsort(obs_stimuli)
+axes.plot(obs_stimuli[sorted_indices], true_firing_rate[sorted_indices], lw = 2, c = 'r', label = 'ground truth')
```
BalzaniEdoardo:
```suggestion
axes.plot(
    obs_stimuli[sorted_indices], 
    true_firing_rate[sorted_indices], 
    lw = 2, c = 'r', label = 'ground truth',
)
```

docs/background/efgp.md
```diff
@@ -0,0 +1,325 @@
...
+import time
+
+dataset_sizes = jnp.logspace(1, 4, num = 9, dtype = int)
+times = []
+for (i, size) in enumerate(dataset_sizes):
+    key = jr.key(seed = i)
+    stimuli, firing_rate, true_firing_rate = generate_observations(key, size, stimulus_min, stimulus_max, noise_sigma)
+    traced = compute_predictive_mean.trace(pred_stimuli, stimuli, firing_rate, noise_sigma)
+    lowered = traced.lower()
+    compiled = lowered.compile()
```
BalzaniEdoardo:
```suggestion

def compile(func, *args, **kwargs):
    """Compile ``func`` ahead of time for the shapes of the given arguments.                                                                                                                           
    
    ``jax.jit`` compiles on the first call, so timing that call would include tracing                                                                                                                             and compilation.                                                                                                                               
    """                                                                                                                                                                                    
    return jax.jit(func).trace(*args, **kwargs).lower().compile()
    

dataset_sizes = jnp.logspace(1, 4, num = 9, dtype = int)
times = []
for (i, size) in enumerate(dataset_sizes):
    key = jr.key(seed = i)
    stimuli, firing_rate, true_firing_rate = generate_observations(key, size, stimulus_min, stimulus_max, noise_sigma)
    compiled = compile(
        compute_predictive_mean, 
        stimuli, firing_rate, true_firing_rate, noise_sigma,
     )

```

docs/background/efgp.md
```diff
@@ -0,0 +1,325 @@
...
+    start_time = time.perf_counter()
+    mu_pred = compiled(pred_stimuli, stimuli, firing_rate, noise_sigma)
```
BalzaniEdoardo:
```suggestion
    mu_pred = compiled(pred_stimuli, stimuli, firing_rate, noise_sigma).block_until_ready()
```

docs/background/efgp.md
```diff
@@ -0,0 +1,325 @@
...
+axes.plot(test_stimuli_diffs, approx_kernel, lw = 2, c = 'b', label = 'approximated kernel', alpha = 0.75)
+axes.set_xlabel('difference between stimuli')
+axes.set_ylabel('covariance between stimuli')
+axes.legend(frameon = False)
+```
+
+With such a high-quality approximation to the kernel, the EFGP approximate predictions are practically indistinguishable from the exact GP prediction.
+
+```{code-cell} ipython3
+@jax.jit
```
BalzaniEdoardo: If we add jit as part of the compile this is not-necessary. 

I think jitting at compile time makes it clear where the `.trace()` comes from

docs/background/efgp.md
```diff
@@ -0,0 +1,325 @@
...
+axes[2].plot(stimuli_range, samples_large.T, lw = 2, c = 'b', alpha = 0.3)
+axes[2].set_title('length scale = 1.0')
+axes[2].set_xlabel('stimulus')
+```
+
+Once we specify our prior beliefs by choosing a kernel, finding the predictive mean of the GP is remarkably simple!
+When we try it out on the synthetic dataset from above, we can see that the GP is doing a fairly good job at capturing the relationship between the stimulus strength and the neural firing rate.
+
+```{code-cell} ipython3
+@jax.jit
```
BalzaniEdoardo: as below (i would remove this one)

docs/background/efgp.md
```diff
@@ -0,0 +1,325 @@
...
+
+```{code-cell} ipython3
+dataset_sizes = jnp.logspace(1, 4, num = 9, dtype = int)
+efgp_times = []
+for (i, size) in enumerate(dataset_sizes):
+    key = jr.key(seed = i)
+    stimuli, firing_rate, true_firing_rate = generate_observations(key, size, stimulus_min, stimulus_max, noise_sigma)
+    pred_eval = basis.evaluate(pred_stimuli)
+    obs_eval = basis.evaluate(stimuli)
+    traced = compute_efgp_mean.trace(pred_eval, obs_eval, firing_rate, noise_sigma)
```
BalzaniEdoardo: re-use compile utility

BalzaniEdoardo (commented):
I left some review mostly on code simplifications. I think overall the note flows nicely, the code lines can be very long at times and they may be cropped out of the rendered code cell on the doc website.

Consider shorten some of the variable names, create utility functions (I created one for compile) and similar, split long lines into multiple lines, etc.

BalzaniEdoardo:
Splitting the basis in `BaseFourer` (common machinery), `FoureirEval` and `FourierGP`

## Shared → `BaseFourier`

- `evaluate` — reads only `_freq_combinations`, `_has_zero_phase`, `_shift_angles`, `weights`
- `_shift_angles` — hook, also the future `FourierConv` 
- `_freq_combinations` + `_has_zero_phase` — the contract; each subclass fills them its own way
- `weights` + `_check_weights`
- `n_basis_funcs` — read-only derived; `2*K - _has_zero_phase` already equals `len(weights)` for the GP
- `ndim` / `_n_inputs` / `_check_ndim` — GP hardcodes 1 today, not forever
- `evaluate_on_grid`, `__repr__`
- `_is_complex = True`, `_apply_bounds_fill = False`

## `FourierEval` only

- `frequencies` property + setter
- `arange_constructor`, `_process_tuple_frequencies`, `_check_and_sort_frequencies`, `_check_array_properties`, `_format_error_message` — spec parsing
- `frequency_mask` property + setter, `_set_array_frequency_mask`, `_get_frequency_pairs_from_callable`
- `_combinator_builder`, `_signed_box_axes`, `_half_space_selection`, `_move_dc_first`, `_get_all_frequency_pairs` — half-space construction
- `masked_frequencies`
- `set_params` override — co-setting frequencies + mask
- "Resetting ``frequency_mask``" warning
- `bounds` as period, optional, inferred from data

## `FourierGP` only

- `lengthscale`, `variance`, `eps`, `xis`, `frequency_spacing`, `n_frequencies`
- `_get_nodes_weights`, `_grid_params_from_nodes`
- `_shift_angles` override, `_get_samples` override
- `bounds` as construction domain — mandatory, never `None`

## Missing machinery

- setters for `lengthscale` / `variance` / `eps` — in `get_params`, so `set_params(lengthscale=0.5)` raises `AttributeError: no setter`; no CV over kernel hyperparameters
- `bounds` setter that recomputes the grid — `h`, `m` depend on `L`; after `b.bounds=(0, 10)` the object has 17 functions and `clone(b)` has 87
- `bounds` setter rejecting `None` — `__init__` rejects it, the setter doesn't; then `TypeError: cannot unpack non-iterable NoneType`
- one private `_rebuild_grid()` — shared by the three above
- drop `_n_basis_funcs` shadow — `AtomicBasisMixin.__init__` caches it, the derived property shadows it forever
- `_check_n_basis_min` — absent for both; fine if intentional, say so in the dev note

`split_by_feature`, `set_input_shape` and `n_output_features` need nothing new: they all route through `n_basis_funcs`.

## Bugs the split removes

Both inherited setters mutate `_freq_combinations` without touching `weights` or `xis`, on `FourierGP(lengthscale=0.2, bounds=(0., 1.), eps=1e-4)`:

| action | result |
| --- | --- |
| `bas.frequency_mask = "no-intercept"` | `compute_features` raises `Incompatible shapes for broadcasting: shapes=[(10, 16), (17,)]` (the collab's `(10, 6376)` vs `(6377,)`) |
| `bas.frequencies = 3` | `n_basis_funcs` 17 → 5, `weights` stays `(17,)`, `xis` stays `(9,)` |
| `bas.bounds = (0., 10.)` | `n_basis_funcs` stays 17, `clone(bas)` returns 87 — sklearn clone contract broken |
| `bas.set_params(lengthscale=0.5)` | `AttributeError: property 'lengthscale' of 'FourierGP' object has no setter` |

PR #582, opened 2026-06-19
PP-GLM validator

macari216:
This PR adds PPGLM validator, MockPPGLM, params and data modules and validator tests. First of three stacked PRs (validator -> log-likelihood -> base class). 

PPGLMValidator inherits from GLMValidator. Tests here cover only PPGLM-specific checks. Since PPGLM uses GLMParams module, all params checks shuld pass automatically. _validate_inputs does not exist by design since those checks run during input preprocessing instead.


macari216:
This change is part of the following stack:

- #582 ◀

<sub>Change managed by [git-spice](https://abhinav.github.io/git-spice/).</sub>

src/nemos/pp_glm/data.py
```diff
@@ -0,0 +1,25 @@
+"""PP-GLM input data definitions and type aliases."""
+
+import equinox as eqx
+from jaxtyping import Array, Float, Int
+
+class X_ppglm(eqx.Module):
```
BalzaniEdoardo: The class names should be in `PascalCase`

src/nemos/pp_glm/data.py
```diff
@@ -0,0 +1,25 @@
...
+
+    times: Float[Array, "n_events"]
+    ids: Int[Array, "n_events"]
+
+
+class y_ppglm(eqx.Module):
+    """Preprocessed spikes for PP-GLM."""
+
+    times: Float[Array, "n_spikes"]
+    ids: Int[Array, "n_spikes"]
```
BalzaniEdoardo: `ids` and `idx` are not informative enough, can you use more interpretable names if possible?

src/nemos/pp_glm/mock_ppglm.py
```diff
@@ -0,0 +1,138 @@
+from typing import Any, Callable, Optional, Union
```
BalzaniEdoardo: this all class is for testing, right? it should live in a test module, a script in the `tests/` folder, or in `tests/conftest.py` if it is used across modules.

src/nemos/pp_glm/validation.py
```diff
@@ -0,0 +1,213 @@
...
+from ..base_validator import RegressorValidator
+from ..glm.params import GLMParams, GLMUserParams
+from ..glm.validation import GLMValidator, from_glm_params, to_glm_params
+from .params import PPGLMParamsWithKey
+from .data import X_ppglm, y_ppglm
+
+
+def to_pp_glm_params_with_key(params: GLMParams, random_key: jnp.array):
+    """Map from PPGLMParams to PPGLMParamsWithKey by appending a jax random key.
+    The key is converted from uint32 to float to avoid solver initialization error"""
```
BalzaniEdoardo:
```suggestion
    """Map from PPGLMParams to PPGLMParamsWithKey.
    
     Map from PPGLMParams to PPGLMParamsWithKey by appending a jax random key.
    The key is converted from uint32 to float to avoid solver initialization error"""
```

src/nemos/pp_glm/validation.py
```diff
@@ -0,0 +1,213 @@
...
+    return PPGLMParamsWithKey(params, random_key.astype(float))
```
BalzaniEdoardo:
```suggestion
    return PPGLMParamsWithKey(params, random_key.astype(params.coef.dtype))
```

Maybe even more robust.

src/nemos/pp_glm/validation.py
```diff
@@ -0,0 +1,213 @@
...
+        y: Optional[y_ppglm] = None,
+    ):
+        """
+        Validate consistency between parameters and inputs for PP-GLM.
+
+        For single-neuron PP-GLM, only validates feature consistency with X.
+        Does not validate y since it's 1D (single neuron, no neuron axis to check).
+        """
+        if X is not None:
+            n_features = jax.tree_util.tree_map(lambda p: p.shape[0], params.coef)
```
BalzaniEdoardo, line 102: this returns a stuct of the same kind of `coef`, so the division after doesn't work if for example, `coef` is a list. That can happen in GLM and GLMHMM, one can structure coef as a pytree.


```suggestion
            n_features = jax.tree.reduce(jnp.add, jax.tree_util.tree_map(lambda p: p.shape[0], params.coef))
```

src/nemos/pp_glm/validation.py
```diff
@@ -0,0 +1,213 @@
...
+
+            if n_predictors_params != predictors_X.size:
+                raise ValueError(
+                    "Inconsistent number of features. "
+                    f"Model coefficients assume {n_predictors_params} groups and {self.n_basis_funcs} basis functions, "
+                    f"X has {predictors_X.size} groups instead!"
+                )
+
+            # there might be a better way to do this
+            if not jnp.allclose(jnp.arange(predictors_X.size), predictors_X):
```
BalzaniEdoardo: This could be a way `all(predictors_X == jnp.arange(predictors_X.size))`.

src/nemos/pp_glm/validation.py
BalzaniEdoardo: Also, this matters only if it makes sense to allow `X `to be pytree. One convenient idea is that `X` could be a list or a dict of the per-area presynaptic neurons. it would facilitate bookkeeping.

BalzaniEdoardo: And if we allow that, we need to add a check on the tree structure, specifically that `X` has the same struct as `params.coef`.

BalzaniEdoardo (changes requested):
Some comments that should be quick, except the `X` as pytree comment. Technically, it is easy to fix in this PR, but for that to fully work, all the calculation should be tree compatible.

src/nemos/pp_glm/validation.py
macari216: > Also, this matters only if it makes sense to allow `X `to be pytree. One convenient idea is that `X` could be a list or a dict of the per-area presynaptic neurons. it would facilitate bookkeeping.

I ‘m not sure this is easily compatible because the log-likelihood computation logic is very different from the discrete GLM. There, all leaves' pre-nonlinearity contributions are summed before a single inverse_link_function call. Here instead nonlinearity is applied per spike inside the scan, so if we sum leaf contributions after the scan we would sum post-nonlinearity values which is wrong (though summing them before the nonlinearity, inside the scan, should still work). Also, ‘y’ contains indices of y spikes into sorted X, so if X became multiple leaves, y/mc samples would need to become pytrees too, with a separate timestamp_idx per leaf.

macari216: But overall I think it's possible to make it work and hopefully without additional memory overhead (compared to a single-leaf X) in the vmap-over-scan with dynamic indexing log-likelihood computation

src/nemos/pp_glm/data.py
macari216: Changed to PredictorsPPGLM, SpikesPPGLM, and MCSamplePPGLM

src/nemos/pp_glm/data.py
macari216: in X: changed 'ids' -> 'predictor_ids'
in y: 'ids' -> 'neuron_ids'
in y and mc sample: 'idx' -> 'timestamp_idx' (alternatively 'idx_in_X'?)

src/nemos/pp_glm/mock_ppglm.py
macari216: Moved it in test_ppglm_validator; I don't think it's reusable

src/nemos/pp_glm/validation.py
macari216: done

src/nemos/pp_glm/validation.py
macari216: agreed, thanks

src/nemos/pp_glm/validation.py
macari216: yes that's better!

src/nemos/pp_glm/data.py
```diff
@@ -7,20 +7,20 @@
 class PredictorsPPGLM(eqx.Module):
     """Preprocessed predictors for PP-GLM."""
 
-    times: Float[Array, "n_events"]
-    predictor_ids: Int[Array, "n_events"]
+    times: Float[Array, "n_events"]  # noqa: F821
```
macari216, line 10: suppresses flake8 check for undefined names

src/nemos/pp_glm/validation.py
```diff
@@ -1,24 +1,28 @@
...
+import numpy as np
+from pynapple import IntervalSet, Ts, Tsd, TsGroup
 
 from ..base_validator import RegressorValidator
 from ..glm.params import GLMParams, GLMUserParams
 from ..glm.validation import GLMValidator, from_glm_params, to_glm_params
 from .data import PredictorsPPGLM, SpikesPPGLM
 from .params import PPGLMParamsWithKey
 
+TimeSeriesInput: TypeAlias = "Union[np.ndarray, jnp.ndarray, list, dict, Ts, TsGroup]"
```
macari216, line 17: This lists all allowed raw input types (using the same format as DESIGN_INPUT_TYPE)

tox.ini
```diff
@@ -79,9 +79,9 @@ commands =
     black docs/scripts
     black tests
     isort tests --profile=black
-    flake8 --config={toxinidir}/tox.ini src
-    flake8 --config={toxinidir}/tox.ini docs/scripts
-    flake8 --config={toxinidir}/tox.ini tests
+    flake8 --config="{toxinidir}/tox.ini" src
+    flake8 --config="{toxinidir}/tox.ini" docs/scripts
+    flake8 --config="{toxinidir}/tox.ini" tests
```
macari216, line 85: works with path names that contain spaces

src/nemos/pp_glm/validation.py
```diff
@@ -0,0 +1,372 @@
...
+            # check recording_time span doesn't extend far beyond data
+            uncovered = span.set_diff(obj.time_support.time_span())
+            if uncovered.tot_length() > tol:
+                raise ValueError(
+                    f"recording_time span extends more than {tol}s beyond "
+                    f"{name} time support. Uncovered duration: {uncovered.tot_length():.3f}s. "
+                    f"recording_time span should match the time support of your data."
+                )
+
+    def _validate_time_series(self, time_series, name: str) -> int:  # noqa: C901
```
BalzaniEdoardo: we are using the convention that types are in the signature but not in the docsrings

BalzaniEdoardo: apply this rule everywhere

src/nemos/pp_glm/validation.py
```diff
@@ -0,0 +1,372 @@
...
+                "\nThe provided coef, intercept and random_key have shapes ``{}`` and ``{}`` "
+                "instead."
+            ),
+        ),
+        *RegressorValidator.params_validation_sequence[3:],
+        ("validate_intercept_shape", None),
+        ("validate_random_key", None),
+    )
+
+    def validate_random_key(self, random_key, dtype: type = jnp.uint32):
```
BalzaniEdoardo: type annotations

src/nemos/pp_glm/validation.py
```diff
@@ -0,0 +1,372 @@
...
+        Raises
+        ------
+        ValueError
+            If any component time series is empty.
+        TypeError
+            If time_series is not one of the supported types.
+        """
+
+        # convert jax to numpy
+        if isinstance(time_series, jax.Array):
```
BalzaniEdoardo, line 215: this function is too complex, i would split the cases

src/nemos/pp_glm/validation.py
BalzaniEdoardo: this also doesn't work, it is part of the pipeline but returns None instead of the parameters. The pipeline is all

validated_params = getattr(self, name)(validated_params, **kwargs)

so it passes the params in and out

BalzaniEdoardo: also it will receive the params as inupt, not the random key

src/nemos/pp_glm/validation.py
```diff
@@ -0,0 +1,372 @@
...
+
+            if not all(predictors_X == jnp.arange(predictors_X.size)):
+                raise ValueError(
+                    "Inconsistent predictor IDs. "
+                    f"Predictor IDs must be consecutive integers from 0 to {predictors_X.size - 1}."
+                )
+
+    def get_empty_params(self, X, y) -> GLMParams:
+        """Return the param shape given the input data."""
+        n_features = int(jnp.unique(X.ids).size * self.n_basis_funcs)
```
BalzaniEdoardo: should be updated

src/nemos/pp_glm/validation.py
```diff
@@ -0,0 +1,372 @@
...
+        """
+        if n_series <= 1:
+            raise ValueError(
+                f"y must contain more than 1 time series for {type(self).__name__}, "
+                f"got {n_series}."
+            )
+
+    def get_empty_params(self, X, y) -> GLMParams:
+        """Return the param shape given the input data."""
+        n_features = int(jnp.unique(X.ids).size * self.n_basis_funcs)
```
BalzaniEdoardo: updated too (X.ids doesn't exist)

BalzaniEdoardo (changes requested):
Some changes

BalzaniEdoardo (approved):
great! LGTM

PR #583, opened 2026-06-24
Fix ND Fourier Basis

### LINKED ISSUE #556: Multi-Dimensional Fourier Basis

BalzaniEdoardo:
This PR fixes #556 by filtering redundant frequency pairs in high dimension.

The changes are mostly in the frequency pairs builder. A full grid of positive an negative frequencies is built and then filtered out appropriately.

In particular, the rule for identifying non-redundant frequency indexes is:

For every index of the type $J = (0,0,0,..., m_i, m_{i+1},...)$ with $i>0$, enforce $m_i > 0$, otherwise  $J$ and $-J$ will result in the same basis function.

```python
>>> import nemos as nmo
>>> import numpy as np
>>> bas = nmo.basis.FourierEval(ndim=3,frequencies=[np.arange(2), np.arange(2), np.arange(2)], frequency_mask="all")
>>> bas._freq_combinations
Array([[ 0.,  0.,  0.,  0.,  0.,  1.,  1.,  1.,  1.,  1.,  1.,  1.,  1., 1.],
       [ 0.,  0.,  1.,  1.,  1., -1., -1., -1.,  0.,  0.,  0.,  1.,  1., 1.],
       [ 0.,  1., -1.,  0.,  1., -1.,  0.,  1., -1.,  0.,  1., -1.,  0., 1.]], dtype=float32)
```

## Question for reviewers

For now, I kept the possibility of setting a boolean mask for masking the frequency, but that's more inconvenient than actually providing a function that filters them, since now the mask has to have a less intuitive shape and symmetry:

- The shape of the mask is $(m_0 + 1, 2 \cdot m_1 + 1, ...., 2\cdot m_n+1)$ where the first axis has only positive freq + 0 (if the DC term is included). The other axis have positive and negative freqs, and that's why they have 2m+1 terms.
- The symmetry is that for every multi-index `J`, `mask[0, J] == mask[0, -J]`, or else `np.all(np.flip(mask[0]) == mask[0])`. The first axis is always positive, so there is no redundant frequencies, but when the first frequency is. 0, then `(0,-J)` and `(0,J)` represent are redundant and so they should have  a consistent mask.

Do we want to keep this masking or do we want to only allow frequency selection via a callable?


sjvenditto (changes requested):
As we discussed offline: I would change the ND boolean max to be instead be a 1D mask on the final set of frequency combinations. This way we don't need to do any checks for symmetry on redundant frequency pairs in the frequency box. Then, in the docstring and in the basis background page on ND Fourier basis, I would give a pointer on how to access the set of frequency combinations from an initialized basis object (i.e. `basis._freq_combinations` or add public attribute) to use to construct the mask.

BalzaniEdoardo:
addressed @sjvenditto concerns.

I figured we already had a read-only property showing all the frequency combinations, it is called `masked_frequencies`, I pointed to that in the docstrings and when I talk about how to filter frequencies with a boolean mask.

PR #585, opened 2026-06-26
Basis bounds safety

BalzaniEdoardo:
## Summary
This basis add safety guardrails for basis as transformers for `Eval` guaranteeing input independent transformations. 

Before this PR, the `transform` method of `Eval` transformers with no fixed `bounds` would generate a basis set that adaptively covers the input data range. This would have created silent bugs when transformers were used in cross-validation (a common use case): the `transform` map wasn't consistent across folds. If folds covered different ranges the basis set between folds would have differed. This is most likely not desirable since the resulting features wouldn't be comparable. 

In this PR we require `bounds` to be set for `TransformerBasis` with a runtime check, similar to that of the input shape.  Conceptually, this separate the behavior of `compute_features` from that of`transform`:

- `compute_features` keeps working as usual. When `bounds` are not set, it is an input dependent map that covers the data domain. This behavior is conceptually similar to a z-score, which is not a fixed non-linearity. 
- `transform` will behave as a fixed map instead.

## Motivation

When an `Eval` basis has `bounds=None`, its domain is taken from the min/max of the input (see `min_max_rescale_samples`). This is fine for a one-shot `compute_features`, but silently incorrect when the basis is wrapped in a `TransformerBasis` and cross-validated: each fold re-fits the domain to its own data range, so a *different* transformation is applied per fold. The model selection then compares configurations that are not on a common domain, with no error or warning. The only safe configuration is explicit `bounds`, so we now require it at the point where the hazard exists (transformer `fit`/`transform`/`fit_transform`) rather than leaving it to the user to know.

### `_bounds_define_domain` flag

A class attribute `_bounds_define_domain` is added to `Basis` (default `False`) and set to `True` on the base classes whose domain is data-derived when `bounds` is unset:

| Base class | Eval subclasses gated |
|---|---|
| `SplineBasis` | `BSplineEval`, `CyclicBSplineEval`, `MSplineEval` |
| `RaisedCosineBasisLinear` / `RaisedCosineBasisLog` | `RaisedCosineLinearEval`, `RaisedCosineLogEval` |
| `OrthExponentialBasis` | `OrthExponentialEval` |
| `FourierBasis` | `FourierEval` |

The flag lives on the math base class, so the corresponding `Conv` bases inherit `True` as well — but they carry no `bounds` attribute, which the gate uses to exclude them (see below).

### Transformer safety gate

`TransformerBasis._check_initialized` (called by `fit`, `transform`, `fit_transform`) now iterates over the components and raises `RuntimeError` when a component is `_bounds_define_domain=True`, exposes a `bounds` attribute, and has any unset (`None`) bound. The three conditions together select exactly the unsafe case:

- `_bounds_define_domain` excludes bases with a fixed domain (e.g. `IdentityEval`).
- `hasattr(b, "bounds")` excludes `Conv` bases (flag inherited, but no `bounds`) and `Zero`.
- the `None`-leaf check (with `is_leaf` keeping `None`) catches both `bounds=None` and a partially-set `(None, high)`.

### Eval mixin split

`EvalBasisMixin` is split into two layers:

- `EvalBasisMixin` — the evaluation lifecycle (`_compute_features` without fill, `setup_basis`) shared by all evaluational bases.
- `BoundedEvalBasisMixin(EvalBasisMixin)` — owns `bounds`/`fill_value` storage, the out-of-bounds fill in `_compute_features`, and `_apply_fill_value`.

All bounded `Eval` bases now inherit `BoundedEvalBasisMixin`. `Zero` uses the bare `EvalBasisMixin` and no longer carries a `bounds` attribute at all, which let us delete its special-case `bounds` property/setter (it previously had to override `bounds` to forbid them). `IdentityEval` inherits `BoundedEvalBasisMixin` but keeps `_apply_bounds_fill = False` (its `evaluate` already masks out-of-bounds samples) and leaves `_bounds_define_domain` at `False`, so it is exempt from the gate.

### Docs

The transformer how-to, the sklearn pipeline CV demo, and the variable-selection-zero-basis guide now set `bounds` on their `Eval` components, with prose explaining why a transformer needs a fixed domain. New docstring `Notes` on `BoundedEvalBasisMixin._compute_features` describe the data-derived-domain behavior; a `CustomBasis` note warns that data-dependent user functions are unsafe across folds (see open questions).

### Additional unrelated float 64 enabling

After merging @rob-the-bot PR, the convergence issues in the tutorials with low thresholds and float32, which are inevitable, becomes more evident. 

I enabled float 64 precision in all tutorials, also removed the tolerance which does not compromise the fit. The only tutorial which still does not converge is the place cell. The design matrix is rank deficient, I did not check the conditioning. Even small `Ridge` penalization does have a large effect on the coefficients. I'll open an issue about this. 

1. **Gate location.** The check runs at `fit`/`transform`/`fit_transform`. A `to_transformer()`-time check would fail faster, but cannot catch `bounds` set/cleared afterwards or swapped in via a `param_grid` (`transformerbasis__basis=...`). The runtime location covers all paths at the cost of re-checking on each call. Keep at runtime?
2. **`CustomBasis`.** We cannot introspect user-supplied functions, so it is exempt and only documented as potentially unsafe (a function that normalizes by the input mean/std would differ across folds). Is a docstring warning sufficient, or do we want an opt-in flag for users to declare a custom basis data-dependent?

### DIFF
--- a/docs/how_to_guide/plot_05_transformer_basis.md
+++ b/docs/how_to_guide/plot_05_transformer_basis.md
@@ -99,7 +99,15 @@ As with any `sckit-learn` transformer, the `TransformerBasis` implements `fit`, 
 
 ## Setting up the TransformerBasis
 
-At this point, we have an object equipped with the necessary methods. So, all we need to do is concatenate the inputs into a single array and call fit_transform, right? **Not quite.** By default, `TransformerBasis` assumes that `inp` has one column per input in the composition. So in this example, with two one-dimensional basis functions, `TransformerBasis` expects an input of shape `(n_samples, 2)`.
+At this point, we have an object equipped with the necessary methods. So, all we need to do is concatenate the inputs into a single array and call fit_transform, right? **Not quite.**
+
+The `TransformerBasis` has two additional requirements: input shape and bounds.
+
+### Setting the `input_shape`
+
+#### The Problem
+
+By default, `TransformerBasis` assumes that `inp` has one column per input in the composition. So in this example, with two one-dimensional basis functions, `TransformerBasis` expects an input of shape `(n_samples, 2)`.
 
 However, our actual input has shape `(n_samples, 6)`:
 
@@ -109,7 +117,7 @@ However, our actual input has shape `(n_samples, 6)`:
 
 When concatenated, this results in a **6-column input**, and our `TransformerBasis` doesn't know how to split those 6 columns.
 
-This mismatch leads to an error when calling fit_transform:
+This mismatch leads to an error when calling fit_transform (the cell also sets `bounds`, the second requirement, so that the failure below is about the input shape only; bounds are covered [in the next section](#setting-up-the-bounds)):
 
 
 ```{code-cell} ipython3
@@ -117,48 +125,18 @@ This mismatch leads to an error when calling fit_transform:
 
 # reinstantiate the basis transformer for illustration purposes
 composite_basis = counts_basis + speed_basis
-trans_bas = (composite_basis).to_transformer()
-# concatenate the inputs
-inp = np.concatenate([counts, speed[:, np.newaxis]], axis=1)
-print(inp.shape)
-
-trans_bas.fit_transform(inp)
-
-```
-
-There is a second requirement: any basis that derives its domain from `bounds` must have `bounds` set before it can be used as a transformer. This is a deliberate safety choice. With `bounds` left unset, the domain is inferred from the data, so during cross-validation each fold would re-fit the domain to its own data range. The transformation would then differ from fold to fold, which is rarely what you want when cross-validating.
-
-Let's see this with an example:
-
-```{code-cell} ipython3
-:tags: [raises-exception]
-
-# reinstantiate the basis transformer for illustration purposes
-composite_basis = counts_basis + speed_basis
-trans_bas = (composite_basis).to_transformer()
-# set the input shape but NOT the bounds
-trans_bas.set_input_shape(counts, speed)
-# concatenate the inputs
-inp = np.concatenate([counts, speed[:, np.newaxis]], axis=1)
-print(inp.shape)
-
-trans_bas.fit_transform(inp)
-
-```
-
-### Define the `bounds`
-
-As shown above, any basis whose domain is defined by `bounds` must have it set before the basis is used as a transformer. Only `Eval` bases carry `bounds`; convolutional bases do not.
-
-Let's set the bounds on the `Eval` component (`speed`) of `composite_basis`.
-
-```{code-cell} ipython3
-
-# get the Eval basis using its label and set the bounds
+# set the bounds now (see "Setting up the bounds" below for why this is required)
 composite_basis["speed"].bounds = (-2, 2)
+trans_bas = (composite_basis).to_transformer()
+# concatenate the inputs
+inp = np.concatenate([counts, speed[:, np.newaxis]], axis=1)
+print(inp.shape)
+
+trans_bas.fit_transform(inp)
+
 ```
 
-### Defining the Input Shape
+#### The Solution: Defining the Input Shape
 
 :::{note}
 If each basis accepts a single column of the input, as in [this example](sklearn-how-to), then the default behavior works fine, and `set_input_shape` does not need to be called.
@@ -222,6 +200,50 @@ out2 = trans_bas.fit_transform(inp2)
 ```
 
 
-### Learn more
+### Setting up the `bounds`
+
+#### The Problem
+
+Any basis that derives its domain from `bounds` must have `bounds` set before it can be used as a transformer. This is a deliberate safety choice. With `bounds` left unset, the domain is inferred from the data, so during cross-validation each fold would re-fit the domain to its own data range. The transformation would then differ from fold to fold, which is rarely what you want when cross-validating.
+
+Calling `fit`, `transform`, or `fit_transform` on a `TransformerBasis` whose domain-defining `bounds` are unset raises a `RuntimeError`.
+
+```{code-cell} ipython3
+:tags: [raises-exception]
+
+# reinstantiate the basis transformer for illustration purposes
+composite_basis = counts_basis + speed_basis
+trans_bas = (composite_basis).to_transformer()
+# set the input shape but NOT the bounds
+trans_bas.set_input_shape(counts, speed)
+# concatenate the inputs
+inp = np.concatenate([counts, speed[:, np.newaxis]], axis=1)
+print(inp.shape)
+
+trans_bas.fit_transform(inp)
+
+```
+
+#### The Solution
+
+Noting that only `Eval` bases carry `bounds` while convolutional bases do not, we can fully set up the `TransformerBasis` by fixing the bounds for the `Eval` component (`speed`) of `composite_basis`.
+
+
+```{code-cell} ipython3
+
+# get the Eval basis using its label and set the bounds
+composite_basis["speed"].bounds = (-2, 2)
+
+# set the input shape
+composite_basis.set_input_shape(counts, speed)
+
+# convert to a transformer
+trans_bas = composite_basis.to_transformer()
+
+# transform the data
+X = trans_bas.fit_transform(inp)
+```
+
+## Learn more
 
 If you want to learn more about how to select basis' hyperparameters with `sklearn` pipelining and cross-validation, check out [this how-to guide](sklearn-how-to).
--- a/docs/how_to_guide/plot_05_transformer_basis.md
+++ b/docs/how_to_guide/plot_05_transformer_basis.md
@@ -205,7 +205,7 @@ out2 = trans_bas.fit_transform(inp2)
 
 #### The Problem
 
-Any basis that derives its domain from `bounds` must have `bounds` set before it can be used as a transformer. This is a deliberate safety choice. With `bounds` left unset, the domain is inferred from the data, so during cross-validation each fold would re-fit the domain to its own data range. The transformation would then differ from fold to fold, which is rarely what you want when cross-validating.
+Any basis that derives its domain from `bounds` must have `bounds` set explicitly before it can be used as a transformer. This is a deliberate safety choice. With `bounds` left unset, the domain is inferred from the data, so during cross-validation each fold would re-fit the domain to its own data range. The transformation would then differ from fold to fold, which is rarely what you want when cross-validating.
 
 Calling `fit`, `transform`, or `fit_transform` on a `TransformerBasis` whose domain-defining `bounds` are unset raises a `RuntimeError`.
 
--- a/docs/how_to_guide/plot_05_transformer_basis.md
+++ b/docs/how_to_guide/plot_05_transformer_basis.md
@@ -200,6 +200,7 @@ out2 = trans_bas.fit_transform(inp2)
 ```
 
 
+(transformer-basis-bounds)=
 ### Setting up the `bounds`
 
 #### The Problem
--- a/docs/how_to_guide/plot_05_transformer_basis.md
+++ b/docs/how_to_guide/plot_05_transformer_basis.md
@@ -227,7 +227,7 @@ trans_bas.fit_transform(inp)
 
 #### The Solution
 
-Noting that only `Eval` bases carry `bounds` while convolutional bases do not, we can fully set up the `TransformerBasis` by fixing the bounds for the `Eval` component (`speed`) of `composite_basis`.
+We can finish setting up the `TransformerBasis` by setting the bounds for the `speed` component (`speed`) of `composite_basis`. We do not need to do this for the `position` component because it is `Conv` basis object, and only `Eval` basis objects have `bounds`. 
 
 
 ```{code-cell} ipython3
--- a/docs/how_to_guide/plot_05_transformer_basis.md
+++ b/docs/how_to_guide/plot_05_transformer_basis.md
@@ -50,8 +50,8 @@ counts = np.random.poisson(size=(100, 5))
 speed = np.random.normal(size=(100))
 
 # create a composite basis
-counts_basis = nmo.basis.RaisedCosineLogConv(5, window_size=10)
-speed_basis = nmo.basis.BSplineEval(5)
+counts_basis = nmo.basis.RaisedCosineLogConv(5, label="counts", window_size=10)
+speed_basis = nmo.basis.BSplineEval(5, label="speed")
 composite_basis = counts_basis + speed_basis
 
 # compute the features
@@ -115,7 +115,7 @@ This mismatch leads to an error when calling fit_transform:
 ```{code-cell} ipython3
 :tags: [raises-exception]
 
-# reinstantiate the basis transformer for illustration porpuses
+# reinstantiate the basis transformer for illustration purposes
 composite_basis = counts_basis + speed_basis
 trans_bas = (composite_basis).to_transformer()
 # concatenate the inputs
@@ -124,6 +124,38 @@ print(inp.shape)
 
 trans_bas.fit_transform(inp)
 
+```
+
+There is a second requirement: any basis that derives its domain from `bounds` must have `bounds` set before it can be used as a transformer. This is a deliberate safety choice. With `bounds` left unset, the domain is inferred from the data, so during cross-validation each fold would re-fit the domain to its own data range. The transformation would then differ from fold to fold, which is rarely what you want when cross-validating.
+
+Let's see this with an example:
+
+```{code-cell} ipython3
+:tags: [raises-exception]
+
+# reinstantiate the basis transformer for illustration purposes
+composite_basis = counts_basis + speed_basis
+trans_bas = (composite_basis).to_transformer()
+# set the input shape but NOT the bounds
+trans_bas.set_input_shape(counts, speed)
+# concatenate the inputs
+inp = np.concatenate([counts, speed[:, np.newaxis]], axis=1)
+print(inp.shape)
+
+trans_bas.fit_transform(inp)
+
+```
+
+### Define the `bounds`
+
+As shown above, any basis whose domain is defined by `bounds` must have it set before the basis is used as a transformer. Only `Eval` bases carry `bounds`; convolutional bases do not.
+
+Let's set the bounds on the `Eval` component (`speed`) of `composite_basis`.
+
+```{code-cell} ipython3
+
+# get the Eval basis using its label and set the bounds
+composite_basis["speed"].bounds = (-2, 2)
 ```
 
 ### Defining the Input Shape
@@ -151,7 +183,7 @@ axis. For example, if the input is a 4D tensor, one needs to provide the last 3 
 x = np.random.randn(10, 3, 2, 1)
 
 # define and setup the basis
-basis = nmo.basis.BSplineEval(5).set_input_shape((3, 2, 1))
+basis = nmo.basis.BSplineEval(5, bounds=(-2, 2)).set_input_shape((3, 2, 1))
 
 X = basis.to_transformer().transform(
     x.reshape(10, -1)  # reshape to 2D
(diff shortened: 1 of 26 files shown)


docs/how_to_guide/plot_05_transformer_basis.md
```diff
@@ -126,6 +126,38 @@ trans_bas.fit_transform(inp)
 
 ```
 
+There is a second requirement: any basis that derives its domain from `bounds` must have `bounds` set before it can be used as a transformer. This is a deliberate safety choice. With `bounds` left unset, the domain is inferred from the data, so during cross-validation each fold would re-fit the domain to its own data range. The transformation would then differ from fold to fold, which is rarely what you want when cross-validating.
```
billbrod: I think this content is good, but the order is weird. I'd finish the top-level `Setting up the TransformerBasis` section with something like: "TransformerBasis has two additional requirements: input shape and bounds." and then have separate sub-headers about each of them, each of which has their own sub-headers with something like "The problem", "The solution."

docs/how_to_guide/plot_05_transformer_basis.md
```diff
@@ -126,6 +126,38 @@ trans_bas.fit_transform(inp)
...
+As shown above, any basis whose domain is defined by `bounds` must have it set before the basis is used as a transformer. Only `Eval` bases carry `bounds`; convolutional bases do not.
+
+Let's set the bounds on the `Eval` component (`speed`) of `composite_basis`.
+
+```{code-cell} ipython3
+
+# get the Eval basis using its label and set the bounds
+composite_basis["speed"].bounds = (-2, 2)
+```
+
```
billbrod: unrelated to this, but your "Learn more" header is at the wrong level -- it should be `##`, not `###`

docs/how_to_guide/plot_06_sklearn_pipeline_cv_demo.md
```diff
@@ -192,7 +192,8 @@ pipeline = Pipeline(
     [
         (
             "transformerbasis",
-            nmo.basis.RaisedCosineLinearEval(6).to_transformer(),
+            # bounds must be set to use an Eval basis as a transformer
```
billbrod: could have a little admonition above this that links to the other docs page with "see here for more info about bounds"

billbrod: should also say something like "we need to define bounds explicitly so it's consistent across folds"

docs/how_to_guide/variable_selection_zero_basis.md
```diff
@@ -96,8 +96,14 @@ from sklearn.pipeline import Pipeline
 import numpy as np
 
 # Define complete basis configurations
-position_basis = nmo.basis.BSplineEval(n_basis_funcs=10)
-theta_basis = nmo.basis.CyclicBSplineEval(n_basis_funcs=8)
+position_basis = nmo.basis.BSplineEval(
+    n_basis_funcs=10,
+    bounds=(position.min(), position.max()),
```
billbrod, line 101: why is this here?

src/nemos/basis/_basis.py
```diff
@@ -197,6 +197,11 @@ class Basis(Base, abc.ABC, BasisTransformerMixin):
         If ``axis`` different from 0 is provided as a keyword argument (samples must always be in the first axis).
     """
 
+    # Whether ``bounds=None`` makes the basis derive its domain from the input data
+    # (e.g. via ``min_max_rescale_samples``). Set to True on the rescaling base classes;
+    # the transformer safety gate requires explicit ``bounds`` for these.
+    _bounds_define_domain = False
```
billbrod: why is this here instead of in the Eval mixin? for the transformer safety check, you could just check whether this exists and is True, rather than whether this is True and whether bounds exists

src/nemos/basis/_basis_mixin.py
```diff
@@ -899,7 +943,7 @@ def to_transformer(self) -> TransformerBasis:
         >>> from sklearn.model_selection import GridSearchCV
         >>> # load some data
         >>> X, y = np.random.normal(size=(30, 1)), np.random.poisson(size=30)
-        >>> basis = nmo.basis.RaisedCosineLinearEval(10).set_input_shape(1).to_transformer()
+        >>> basis = nmo.basis.RaisedCosineLinearEval(10, bounds=(X.min(), X.max())).set_input_shape(1).to_transformer()
```
billbrod, line 968: probably worth saying something explicitly about bounds in this docstring, like I said in the top-level comment.

src/nemos/basis/_custom_basis.py
```diff
@@ -367,6 +367,17 @@ def compute_features(
         :
             The resulting design matrix, with one row per sample and one column per output feature.
 
+        Notes
+        -----
+        ``CustomBasis`` does not derive any state from the data: the functions in ``self.funcs`` are
```
billbrod: I think this sentence is opaque for many users, and the paragraph itself is tricky. maybe something like:

> `CustomBasis` does not change as a result of its inputs: the functions in `self.funcs` are applied to the raw input, and `bounds` only controls which samples are replaced by `fill_value` when they fall outside the given interval. The transformation is therefore identical across inputs *as long as your functions are sample-wise*. If a function's behavior depends on a data-dependent quantity (e.g. normalizing by the input mean or std), the transformation will differ from one input to the next, which is unsafe in some contexts, e.g. inside a cross-validation loop. Instead, provide functions whose output depends only on each sample, or fix any data-dependent quantity to obtain an identical transformation across all possible inputs.

I'm actually not sure what you mean by "as long as your functions are sample-wise".

billbrod: but I think this point is too important to be lost in a notes in `compute_features`

src/nemos/basis/_transformer_basis.py
```diff
@@ -58,7 +59,7 @@ class TransformerBasis:
     >>> # Generate data
     >>> num_samples, num_features = 10000, 1
     >>> x = np.random.normal(size=(num_samples, ))  # raw time series
-    >>> basis = BSplineEval(10)
+    >>> basis = BSplineEval(10, bounds=(x.min(), x.max()))
```
billbrod, line 68: should explain this in docstring

src/nemos/basis/_transformer_basis.py
```diff
@@ -134,6 +135,28 @@ def _check_initialized(basis):
...
+            # ``hasattr`` check excludes convolutional bases, which inherit the flag from the
+            # shared math base class but carry no ``bounds`` (their ``__getattr__`` raises).
+            # ``is_leaf=lambda x: x is None`` keeps ``None`` as a leaf, so this catches
+            # both ``bounds=None`` and a partially-unset ``(None, high)``.
+            if (
+                getattr(b, "_bounds_define_domain", False)
+                and hasattr(b, "bounds")
+                and any(
+                    v is None
+                    for v in jax.tree.leaves(b.bounds, is_leaf=lambda x: x is None)
```
billbrod: `is_leaf` here is excluding the Additive and Multiplicative basis objects?

billbrod: if so, say that

tests/test_transformer_basis.py
```diff
@@ -1304,3 +1320,25 @@ def test_double_transformer():
     assert (
         id(tbas.basis) != id(tbas2.basis) != id(tbas3)
     ), "The basis was shallow copied!"
+
+
+@pytest.mark.parametrize(
+    "bas, expectation",
+    [
+        (
```
billbrod: I know it's annoying, but is it worth iterating through all of nemos's basis objects as part of this test? the individual test is pretty quick yeah?

tests/test_transformer_basis.py
```diff
@@ -1304,3 +1320,25 @@ def test_double_transformer():
...
+            nmo.basis.BSplineEval(5),
+            pytest.raises(RuntimeError, match="BSplineEval.+bounds"),
+        ),
+        (nmo.basis.BSplineEval(5, bounds=(0, 1)), does_not_raise()),
+        (nmo.basis.BSplineConv(5, 10), does_not_raise()),
+        (nmo.basis.IdentityEval(), does_not_raise()),
+        (nmo.basis.Zero(), does_not_raise()),
+        (nmo.basis.CustomBasis(lambda x: x), does_not_raise()),
```
billbrod: should probably also test a CustomBasis that fails

billbrod (changes requested):
Good catch on this. Are there cases where someone would use the standard basis (not `TransformerBasis`) when they might run into this? Should we raise a warning if `bounds=None` to let them know the transformation is data-dependent? That might be too many warnings... The `Eval` docstring has this information in a note, but I worry that might be easy to skip. We could turn it into an attention admonition in the main body of the docstring instead?

- Can we set bounds at initialization of `TransformerBasis`? I think including it in that docstring more clearly signals to the user that it's something they have to think about it (even though we can't make it necessary because of the basis objects that don't require it)
    - also, this docstring should talk more explicitly about bounds somewhere

Your questions:
- I think the runtime location is correct. If you add `bounds` to initialization like I noted above, that hopefully signals it clearer to the user. How much time is spent on the call?
- CustomBasis: 
    - CustomBasis is kind of an advanced usage, but this is an easy mistake to make. I'd say raise a warning on initialization if `bounds is None`, pointing people to docstring.
    - bounds is currently unexplained in the docstring, so I'd add more of an explanation there, including either the note or attention admonition to explain it in more detail.
    - I'd add a doctest that tries to make this point as well.

docs/how_to_guide/plot_05_transformer_basis.md
BalzaniEdoardo: I did reorganize the content, let me know if it matches what you are describing

docs/how_to_guide/variable_selection_zero_basis.md
BalzaniEdoardo: it is used as a transformer in a Pipeline

src/nemos/basis/_basis.py
BalzaniEdoardo: I think you are right, I'll change that

src/nemos/basis/_custom_basis.py
BalzaniEdoardo: I did add a better explanation of how the bounds work for this basis. I also made the sample-wise concept clear: it means that the funcs output doesn't depend on the sample index. It's better explained (I believe) in the docstrings, tell me if it is clear enough. In case it's not clear enough, can you help me rephrase it, what it literally mean is that:

```python
>>> func(x[i]) == func(x)[i]
```

src/nemos/basis/_transformer_basis.py
BalzaniEdoardo: I clarified. the loop is on all the atomic components, but the FourierEval is both atomic and potentially n-dim. So the bounds for it can be a list of bounds one per dimension. We need to make sure that all bounds are set.

tests/test_transformer_basis.py
BalzaniEdoardo: Fair, I did not want to do a tautological test using the attribute `_bounds_define_domain` since it has a default behavior for any new Eval basis. 

I made sure that i hardcode the expectation for how bounds are treated in each implemented basis, and checked that all basis appear in the hard-coded list. This way we will be forced to mark the expected bounds behavior in each new basis.

BalzaniEdoardo (commented):
I responded to most of the comments

BalzaniEdoardo:
> Good catch on this. Are there cases where someone would use the standard basis (not `TransformerBasis`) when they might run into this? Should we raise a warning if `bounds=None` to let them know the transformation is data-dependent? That might be too many warnings... The `Eval` docstring has this information in a note, but I worry that might be easy to skip. We could turn it into an attention admonition in the main body of the docstring instead?

One can do cross-validation with a loop. But that's the same risk you would incur if you use `scipy` zscore function. I think that's up to the user to be mindful of what operation is being applied.

>     * Can we set bounds at initialization of `TransformerBasis`? I think including it in that docstring more clearly signals to the user that it's something they have to think about it (even though we can't make it necessary because of the basis objects that don't require it)

Not yet, currently AdditiveBasis do not have a bounds setter for example, and the reason is that it is not trivial for a user to pass the right bounds list, given that some basis do not have the property (`Conv` basis for example). 

>       * also, this docstring should talk more explicitly about bounds somewhere

This has been addressed.

> Your questions:
> 
>     * I think the runtime location is correct. If you add `bounds` to initialization like I noted above, that hopefully signals it clearer to the user. How much time is spent on the call?
> 
>     * CustomBasis:
>       
>       * CustomBasis is kind of an advanced usage, but this is an easy mistake to make. I'd say raise a warning on initialization if `bounds is None`, pointing people to docstring.

This could be a misconception, the bounds of CustomBasis do not rescale, so they won't determine the domain. The behavior of the transformation (if it is a fixed map or if it a data dependent one) is out of our control. It depends on what the user passes as funcs. 

>       * bounds is currently unexplained in the docstring, so I'd add more of an explanation there, including either the note or attention admonition to explain it in more detail.

I added an explanation.

>       * I'd add a doctest that tries to make this point as well.

Added too

docs/how_to_guide/plot_05_transformer_basis.md
```diff
@@ -99,45 +99,35 @@ As with any `sckit-learn` transformer, the `TransformerBasis` implements `fit`,
...
+- `counts` consists of **5 columns**
 
-trans_bas.fit_transform(inp)
+- `speed` consists of **1 column**
 
-```
+When concatenated, this results in a **6-column input**, and our `TransformerBasis` doesn't know how to split those 6 columns.
 
-There is a second requirement: any basis that derives its domain from `bounds` must have `bounds` set before it can be used as a transformer. This is a deliberate safety choice. With `bounds` left unset, the domain is inferred from the data, so during cross-validation each fold would re-fit the domain to its own data range. The transformation would then differ from fold to fold, which is rarely what you want when cross-validating.
+This mismatch leads to an error when calling fit_transform (the cell also sets `bounds`, the second requirement, so that the failure below is about the input shape only; bounds are covered [in the next section](#setting-up-the-bounds)):
```
billbrod:
```suggestion
This mismatch leads to an error when calling fit_transform:
```

too much detail I think. the comment in the code block is sufficient.

docs/how_to_guide/plot_05_transformer_basis.md
```diff
@@ -146,19 +136,7 @@ trans_bas.fit_transform(inp)
...
-
-# get the Eval basis using its label and set the bounds
-composite_basis["speed"].bounds = (-2, 2)
-```
-
-### Defining the Input Shape
+#### The Solution: Defining the Input Shape
 
 :::{note}
 If each basis accepts a single column of the input, as in [this example](sklearn-how-to), then the default behavior works fine, and `set_input_shape` does not need to be called.
```
billbrod: this note should probably be moved to the beginning of the `Setting the input shape` section

docs/how_to_guide/plot_05_transformer_basis.md
```diff
@@ -222,6 +200,51 @@ out2 = trans_bas.fit_transform(inp2)
 ```
 
 
-### Learn more
+(transformer-basis-bounds)=
+### Setting up the `bounds`
+
+#### The Problem
+
+Any basis that derives its domain from `bounds` must have `bounds` set before it can be used as a transformer. This is a deliberate safety choice. With `bounds` left unset, the domain is inferred from the data, so during cross-validation each fold would re-fit the domain to its own data range. The transformation would then differ from fold to fold, which is rarely what you want when cross-validating.
```
billbrod:
```suggestion
Any basis that derives its domain from `bounds` must have `bounds` set explicitly before it can be used as a transformer. This is a deliberate safety choice. With `bounds` left unset, the domain is inferred from the data, so during cross-validation each fold would re-fit the domain to its own data range. The transformation would then differ from fold to fold, which is rarely what you want when cross-validating.
```

docs/how_to_guide/plot_05_transformer_basis.md
```diff
@@ -222,6 +200,51 @@ out2 = trans_bas.fit_transform(inp2)
...
+inp = np.concatenate([counts, speed[:, np.newaxis]], axis=1)
+print(inp.shape)
+
+trans_bas.fit_transform(inp)
+
+```
+
+#### The Solution
+
+Noting that only `Eval` bases carry `bounds` while convolutional bases do not, we can fully set up the `TransformerBasis` by fixing the bounds for the `Eval` component (`speed`) of `composite_basis`.
```
billbrod:
```suggestion
We can finish setting up the `TransformerBasis` by setting the bounds for the `speed` component (`speed`) of `composite_basis`. We do not need to do this for the `position` component because it is `Conv` basis object, and only `Eval` basis objects have `bounds`. 
```

or something like that, just reverse the order of the two points here.

docs/how_to_guide/plot_05_transformer_basis.md
```diff
@@ -99,45 +99,35 @@ As with any `sckit-learn` transformer, the `TransformerBasis` implements `fit`,
 
 ## Setting up the TransformerBasis
 
-At this point, we have an object equipped with the necessary methods. So, all we need to do is concatenate the inputs into a single array and call fit_transform, right? **Not quite.** By default, `TransformerBasis` assumes that `inp` has one column per input in the composition. So in this example, with two one-dimensional basis functions, `TransformerBasis` expects an input of shape `(n_samples, 2)`.
+At this point, we have an object equipped with the necessary methods. So, all we need to do is concatenate the inputs into a single array and call fit_transform, right? **Not quite.**
 
-However, our actual input has shape `(n_samples, 6)`:
+The `TransformerBasis` has two additional requirements: input shape and bounds.
```
billbrod:
```suggestion
The `TransformerBasis` has two additional requirements: input shape (all basis objects) and bounds (`Eval` only).
```

or something to make this clear at the beginning of this section

src/nemos/basis/_basis_mixin.py
```diff
@@ -934,6 +950,12 @@ def to_transformer(self) -> TransformerBasis:
         """
         Turn the Basis into a TransformerBasis for use with scikit-learn.
```
billbrod, line 951: is it possible for this method to require a `bounds` argument? to really signal that this is necessary.

src/nemos/basis/_custom_basis.py
```diff
@@ -163,6 +163,29 @@ class CustomBasis(BasisMixin, BasisTransformerMixin, Base):
...
+
+    .. warning::
+
+        Functions that normalize by a quantity computed from the input are unsafe here, because that
+        quantity — and so the transformation — changes from one input to the next, whereas across
+        cross-validation folds it must stay identical. The fix is to precompute any such quantity once
+        (e.g. on the training set) and pass it in as a constant, so the output depends only on each sample.
+        For the z-scoring example, this means precomputing the centering and scaling instead of recomputing
+        them from each input: replace ``lambda x: (x - x.mean()) / x.std()`` with
+        ``lambda x: (x - mean) / std``, where ``mean`` and ``std`` are fixed.
```
billbrod:
```suggestion
        cross-validation folds it must stay identical. 
        
        Thus, you should precompute any such quantity once (e.g. on the training set) and pass it in
        as a constant, so the output depends only on each sample. When z-scoring, for example, this 
        means precomputing the centering and scaling instead of recomputing them from each input:
        replace ``lambda x: (x - x.mean()) / x.std()`` with ``lambda x: (x - mean) / std``, where ``mean``
        and ``std`` are fixed. See Examples below for an example of this.
```

billbrod (changes requested):
Couple minor changes here. I was misunderstanding how CustomBasis worked, the changes to the docstrings make it much clearer.

I do think it would be nice to make it more obvious to set `bounds` for `TransformerBasis` -- as I mentioned in a comment, is it possible to have it be a required argument for `to_transformer()` of `Eval` basis objects? Or  does the AdditiveBasis issue still apply?

src/nemos/basis/_basis_mixin.py
BalzaniEdoardo: Not really, the additive basis has no bounds setter, so exposing it would be problematic. In a composite basis, one would need to remember which basis in the composition has bounds and which one hasn't etc.

billbrod (approved):
Looks great! I still wish there was an easier way to make `bounds` required (when it is), but hopefully these changes simplify things.

PR #587, opened 2026-07-08
Add a fit intercept parameter

BalzaniEdoardo:
## Summary

Some models, including the Fourier approximate GP regression we plan to implement, are more easily implemented if the intercept term is included as part of the design matrix and regularized appropriately. To allow this behavior, we introduced a new parameter for the GLM: a boolean flag `fit_intercept` that by default is True (usual behavior) and when False it pins the intercept to zero. 

The implementation is done by marking which parameter sub-tree must be kept frozen + an initialization that enforces the intercept to be frozen and pinned to zero. 

The frozen parameter machinery is implemented at the level of the `BaseRegressor` and will be hopefully re-usable to allow partial parameter specifications in the future. 

A new weights initialization procedure is introduced for the `fit_intercept=False` case. Specifically, since the intercept is pinned to 0, the coefficients must be set to somewhat match the mean firing rate of the neuron(s). This limits the risk of nans, or infs in the log-likelihood which may occur when the predicted rate is very far off the observed one.  

The initialization chosen is to set the weights to a constant $c$ minimizing,

$$
c^* = \text{argmin}_{c \in \mathbb{R}} \sum_t (s_t \cdot c - f^{-1}(\bar{y}))^2 ,
$$

where $s_t = \sum_i X_{ti}$, the sum over rows of the design matrix $X$, $\bar{y} = \frac{1}{T} \sum_t y_t$ is the mean firing rate per bin and $f$ is the non-linearity. This guarantees that the predicted rate at the initial parameters $f(X_t \cdot w_0) = f(s_t \cdot c^*)$ approximates the mean rate of the neuron(s).

## Motivation

Not including the intercept term enables regularizing over all the GLM parameters (the intercept is unregularized currently). This feature is new and enables some interesting model specifications. For example, one can perform approximate GP regression with NeMoS with an appropriately defined design matrix, but this model instance requires that all parameters are ridge regularized. 

The more limited implementation would have been to pin the intercept to zero at the GLM level. However we opted for a solution that is more aligned with the package future plans: we introduced additional `BaseRegressor` machinery that marks frozen and active parameters. The frozen parameters are pinned, the active will be learned. This machinery works for the simple case of pinning the intercept to zero, which is all we need here, but should be reusable more generally. 

Ideally, in the future, it will enable to partition a parameter tree into an active and a frozen sub-tree, and learn just what one needs. #485 is an example of a use case, freezing the intercept to a specific value and learning the coefficients.

### `BaseRegressor`

**Class attributes**

- `_active_params`: a tree of booleans with the same structure of the model params  or `None`.  This marks which parameters should be learned, and it is used to split the parameter tree via [`equinox.partition`](https://docs.kidger.site/equinox/api/manipulation/#equinox.partition).
- `_frozen_params`: a tree of arrays or `None` matching the model parameter structure. This includes the parameter that are pinned to a specific value. It is required for iterative fitting loops using `update`; at each iteration we need a reference to the pinned parameter value.

**Methods**

- **New** `_partition_active` method that uses `_active_params` to split the parameter tree in frozen and active.
- **New** `_run_optimizer` method (used by`fit` and `stochastic_fit`). This is a thin wrapper of the solver run method which partitions, initializes, run the solver and recombines the parameters in a single call.
- **New** `_normalize_user_params` method. A hook necessary to fill in the zero intercept as a frozen parameter. By default it does nothing.
- **Updated** `_instantiate_solver`. This method receives an optional `frozen_params` parameter. When this is provided, the loss function is wrapped with a tree recombination step which forms the full parameter tree from the initial active and the frozen parameters.
- **Updated** `initialize_optimizer_and_state`. This public method is invoked before an update loop to initialize the solver and the state. The method was updated to partition the initial parameters, and cache the frozen parameter for the update loop.

### `glm` Module 

- **New** `initialize_constant_coef_matching_mean_rate`. This performs the initialization described in the summary. The formula for the optimal c is analytical and fast to compute. Additionally, a machine precision correction is added to take care of the case of zeros design matrix, preventing `c=inf` or `c=nan`
- **Updated** `_resolve_default_solver` which for now, sets `"LBFGS"` for ridge regression and no intercept. This is only temporary because currently the Hessian is not via autodiff. @wulfdewolf is working on a full autodiff path in another branch. After that is done we can drop this extra bit of machinery.
- **Updated** `_model_specific_initialization`. This method now calls different initialization paths depending on `fit_intercept`. 
- **New** `_normalize_user_params` overrides. This method takes care of setting the frozen parameters for the intercept to a zeros. If the user provided initial parameters that include the intercept and the model is set to `fit_intercept=False`, the method warns the user that the intercept is not being fit and will be pinned to zero.
- **Updated** `fit` method. It now relies on `_run_optimizer` to take care of the partial parameter specification machinery at the base regressor level.
- **Updated** `stochastic_fit`: wrap the training context (used by callbacks monitoring the fit convergence etc) in a function that stores the frozen parameters a context attribute, then runs the stochastic optimization. This wrapping is needed because the frozen parameter are populated as part the `_run_optimizer` call, and are not available at context creation. 
- **Updated** `_estimate_resid_degrees_of_freedom` count the dof based on the active parameters only.
- **Updated** `_initialize_optimizer_and_state`. This method now routes down the frozen parameter to `_instantiate_solver`. 
- **Updated** `update`. The method now get the set of active parameter and uses the cached frozen parameters for computing optimization updates.

### TrainingContext

This context now needs to store the frozen parameters. The user is not exposed to frozen and active params in training,  the parameters are recombined as a property. In order to define the property, I dropped the `dataclass` and used a standard `__init__` method. Additionally, I added a simple repr that mimicked that of a `dataclass`.

### DIFF
--- a/src/nemos/glm/initialize_parameters.py
+++ b/src/nemos/glm/initialize_parameters.py
@@ -177,11 +177,11 @@ def initialize_intercept_matching_mean_rate(
 
 def initialize_constant_coef_matching_mean_rate(
     inverse_link_function: Callable,
-    X: Union[dict, jnp.ndarray],
+    X: Union[Pytree, jnp.ndarray],
     y: jnp.ndarray,
     empty_coef: Union[Pytree, jnp.ndarray],
     eps: Optional[float] = None,
-) -> Union[dict, jnp.ndarray]:
+) -> Union[Pytree, jnp.ndarray]:
     r"""
     Initialize coefficients as a constant matching the mean rate, with no intercept.
 
@@ -255,4 +255,6 @@ def initialize_constant_coef_matching_mean_rate(
     scale = (jnp.sum(row_sum) + eps) / (jnp.sum(row_sum**2) + eps)
     const = eta_target * scale  # (n_out,)
 
-    return jax.tree_util.tree_map(lambda leaf: jnp.ones(leaf.shape) * const, empty_coef)
+    return jax.tree_util.tree_map(
+        lambda leaf: (jnp.ones_like(leaf) * const).astype(leaf.dtype), empty_coef
+    )
--- a/src/nemos/glm/initialize_parameters.py
+++ b/src/nemos/glm/initialize_parameters.py
@@ -255,6 +255,4 @@ def initialize_constant_coef_matching_mean_rate(
     scale = (jnp.sum(row_sum) + eps) / (jnp.sum(row_sum**2) + eps)
     const = eta_target * scale  # (n_out,)
 
-    return jax.tree_util.tree_map(
-        lambda leaf: (jnp.ones_like(leaf) * const).astype(leaf.dtype), empty_coef
-    )
+    return jax.tree_util.tree_map(lambda leaf: jnp.full_like(leaf, const), empty_coef)
--- a/src/nemos/glm/initialize_parameters.py
+++ b/src/nemos/glm/initialize_parameters.py
@@ -1,6 +1,6 @@
 """Initialization of GLM parameters."""
 
-from typing import Callable
+from typing import Callable, Optional, Union
 
 import jax
 import jax.numpy as jnp
@@ -15,6 +15,7 @@ from ..inverse_link_function_utils import (
     norm_cdf,
     softplus,
 )
+from ..typing import Pytree
 from ..utils import one_over_x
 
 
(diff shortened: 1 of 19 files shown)


src/nemos/glm/initialize_parameters.py
```diff
@@ -172,3 +173,86 @@ def func(x, mean_x):
         raise non_finite_error
 
     return out
+
+
+def initialize_constant_coef_matching_mean_rate(
```
wulfdewolf, line 178: [crazy that `sklearn` just initialises to 0](https://github.com/scikit-learn/scikit-learn/blob/cc50648cc1b759b53a4edbce0f3bb6c237349448/sklearn/linear_model/_glm/glm.py#L266)

tests/conftest.py
```diff
@@ -102,6 +104,39 @@ def all_subclasses(cls):
     return seen
 
 
+@pytest.fixture
+def setup_solver():
```
wulfdewolf, line 108: I think many of the regularizer tests could use this as well

tests/test_glm.py
```diff
@@ -1631,6 +1632,200 @@ class _TwoLeafModule(eqx.Module):
...
+    def test_malformed_structure_raises(
+        self, poissonGLM_model_instantiation, n_elements
+    ):
+        """A user-params tuple whose length is not two is rejected."""
+        X, y, model, true_params, _ = poissonGLM_model_instantiation
+        model.fit_intercept = False
+        bad_params = (true_params.coef,) * n_elements
+        with pytest.raises(ValueError, match="length two"):
+            model._normalize_user_params(bad_params, X, y)
+
```
wulfdewolf, line 1684: Maybe a test that verifies that the warning is thrown if the user passes an initial value but `fit_intercept=False`?
And generally a test that the behaviour in that case is correct

wulfdewolf (commented):
This looks very clean to me.
Choosing for the abstracted interface with frozen and active parameters at the `BaseRegressor` level seems optimal imo, especially with more models planned in the future.

A minor concern for further down the line is that already some edge cases are introduced:
- the GLM-HMM doesn't allow parameter freezing, which adds a bunch of pass-through/untouched parameters
- the Newton solver doesn't work with a frozen intercept, which also adds a bunch of checks and tests just to ensure things don't break

I don't think this is an issue as of now, but with more and more features added it might become painful.

The PR summary says we can remove forcing LBFGS over Newton, but the code seems to have kept this?
Since the Newton solver is now fully autodiff, can we remove this code?

Right now the interaction with regularization strength is that if you want to pass parameter-wise strengths, you include the fixed parameters. For the GLM that is fine since it is only the intercept, but for other models this might be something to think about.

I further think some (less important) tests could be added to check behaviour and warnings when the user passes an intercept initial value but `fit_intercept=False` (see comment).

src/nemos/glm/initialize_parameters.py
BalzaniEdoardo: yep, sometimes it can result in nan or infs in the loglikelihood

wulfdewolf (approved):
LGTM!

PR #588, opened 2026-07-09
Fft convolve

BalzaniEdoardo:
This PR adds FFT convolution capability to the `convolve` module of nemos.

`convolve.create_convolutional_predictor` gains a new parameter `use_fft`, default `None`. `True`/`False` force the backend; `None` resolves it per input array with the following decision tree:

1. **Device is not CPU** (e.g. GPU/TPU) → use the direct convolution, letting XLA select and fuse device-optimized primitives. Our heuristic is a CPU cost model and the FFT-vs-direct crossover on an accelerator is hardware-dependent, so we do not override the compiler's own algorithm selection. (Under `jit`, an abstract value has no concrete device and takes this branch too, which keeps `use_fft=None` jit-safe.)
2. **Device is CPU** → compare the kernel length against the convolution *block size*: choose FFT when `window_size > FACTOR * log2(block_size)` (`FACTOR = 3.0`, conservative and tunable), otherwise direct. `block_size` is `batch_size_samples` — the length each FFT actually operates on — which equals the sample-axis length only when there is no sample batching.

When FFT is selected, the sample axis is convolved either as a single FFT (no batching) or, when `batch_size_samples < n_samples`, via overlap-save in blocks of `batch_size_samples`. Output matches the direct convolution up to floating-point round-off, and every convolutional basis inherits the option for free (it is a passthrough `conv_kwargs`).

### Backend

| Function | Role |
|---|---|
| `_CORR_VEC_FFT` | FFT primitive: `jax.scipy.signal.fftconvolve(mode="valid")` vmapped with the *same* axis layout as `_CORR_VEC`, so the batching helpers route to either transparently. The two-level vmap reduces each call to 1D operands, so `fftconvolve` does a strictly 1D convolution over the sample axis (it convolves over all axes on n-d input — the vmap is what keeps it 1D). |
| `_fft_convolve` | Single full-length FFT over the whole sample axis. Batches over channels/basis, not samples. |
| `_fft_overlap_save_convolve` | Overlap-save: blocks the sample axis into chunks of `batch_size_samples` overlapping by `window_size − 1`; each block's valid FFT convolution yields `batch_size_samples − window_size + 1` outputs that tile the result. Its pad/step/scan blocking mirrors `_tensor_convolve` exactly, so output is identical to the direct path up to round-off — only the per-block primitive differs. |

`jax.scipy.signal.fftconvolve` was chosen over a hand-rolled `rfft`/`irfft`: it supports `mode="valid"`, auto-swaps when the kernel is shorter than the signal, and is better tested.

### DIFF
--- a/src/nemos/convolve.py
+++ b/src/nemos/convolve.py
@@ -39,8 +39,8 @@ _CORR_VEC_FFT = jax.vmap(partial(fftconvolve, mode="valid"), (1, None), 1)
 _CORR_VEC_FFT = jax.vmap(_CORR_VEC_FFT, (None, 1), 2)
 
 # FFT-based convolution is asymptotically cheaper once the kernel length grows large
-# relative to log2(num_samples). This factor sets that crossover for the CPU heuristic;
-# it is deliberately conservative and tunable.
+# relative to log2 of the convolution block size. This factor sets that crossover for the
+# CPU heuristic; it is deliberately conservative and tunable.
 _FFT_WINDOW_LOG_FACTOR = 3.0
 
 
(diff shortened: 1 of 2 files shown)


BalzaniEdoardo:
@ahwillia and @pgree, FYI, with this PR, the `convolve` module is optimized.

pgree:
looks great. one pedantic comment -- the heatmap would be easier to read if it were powers of 10 instead of powers of 2

PR #590, opened 2026-07-11
Approximate LOO-CV

ahwillia:
I'd like to propose that we incorporate support for fast, approximate, leave-one-out cross-validation by leveraging the method in [this paper (Rad & Maleki, 2020)](https://doi.org/10.1111/rssb.12374). This extends a well-known trick to rapidly compute leave-one-out predictions in purely linear models (see e.g. [this blog post](https://robjhyndman.com/hyndsight/loocv-linear-models/))

I found this pretty easy to implement and with the help of Claude I cleaned it up with tutorials + documentation. One potential concern I have is that a user could input a model with a very large number of parameters and the code could take a long time to run. Potentially we could issue a warning. Relatedly, I'm interested what happens if we stress test this code with a large PopulationGLM.

Let me know what you think.


ahwillia:
I ran some more tests and realized that accuracy was poor for non-canonical links (in particular softplus vs exp for Poisson). I'll work on a re-implementation that properly accounts for the link function.

Issue #571, opened 2026-05-30
Vectorize the block-diagonal Newton Hessian via a declared `batched_dims` contract


BalzaniEdoardo:
## Context

The Newton solver builds its Hessian either from the analytic Fisher block (`_get_hess_fn`) or from autodiff (`force_autodiff_hessian`, or models that have no analytic formula). For population models the true Hessian is block-diagonal — one independent block per neuron — but neither path exploits that efficiently:

- **Autodiff path** flattens the parameters and calls `jax.hessian` on the flat loss, producing the full dense `(D, D)` matrix with `D = n_features * n_neurons + n_neurons`. Because that matrix is 2-D, the block-diagonal branch of `_newton_direction` (`tag.structure is BlockDiagonal and H.ndim == 3`) is skipped and the step is computed with a dense `O(D^3)` factorization.
- **Analytic path** (`PopulationGLM._get_hess_fn`, `ClassifierPopulationGLM._get_hess_fn`) builds the blocks with a Python `for neuron_idx in range(n_neurons)` loop followed by `jnp.stack`.

The dimension that separates blocks is currently inferred by `ravel_pytree_nest` from the trailing shape of the leaves. That heuristic can misfire when unrelated dimensions coincide — `n_features == n_neurons`, or a single-feature/single-class model where `coef` and `intercept` both end in a `1`.

## Proposal

1. Add an explicit contract for the batched dimensions — e.g. a `batched_dims` attribute on the parameter class, allowing more than one axis — that fully specifies which axes index independent blocks, rather than guessing from shapes.
2. With that contract, express both Hessian paths as a `vmap` over the batched axes:
   - autodiff: `jax.vmap(jax.hessian(per_block_loss), in_axes=batched_dims)`
   - analytic: `jax.vmap(_glm_hessian_block, in_axes=...)`, replacing the Python loop

   Both emit a `(B..., d, d)` block stack that feeds the existing `H.ndim == 3` branch of `_newton_direction`, scales linearly in the number of blocks, and never materializes the full `(D, D)`.

## Scope

- Parameter classes declare their batched dims; flattening/dispatch use the declared axes instead of the trailing-dim heuristic in `ravel_pytree_nest`.
- The autodiff Hessian for population models returns `(n_neurons, d, d)` and takes the block-diagonal solve path.
- The Python loops in `_get_hess_fn` are replaced by `vmap`.
[...]

PR #592, opened 2026-07-20
Autodiff newton with tree support

wulfdewolf:
## Summary
This PR removes analytic hessian computation for the Newton solver, moving to fully rely on autodiff.
In the population models we do not generate the full autodiff hessian, but we vmap over subproblems.
Everything works in pytree space.

## Highlights
The `HessianTag` class now has a `batch_axes` argument:
```
@dataclass(frozen=True)
class HessianTag:
    structure: type
    property: type
    batch_axes: Any = None
```

This allows us to specify batch axes for population models, such that the Newton solver knows where to vmap:
```
    def _solve(self, hess, grad):
        operator = lx.PyTreeLinearOperator(
            hess,
            jax.eval_shape(lambda: grad),
            tags=self._operator_tags,
        )

        return lx.linear_solve(
            operator,
            jax.tree.map(lambda x: -x, grad),
            self._linear_solver,
        ).value

    def _newton_direction(self, grad, H):
        if self._hess_tag.structure is BlockDiagonal:
            return jax.vmap(
                self._solve,
                in_axes=(0, self._hess_tag.batch_axes),
                out_axes=self._hess_tag.batch_axes,
            )(H, grad)
        else:
            return self._solve(H, grad)
```

We use `lineax.PyTreeOperator` to stay in tree space everywhere.

All analytic code has been removed. 
We only need to set the right hessian tag and pass it to the Newton solver, and it knows what to do.

In the `PopulationGLM` we do have a specific hessian function:
```
    def _get_hess_fn(self):
        def per_neuron(params, X, y, mask):
            def loss(params):
                rate = self._predict(params, X, feature_mask=mask)
                return self._observation_model._negative_log_likelihood(y, rate)

            return jax.hessian(loss)(params)

        return lambda params, X, y: jax.vmap(
            per_neuron,
            in_axes=(
                self._hess_tag.batch_axes,
                None,
                1,
                1,
            ),
        )(params, X, y, self._feature_mask)
```

Which allows us to keep model specific Hessian stuff independent from `Newton`.

The above would give an unpenalized hessian, hence we have similar functionality in the regularizer class:
```
    def get_hess_fn(self, params, strength, batch_axes=None) -> Callable | None:
        if self._hess_tag is None:
            return None

        filter_kwargs = self._get_filter_kwargs(strength=strength, params=params)

        def hessian(p, kwargs):
            return jax.hessian(self._penalization)(p, kwargs)

        if batch_axes is None:
            return lambda p: hessian(p, filter_kwargs)

        kwargs_axes = self._filter_kwargs_batch_axes(params, filter_kwargs, batch_axes)
        batched = jax.vmap(hessian, in_axes=(batch_axes, kwargs_axes))
        return lambda p: batched(p, filter_kwargs)
```

The base regressor then combines both penalties, if needed:

```
tree_utils.tree_add(model_hess_fn(params, *args), regularizer_hess_fn(params))
```

We then pass the resulting combined hessian function to the Newton solver.

## Testing
- Removed all tests relating to analytic v. autodiff
- Added thorough convergence testing
- Added tests to verify the hessian tags are correct
- Added tests to verify if block hessians are the same as full autodiff hessians

## Related
#571


src/nemos/solvers/_hess.py
```diff
@@ -128,4 +130,5 @@ def combine_hessian_tags(
     return HessianTag(
         structure=combine_structure(t1.structure, t2.structure),
         property=prop,
+        batch_axes=t1.batch_axes,
```
BalzaniEdoardo, line 135: Add to docstrings that the batching is determined by the first tag, and therefore it is important to combine the reg and the model tag in the right order (model hessian tag first)

src/nemos/solvers/_newton.py
```diff
@@ -147,46 +135,57 @@ def init_state(self, init_params, *args):
...
-        # init_state; reuse them rather than re-inferring per iteration.
-        operator = lx.MatrixLinearOperator(H, self._operator_tags)
-        return lx.linear_solve(operator, -g_flat, self._linear_solver).value
+        return lx.linear_solve(
+            operator,
+            jax.tree.map(lambda x: -x, grad),
+            self._linear_solver,
+        ).value
+
+    def _newton_direction(self, grad, H):
```
BalzaniEdoardo, line 190: this pr is great, so i do the extreme neatpeaks: this signature is the reverse of that of `_solve`, I would match the two

src/nemos/solvers/_newton.py
```diff
@@ -147,46 +135,57 @@ def init_state(self, init_params, *args):
...
+        descent = (
+            sum(
+                jnp.vdot(g, s)
+                for g, s in zip(
+                    jax.tree.leaves(grad),
+                    jax.tree.leaves(step),
+                )
+            )
+            < 0
+        )
```
BalzaniEdoardo: can you do 
```python
jax.tree.reduce(jnp.add, jax.tree.map(jnp.vdot, grad, step))
```
instead of the loop?

Even simpler, we can rely on lineax (I know it because that's what optimistix does

```
from lineax.internal import tree_dot
tree_dot(grad, step)
```

src/nemos/solvers/_newton.py
```diff
@@ -211,28 +210,23 @@ def update(
...
-        params_flat, _ = ravel_pytree_nest(params)
-
-        (fval, aux), g_flat = self._vag_flat_with_aux(params_flat, *args)
-
-        gnorm = jnp.linalg.norm(g_flat)
+        (fval, aux), grad = self._gradient(params, *args)
+        gnorm = jnp.sqrt(
+            jax.tree_util.tree_reduce(
+                lambda a, b: a + b,
+                jax.tree.map(lambda x: jnp.vdot(x, x), grad),
```
BalzaniEdoardo: same as before, either use `jnp.add` or lineax `tree_dot`

test.py
BalzaniEdoardo, line 1: Is this a leftover?

BalzaniEdoardo (changes requested):
This basically LGTM except a few very very minor comments.

The only question I have is if we are still checking that the vmap Newton step matches a regular step.

If not, what i would do is set up the Newton solver of PopulationGLM both via the current vmap based hessian construction, or passing None as the hessian. 

Then get the solver and run updates matching outputs. you can parametrize having a feature mask or not, etc. 

The hessian None will be our ground truth since use plain autodiff on the full problem.

BalzaniEdoardo (approved):
LGTM! thanks for this cleanup

PR #593, opened 2026-07-20
Refactor terminology: `epoch`->`passes` + improve dataloder note

BalzaniEdoardo:
## Summary

This PR focuses on two aims:

- Drop the machine-learning `epoch` terminology that conflicts with `pynapple`'s recording epochs. Since `pynapple` is ubiquitous in the documentation, the conflicting terminology was too confusing, especially for neuroscientist not familiar with the ML literature.
- Improve the custom pynapple data loader doc note. In particular, a new, multi-(pynapple)epoch friendly dataloader is provided. This loader can be used as a template of how to handle stochastic fits and batching in the presence of a disjoint time series.

## Major changes

- `epoch` in the machine-learning sense has been replaced with `passes` throughout the package.
- A new session setting up a `MultiEpochPynappleDataLoader` class is added to `custom_data_loader.md`. The whole note has been restructured for flow and clarity.

### DIFF
--- a/docs/how_to_guide/batching/custom_callbacks_and_termination.md
+++ b/docs/how_to_guide/batching/custom_callbacks_and_termination.md
@@ -129,7 +129,7 @@ class EarlyStoppingCallback(nmo.callbacks.Callback):
     y_test :
         Test target data.
     patience :
-        Number of training epochs (passes through the data) with no improvement before stopping.
+        Number of passes with no improvement before stopping.
     min_delta :
         Minimum decrease in loss to count as an improvement.
     """
--- a/docs/how_to_guide/batching/custom_callbacks_and_termination.md
+++ b/docs/how_to_guide/batching/custom_callbacks_and_termination.md
@@ -54,7 +54,7 @@ nap.nap_config.suppress_conversion_warnings = True
 np.random.seed(123)
 ```
 
-First, quickly set up and run a stochastic fit for 10 epochs:
+First, quickly set up and run a stochastic fit for 10 passes:
 
 ```{code-cell} ipython3
 _simulate_and_write_to_disk()
@@ -80,7 +80,7 @@ glm = nmo.glm.PopulationGLM(
     solver_kwargs={"stepsize": 0.05, "acceleration": False},
 )
 
-glm.stochastic_fit(loader, num_epochs=10, callbacks=batch_logger)
+glm.stochastic_fit(loader, n_passes=10, callbacks=batch_logger)
 
 plot_loss_history(batch_logger.loss_history)
 ```
@@ -89,13 +89,13 @@ plot_loss_history(batch_logger.loss_history)
 
 +++
 
-Instead of guessing how many epochs we need, undershooting and restarting, or overshooting and waiting too long, we can use callbacks to run optimization until convergence.
+Instead of guessing how many passes we need, undershooting and restarting, or overshooting and waiting too long, we can use callbacks to run optimization until convergence.
 
 +++
 
 As mentioned before, to monitor training progress and the optimization's state during the fitting run, NeMoS has a callback system, allowing to execute custom code on the following events:
 - beginning and end of training
-- before and after each epoch
+- before and after each pass
 - before and after each batch
 
 +++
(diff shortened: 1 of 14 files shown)


BalzaniEdoardo:
@bagibence FYI, we took this executive decision after a long discussion.

docs/how_to_guide/batching/custom_callbacks_and_termination.md
```diff
@@ -129,7 +129,7 @@ class EarlyStoppingCallback(nmo.callbacks.Callback):
     y_test :
         Test target data.
     patience :
-        Number of epochs with no improvement before stopping.
+        Number of training epochs (passes through the data) with no improvement before stopping.
```
gviejo: Do you want to say : "Number of passes with no improvement before stopping"?

docs/how_to_guide/batching/custom_dataloader.md
```diff
@@ -54,30 +57,32 @@ nap.nap_config.suppress_conversion_warnings = True
 np.random.seed(123)
 ```
 
-Load the spike times:
+## The `DataLoader` interface
+
+A [`DataLoader`](nemos.batching.DataLoader) is any object that streams `(X_batch, y_batch)` pairs to [`stochastic_fit`](nemos.glm.PopulationGLM.stochastic_fit). It has to provide three methods:
+- `__iter__`: yields `(X_batch, y_batch)` tuples. It is called once at the start of every epoch and must return a fresh iterator each time (using `yield` here makes the method a generator, which does this automatically).
```
gviejo:
```suggestion
- `__iter__`: yields `(X_batch, y_batch)` tuples. It is called once at the start of every pass and must return a fresh iterator each time (using `yield` here makes the method a generator, which does this automatically).
```

docs/how_to_guide/batching/custom_dataloader.md
```diff
@@ -54,30 +57,32 @@ nap.nap_config.suppress_conversion_warnings = True
...
+- `n_samples`: the total number of samples in the dataset.
```
gviejo: Can you actually know before iteration? Is it n_samples or n_batches?

docs/how_to_guide/batching/custom_dataloader.md
gviejo, line 93: I would give a minimal docstrings here with the type of inputs

docs/how_to_guide/batching/custom_dataloader.md
```diff
@@ -135,38 +140,162 @@ class PynappleDataLoader(nmo.batching.DataLoader):
...
+
+Putting the pieces together:
+
+```{code-cell} ipython3
+class MultiEpochPynappleDataLoader(nmo.batching.DataLoader):
+    def __init__(self, spike_times, basis, interval_size, bin_size, shuffle=True, seed=123):
+        self.spike_times = spike_times
+        self.basis = basis
+        self.bin_size = bin_size
+        # window length is in bins; the left context to prepend is that many bins, in seconds
```
gviejo: Not super clear sentence

docs/how_to_guide/batching/custom_dataloader.md
```diff
@@ -135,38 +140,162 @@ class PynappleDataLoader(nmo.batching.DataLoader):
...
+        self.context_duration = basis.window_size * bin_size
+        self.shuffle = shuffle
+        self.rng = np.random.default_rng(seed)
+
+        # Equal-duration chunks built within each recording interval, so none crosses a gap; the
+        # sub-interval_size tail of an interval is kept as a final short chunk. Each row is
+        # (chunk_start, chunk_end, interval_start): the interval start lets us clip the left
+        # context without reaching back into the previous interval.
+        self._chunks = np.array([
```
gviejo: Do you want to show the split function here instead?

docs/how_to_guide/batching/stochastic_fit.md
```diff
@@ -187,12 +187,12 @@ Other solvers that can be used for stochastic optimization can be listed with [`
 
 We are ready to start the optimization using the GLM's `stochastic_fit` method.
 
-`stochastic_fit` uses `num_epochs` to control the training duration and does not stop on convergence by default. Unless a callback requests a stop, it will run for the full number of epochs.
+`stochastic_fit` uses `n_passes` to control the training duration and does not stop on convergence by default. Unless a callback requests a stop, it will run for the full number of passes.
```
gviejo:
```suggestion
`stochastic_fit` uses `n_passes` to control the number of passes over the full dataset and does not stop on convergence by default. Unless a callback requests a stop, it will run for the full number of passes.
```

docs/how_to_guide/batching/custom_dataloader.md
BalzaniEdoardo: Good point. We added it because `n_samples` is needed to compute the residual dof of the model, which are used by the scale estimation. it affects the score. The scale is 1 for Poisson and Bernoulli for example, so it is not relevant for those models, but for Gaussian is the standard deviation and must be estimated. 

For example: since the dataloader can start directly form an array of spike times, or from an nwb or from a lazy loaded zarr, and the pre-processing is up to the user, a-priori it is unspecified how many samples the design will have.

BalzaniEdoardo: There is always the option, for a user, to dump an estimate. It will have slightly off dof, but it would still work.

docs/how_to_guide/batching/custom_dataloader.md
BalzaniEdoardo: simplified

docs/how_to_guide/batching/custom_dataloader.md
gviejo: I would write that in the documentation then.

gviejo (approved):

PR #597, opened 2026-07-29
Changes to GLM-HMM update method for custom solvers

sjvenditto:
A small PR that changes the `update` method of `GLMHMM` and validation of `session_starts`.

**Motivation:** Input validation and `session_start` casting was happening inside of `update`, which can lead to suboptimal performance with repeated, redundant checks. Since this method is only called in advanced use cases, I've modified `update` to require a properly formatted boolean array for `session_starts` instead of the more flexible input types allowed in `fit`. Now, validation only occurs a single time when calling `initialize_optimizer_and_state`, and the `update` method will only drop NaN values, which is how it was implemented in the GLM.

Additional changes:
- to keep the `initialize_optimizer_and_state` model agnostic, I've added a custom `validate_inputs` to `HMMValidator` that adds a shape check on `session_starts`. I've noted why this exists in this function, since the shape check is redundant with `validate_and_cast_session_starts` in the normal validation pipeline. 
- `initialize_optimizer_and_state` has been updated to pass optional kwargs to its`validate_inputs` call to accommodate the HMM version
- ~~with the custom `validate_inputs`, it requires `validate_and_cast_session_starts` to occur first in the validation sequence so that `session_starts` is reformatted if it is an `IntervalSet` or list of session start indices.~~ I've updated it so that `validate_input` runs first again. now, there's a kwarg `validate_session_starts`, which defaults to true, on whether `session_starts` get validated in `validate_inputs` if it gets passed. now it is skipped in the validation sequence called by `validate_and_cast_inputs`
- I've added a test for `session_starts` in the new the `validate_input`, `test_validate_inputs_session_starts`
- I've fixed tests that call the GLMHMM `update` to also pass `session_starts`. I've also modified tests that check for validation calls in update, since validation is no longer being done here.

Possible issues:
- the data type of `session_starts` is not being validated, only the shape, to be consistent with how `X` and `y` are validated. should we also add a check that it's a boolean array?
     - in order to ensure that the first sample in `session_starts` is 1, I'm now recasting it s a jax boolean array and forcing the first index to be true after NaNs are dropped
- if there are NaN values in `X` or `y` during a session start, the onus is on the user to shift the session start indices. should we expose the `session_starts` casting / NaN shifting in an example to flag this?


src/nemos/glm_hmm/glm_hmm.py
```diff
@@ -1545,18 +1544,20 @@ def update(
...
-        >>> new_params, new_state = model.update(init_params, opt_state, X, y)
+        >>> opt_state = model.initialize_optimizer_and_state(init_params, X, y, session_starts=session_starts)
+        >>> new_params, new_state = model.update(init_params, opt_state, X, y, session_starts)
         """
-        # validate inputs and session boundaries
-        session_starts = self._validator.validate_and_cast_inputs(
-            X, y, session_starts=session_starts
-        )
+        # find non-nans
+        X, y, session_starts = tree_utils.drop_nans(X, y, session_starts)
```
BalzaniEdoardo: here if there are nans in the middle of a session, they'll be dropped, and the em will assume continuous samples, and this may not be true. 

We may want to add the check on nan placement

BalzaniEdoardo: the other option is dropping all checks

BalzaniEdoardo: after a discussion, we can add `safe=True` as an argument, skipping non-necessary processing.

BalzaniEdoardo (commented):
Add a couple of ideas to keep the implementation bugfree when handling the nans while skipping checks if users need performance

src/nemos/glm_hmm/glm_hmm.py
```diff
@@ -1480,9 +1482,10 @@ def update(
         opt_state: NamedTuple,
         X: DESIGN_INPUT_TYPE,
         y: jnp.ndarray,
+        session_starts: jnp.ndarray,
         *args,
-        session_starts: Optional[jnp.ndarray] = None,
```
BalzaniEdoardo, line 1484: The static checker complains that the signature doesn't match base regressor. One way around it is to make it optional, and internally:
- Assume single session if None (session_start=[1,0,0,0,0]...)
- Use the bool array the user pass otherwise. 

What do you think?

BalzaniEdoardo (approved):
This LGTM except the signature mismatch to reconcile. 

I proposed one solution, let me know if that seems reasonable. Once that is resolved one way or another, we can merge this one

src/nemos/glm_hmm/glm_hmm.py
```diff
@@ -1480,9 +1482,10 @@ def update(
         opt_state: NamedTuple,
         X: DESIGN_INPUT_TYPE,
         y: jnp.ndarray,
+        session_starts: jnp.ndarray,
         *args,
```
BalzaniEdoardo:
```suggestion
        *args,
        session_starts: Optional[jnp.ndarray] = None,

```

BalzaniEdoardo:
Merged a quick fix after the release 2D ago of [JAX v0.11.1](https://github.com/jax-ml/jax/releases/tag/jax-v0.11.1). One of the computation was loosing precision in the vectorized vs regular path. I realized that the reason why this happened in the first place is that in a few spot we were (my fault) computing log(1+x) directly instead of using log1p(x).

src/nemos/glm_hmm/glm_hmm.py
sjvenditto: I think this works

PR #600, opened 2026-08-05
Report an informative error for invalid observation models

### LINKED ISSUE #337: Add more informative error message for invalid observation models
After #329  is merged improve error message as highlighted in the conversation of #335.

potential test for it..
```python
@pytest.mark.parametrize(
    "obs_model, expectation",
    [
        (nmo.observation_models.PoissonObservations(), does_not_raise()),
        (nmo.observation_models.GammaObservations(), does_not_raise()),
        (1, pytest.raises(ValueError, match="Invalid observation model: 1")),
    ],
)
def test_invalid_obs_models(obs_model, expectation):
    with expectation:
        nmo.glm.GLM(observation_model=obs_model)
```

adityasingh2400:
Closes #337.

Passing something that is not an observation model to `GLM` or `GLMHMM` fails with `AttributeError: The provided object does not have the required '_negative_log_likelihood' attribute!`. That message describes an internal method and gives the user nothing to act on.

The root cause is that `check_observation_model` only validates the interface attribute by attribute. When the object implements none of them, the first missing attribute is reported as if the user had written a nearly complete observation model.

`check_observation_model` now detects objects that implement none of the required attributes and raises a `ValueError` naming the offending value, listing the available observation models, and linking the docs. This follows the wording @billbrod suggested in the issue thread. Objects that implement part of the interface still get the existing per-attribute `AttributeError`, which is the useful message in that case, so `test_missing_method` and friends are untouched.

The new message:

```
ValueError: Invalid observation model: 1. The observation model must be one of
['Bernoulli', 'NegativeBinomial', 'Gamma', 'Poisson', 'Gaussian', 'Categorical'],
provided either as a string (for example "Poisson") or as an instance (for example
PoissonObservations()), or a custom object implementing the nemos observation model
interface, see https://nemos.readthedocs.io/en/latest/api/observation_models.html for details.
```

Tested with the reporter's own parametrized test, adapted to the repo's conventions. `TestObservationModel::test_init_observation_type` in `tests/test_base_regressor_subclasses.py` already parametrized `1` and `Regularizer`, so those expectations moved from `AttributeError` to the new `ValueError`, which covers GLM, PopulationGLM and GLMHMM. Two new tests in `tests/test_observation_models.py` cover the direct `check_observation_model` path and assert the message content.

Restoring `src/nemos/observation_models.py` from `main` makes the new tests fail with the old `AttributeError`, and they pass with the change. The full non-solver suite runs clean at 40385 passed and 1236 skipped. black, isort, flake8, ruff and `check_parameter_naming.py` are all clean.

One implementation note for review: `AVAILABLE_OBSERVATION_MODELS` lives in `_observation_model_builder`, which imports from `observation_models`, so the new message imports it function-locally to avoid a cycle. There is a comment marking this in the diff, in case you would rather move the constant.


BalzaniEdoardo (approved):
This LGTM, thanks for the PR,  nice job. Can you merge development? I re-based the PR to development. Our github workflow is: we merge new features to development, when we merge development to main we usually release an new version.

BalzaniEdoardo:
nevermind, I merged development. 

Additionally,  Jenkins is set to not run on external forks for authors with no write access, I am still figuring out the best way around this for new external contributors.

adityasingh2400:
Thanks for the review and for merging development, and sorry for making you do it yourself.

The Jenkins note explains something I was about to chase on #601. Its readthedocs check is failing with "No published documentation at docs/pulls/601", which is the RTD build refusing to serve a tree Jenkins has not published yet. I had read that as an ordering race that would clear once Jenkins finished, but if Jenkins does not run on external forks at all then it is not a race, the docs simply never get published and RTD cannot pass on that PR no matter what I do. Happy to leave it red and let you judge it, rather than pushing anything to try to work around it.

For what it is worth, on #601 I rebased and also normalized the four new `docs/how_to_guide/batching/` notebooks that arrived from development, since the check that PR adds would otherwise fail on the merge result.

BalzaniEdoardo:
> Thanks for the review and for merging development, and sorry for making you do it yourself.
> 
> The Jenkins note explains something I was about to chase on #601. Its readthedocs check is failing with "No published documentation at docs/pulls/601", which is the RTD build refusing to serve a tree Jenkins has not published yet. I had read that as an ordering race that would clear once Jenkins finished, but if Jenkins does not run on external forks at all then it is not a race, the docs simply never get published and RTD cannot pass on that PR no matter what I do. Happy to leave it red and let you judge it, rather than pushing anything to try to work around it.

That's exactly how it works, PR triggers docs build in jenkins, jenkins publish on flatironinstitute/nemos-documentation, and then triggers another RTD build which succeeds. In the case of an external fork, jenkins never runs, so the docs are never built. 

> For what it is worth, on #601 I rebased and also normalized the four new `docs/how_to_guide/batching/` notebooks that arrived from development, since the check that PR adds would otherwise fail on the merge result.

For this branch, since I committed, the CI is going through, we just need to wait for it to finish. For the other PR (#601), either I add an empty commit as a workaround, or I'll figure out something more sustainable.

adityasingh2400:
Correction to what I said above: #601 has gone green, Jenkins included, so the readthedocs failure there was the ordering race after all and not a hard block from the fork restriction. I read your Jenkins note as broader than you meant it. Sorry for the noise.

PR #601, opened 2026-08-05
Add notebook metadata linting to CI

### LINKED ISSUE #316: Add notebook linting to CI
Right now, we're getting diffs show up in the PRs for the myst-nb markdown files, depending on whether they were opened using an editor or jupyter lab directly. That's distracting, so should add something simple to take care of that.

adityasingh2400:
Closes #316.

The myst-nb markdown notebooks under `docs/` carry a jupytext front matter. Saving one from Jupyter Lab rewrites two fields that have no effect on how the notebook is executed or rendered: jupytext stamps whichever version happened to be installed, and the kernel display name is replaced with the one of the kernel that was used. Those rewrites show up as review noise, which is what the issue describes.

The repository is carrying the evidence right now. Across the 34 notebooks, `jupytext_version` has three values (1.16.4 twenty two times, 1.18.1 seven times, 1.19.1 five times) and `display_name` has three (`Python 3 (ipykernel)` thirty times, `Python 3` three times, and `nemos (3.12.11)`, which is somebody's local environment name). Everything else in the header, `extension`, `format_name`, `format_version`, `language` and `name`, is already uniform. These counts are against the merge base and cover the four `docs/how_to_guide/batching/` notebooks that arrived on development during the rebase.

`[tool.jupytext] notebook_metadata_filter` in `pyproject.toml` records the intent in one place, and `scripts/check_notebook_metadata.py` enforces it. The script is check only, it never rewrites. It parses the front matter with `yaml.safe_load` and inspects the resulting dict rather than matching text, so it does not depend on how jupytext happens to format the block. The 34 md headers in the diff have the two fields stripped.

One thing did not work out as expected, and I would rather flag it than quietly leave a setting that does nothing. On jupytext 1.19.5 `notebook_metadata_filter` does not actually remove either field, because the filter only understands top level metadata keys while both of these are nested. I tested that rather than assuming it, by round-tripping a notebook through jupytext with the setting in place. The setting is kept so the intent is recorded and starts working if jupytext gains sub-key support, and the script is what enforces the rule today.

Wiring follows the existing `check-parameter-naming` pattern: a local pre-commit hook, and `python scripts/check_notebook_metadata.py` in `tox -e check`, which is what the CI `tox_check` job runs. There is no `tox -e fix` entry and no unit test, since the CI check is what catches violations.

### Correction to an earlier version of this description

An earlier version of this description said the script was stdlib only and therefore ran inside the `skip_install` tox envs. That was wrong once the script moved to `yaml.safe_load`. `yaml` is PyYAML, a third party package, and it was present only transitively through myst-nb rather than declared anywhere. The first push of that rework turned `tox_check` red with `ModuleNotFoundError: No module named 'yaml'`, on the very check this PR adds.

The fix is in 21895db. `pyyaml` is now declared in the `[testenv:check]` deps and in the `dev` extra, the latter because the pre-commit hook runs with `language: system` against a dev install. I reproduced the failure in a bare virtualenv before and after: without PyYAML the script dies exactly as CI did, and with it the check passes and reports all notebooks clean. `tox_check` is green again. No notebook content changed as part of that fix, only the two dependency declarations.

### Verification

Stripping the fields does not change how the notebooks parse. Reading each of the 34 notebooks with `jupytext.reads(..., fmt="md:myst")` before and after gives identical cell types and sources.

Restoring `docs/` from the merge base makes the check exit 1 and list all 34 files, and it exits 0 on the branch as it stands. I also dirtied `docs/quickstart.md` on its own with `jupytext_version: 1.99.0` and `display_name: someones-local-venv`, and the check flagged that single file with both reasons.

One note: the new hook prints `Executable 'python' not found` when `python` is not on PATH. The existing `check-parameter-naming` hook fails identically in that situation, so this follows the repo's existing `language: system` convention rather than introducing a new failure mode.


BalzaniEdoardo:
Hi, thanks for this other pr too, very useful. can you merge development?

adityasingh2400:
Rebased onto development to clear the conflict.

The conflict was on `docs/how_to_guide/plot_04_batch_glm.md`, which development deleted in #510 while this branch had normalized its metadata. I took the deletion, since the file is gone upstream and the normalization of it is moot.

That rebase also brought in the four new `docs/how_to_guide/batching/` notebooks, which still carry a `jupytext_version` line. Since the whole point of this PR is to make CI reject exactly that, I normalized them too, otherwise the check this PR adds would fail on the merge result.

Verified on the rebased branch: `normalize_notebook_metadata.py --check` reports all notebooks normalized, and the 4 tests in `tests/test_normalize_notebook_metadata.py` pass.

BalzaniEdoardo (changes requested):
We should not rely on regex to remove the fields. If jupytext changes how the metadata is formatted, or our assumption about it are not matched, we may introduce future bugs. 

I prefer adding a configuration for that in the `pyproject.toml`.

```python
[tool.jupytext]                                                                                                                                                                                        
notebook_metadata_filter = "-jupytext.text_representation.jupytext_version,-kernelspec.display_name"      
```

This should be enough to filter the fields out. If not check what can be done [here](https://jupytext.org/reference/advanced-options/).

I would like the following changes:

- The `notebook_metadata_filter` added to the pyproject
- Keep the edits to the md files with the stripped header.
- convert `scripts/check_notebook_metadata.py` to check only (remove the fix option)
- keep the same pre-commit hook and `tox -e check entry`
- drop the `tox -e fix entry`
- drop the test_docs_notebooks_are_normalized, the ci `tox check entry` will catch violations.

Since we are dropping the fixer, the checker machinery can be as simple as:

```python
import yaml


DELIMITER = "---"
CANONICAL_DISPLAY_NAME = "Python 3 (ipykernel)"


def front_matter(text):
    """Return the parsed YAML front matter of a markdown file, or None if it has none."""
    lines = text.splitlines()
    if not lines or lines[0] != DELIMITER:
        return None
    for index, line in enumerate(lines[1:], start=1):
        if line == DELIMITER:
            return yaml.safe_load("\n".join(lines[1:index]))
    return None

def problems(path):
    meta = front_matter(path.read_text(
        encoding="utf-8"))  # yaml.safe_load, or None
    if not isinstance(meta, dict) or "jupytext" not in meta:
        return []  # not a notebook
    found = []
    text_repr = meta["jupytext"].get("text_representation", {})
    if "jupytext_version" in text_repr:
        found.append("carries jupytext_version")
    display_name = meta.get("kernelspec", {}).get("display_name")
    if display_name not in (None, CANONICAL_DISPLAY_NAME):
        found.append(f"display_name is {display_name!r}")
    return found
```

adityasingh2400:
Done, all six points are in 7a798d2. One thing did not work out as expected though, and I would rather flag it than quietly leave a setting that does nothing.

What changed:

- `[tool.jupytext] notebook_metadata_filter` added to `pyproject.toml`
- `scripts/normalize_notebook_metadata.py` replaced by `scripts/check_notebook_metadata.py`, check only, built on the skeleton you wrote. No regex, it parses the front matter with `yaml.safe_load` and inspects the dict, which I agree is the right call
- pre-commit hook and the `tox -e check` entry kept, pointed at the new script
- `tox -e fix` entry dropped
- `test_docs_notebooks_are_normalized` and the rest of the fixer tests dropped
- the stripped md headers kept as they were

The part that did not work: **that filter does not actually remove either field on jupytext 1.19.5.** I tested it rather than assuming, round-tripping a notebook md to ipynb and back inside a directory carrying the config:

| `notebook_metadata_filter` | jupytext_version | display_name | kernelspec.name | format header |
|---|---|---|---|---|
| `-jupytext.text_representation.jupytext_version,-kernelspec.display_name` | kept | kept | kept | kept |
| `-kernelspec` | kept | **removed** | removed | kept |
| `-all` | **removed** | **removed** | removed | removed |
| `kernelspec.name,kernelspec.language,-all` | kept | kept | kept | kept |
| `kernelspec,-kernelspec.display_name` | kept | kept | kept | kept |

The `-kernelspec` row is the useful one: it proves jupytext is reading the config from `pyproject.toml`, since that filter does strip `display_name` by dropping the whole section. So this is not a config discovery problem, the sub-key paths are simply not honoured. The trait help agrees, its examples are all top level keys (`'all'`, `'-all'`, `'widgets,nteract'`, `'kernelspec,jupytext-all'`), and there is no option anywhere in `JupytextConfiguration` that suppresses the version stamp. `-all` does remove both, but it also strips `kernelspec.name` and the jupytext format header, which myst-nb needs to pick a kernel.

So today the checker is the thing actually enforcing this, not the filter. I kept the `pyproject.toml` entry because you asked for it and it records the intent in one place, but I wrote the comment above it to say plainly that it is currently inert for these two fields, so nobody later reads it as working. Happy to drop it instead if you would rather not carry a setting that does nothing yet, just say which you prefer.

Verified locally: the checker passes on the current tree, exits 1 with a useful message when I reintroduce either field, and black and isort are clean on the new script.

adityasingh2400:
Caught and fixed a break I introduced with the rework, in 21895db.

`tox -e check` went red with `ModuleNotFoundError: No module named 'yaml'`. The old script parsed the front matter with a regex and so had no dependencies, while the new one uses `yaml.safe_load` as you asked, which means the check env actually needs PyYAML. It was only present transitively through myst-nb, not declared anywhere.

So `pyyaml` is now in the `[testenv:check]` deps and in the `dev` extra, the latter because the pre-commit hook runs with `language: system` against a dev install.

I reproduced it before and after in a bare virtualenv: without PyYAML the script dies exactly as CI did, with it the check passes and reports all notebooks clean.

adityasingh2400:
`tox_check` is green now after the PyYAML declaration.

The remaining red is `check-external-links`, and it is not from this branch. The only failing entry is the YouTube link in `README.md` and `docs/index.md` coming back `429`, which is the runner being rate limited rather than the link being broken. That link is already on develop and this PR touches neither file, its diff is confined to `pyproject.toml`, `tox.ini`, `.pre-commit-config.yaml`, `scripts/` and the notebook front matter.

Left alone rather than pushed at, since a rerun is all it needs.

BalzaniEdoardo:
> `tox_check` is green now after the PyYAML declaration.
> 
> The remaining red is `check-external-links`, and it is not from this branch. The only failing entry is the YouTube link in `README.md` and `docs/index.md` coming back `429`, which is the runner being rate limited rather than the link being broken. That link is already on develop and this PR touches neither file, its diff is confined to `pyproject.toml`, `tox.ini`, `.pre-commit-config.yaml`, `scripts/` and the notebook front matter.
> 
> Left alone rather than pushed at, since a rerun is all it needs.

sounds good, I'll re-run failed jobs once the ci terminates.

adityasingh2400:
Merged development in, sorry for the delay on that. Your #600 had landed there since I last synced, so the branch was one commit behind.

The merge only brought in the observation model changes from #600 and left this PR's own files alone. `check_notebook_metadata.py` still reports all notebooks clean on the merged tree.

Thanks for offering to re-run the failed jobs.

BalzaniEdoardo:
Updates on the branch:

### The filter works; the two-key version breaks the notebooks

The PR carried `notebook_metadata_filter = "-jupytext.text_representation.jupytext_version,-kernelspec.display_name"` with a comment stating that jupytext only understands top level keys, so the setting is inert. Neither half actually was factual.

Tested on a dirtied copy of the real `docs/quickstart.md` placed inside the repo, so jupytext discovers the repo's own `pyproject.toml`, driving both the CLI round trip and `jupytext.TextFileContentsManager.save`, which is the call Jupyter Lab makes when saving. Same results on jupytext 1.16.4, 1.18.1, 1.19.1 and 1.19.5.

| config | `jupytext_version` | `display_name` | file still readable |
|---|---|---|---|
| none (baseline) | restamped to the local version | local value kept | yes, 53 cells |
| `-jupytext.text_representation.jupytext_version` | removed | kept | yes, 53 cells |
| `-…jupytext_version,-kernelspec.display_name` (as on the PR) | removed | removed | **no** |

What the two-key filter actually does is strip `display_name` on the next Lab save, after which the file cannot be parsed:

```
jupytext/myst.py:282  myst_to_notebook -> nbformat/v4/nbbase.py:170 new_notebook -> validate(nb)
nbformat.validator.NotebookValidationError: 'display_name' is a required property
```

`kernelspec.display_name` is a **required property** of the nbformat v4 schema, so a notebook without it is invalid by construction. This is not a jupytext bug to patch upstream, and it is not filterable by any other means.

CI would not have caught it. The 34 committed notebooks all still have a `display_name`, so the check would pass; the breakage appears the first time somebody saves a notebook from Lab with that config in place.

The likely source of the "kept / kept" observation is config discovery: on the CLI, jupytext resolves the configuration relative to the working directory. Running the same conversion from outside the repo against an absolute path applied no filter at all and reproduced that row exactly. If one runs from the repo root, the filter works.

### Changes

`pyproject.toml`

```toml
notebook_metadata_filter = "-jupytext.text_representation.jupytext_version"
```

The version stamp is removed at the source, on every jupytext version tested, through both the CLI and the Lab save path, with the file still readable. `display_name` stays in the file and it's checked against a default name. The comment is also updated.

`scripts/check_notebook_metadata.py`

- `display_name != CANONICAL_DISPLAY_NAME` replaces `not in (None, CANONICAL_DISPLAY_NAME)`. A missing display name is exactly the state the old config produced, so accepting it meant generating an unreadable notebook. 
- The docstring records the split: the filter handles the version stamp, the script handles the display name.
- The remediation message no longer suggests re-saving through jupytext, for the reason below.

### Verification

| check | result |
|---|---|
| Lab save on a dirtied `quickstart.md`, new config | version gone, `display_name` = `Python 3 (ipykernel)`, kernel name and myst format intact, 53 cells, readable |
| checker on the tree | exit 0 |
| checker vs. `jupytext_version: 1.99.0` + `display_name: someones-local-venv` | exit 1, both reasons reported |
| checker vs. missing `display_name` | flagged, previously accepted |
| `black --check scripts`, `isort --check scripts --profile=black` | pass |

BalzaniEdoardo:
Also, I am now reverting back to editing the yaml to normalize the display name as part of tox fixes. Otherwise one would need to manually fix the affected markdowns at every pre-commit fail

BalzaniEdoardo:
Follow up:

`scripts/check_notebook_metadata.py`

- takes optional path arguments, and still walks `docs/` when given none.
- `--fix` deletes the `jupytext_version` line, resets `display_name` to the canonical value, and inserts it under `kernelspec:` when the key is missing altogether. It exits non zero after rewriting, so a pre-commit run stops and the repaired files get staged rather than committed in their original state.
- the rewrite is line based rather than a YAML dump: PyYAML reorders the keys, and even with `sort_keys=False` it unindents block sequences, turning `  - navigation` into `- navigation`. YAML parsing decides what is wrong, a line edit acts, and everything outside those two lines stays byte for byte.
- the front matter delimiter scan is factored out of `front_matter` into `front_matter_end` and reused by the rewriter, in the same form as `add_download_admonition` in `docs/conf.py`.

`.pre-commit-config.yaml`: the hook becomes `fix-notebook-metadata` and runs with `--fix`. `pass_filenames` is left at its default so pre-commit passes the staged files, and the hook never rewrites a notebook that is not part of the commit.

`tox.ini`: `python scripts/check_notebook_metadata.py --fix` added to `[testenv:fix]`, with the `pyyaml` dependency that env was missing. `[testenv:check]` is unchanged and stays check only.

BalzaniEdoardo:
@adityasingh2400 thank you so much for all contributions!

Issue #569, opened 2026-05-28
Update docstring language to be more general that just neurons


sjvenditto:
With the addition of observation models like Bernoulli and Categorical, our GLM models can be applied to experimental observations beyond neural spiking activity. It may be conducive to update the language to specify that `y` can be other types of observations as well, especially with the GLM-HMM where the default observation model is Bernoulli

BalzaniEdoardo:
closed with #602

PR #602, opened 2026-08-05
Generalize observation model docstrings beyond neural spiking

adityasingh2400:
Refs #569. This covers `observation_models.py` and `glm/glm.py`, not the whole codebase, so I have left the issue open rather than closing it.

Now that the GLM and GLM-HMM accept Bernoulli, Gaussian, Gamma, Negative Binomial and Categorical observations, the target variable is not necessarily neural spiking. In several places the old language was not just neuro centric but wrong for the family being documented. `GammaObservations` was described as modelling "the observed spike counts", and `GaussianObservations` as computing the deviance "for the given spike count data".

**The first commit is docstrings only.** The `y` and `predicted_rate` descriptions now talk about observations and the predicted mean, and each family summary describes the data it actually models, keeping neural spiking as the example for Poisson where it belongs. It also fixes a typo, "The spike coun activity", in the Gamma deviance docstring. Shape notation such as `(n_time_bins, n_neurons)` is left alone, since that is established API vocabulary and renaming it is a much larger change.

**The second commit is separable and touches public signatures, so please feel free to drop it on its own if you would rather handle the rename separately or not at all.** The first argument of `Observations.deviance` is currently called `spike_counts` on the abstract class and Poisson, `neural_activity` on Gamma and Gaussian, and `observations` on Bernoulli, Negative Binomial and Categorical. The keyword a caller must use therefore depends on which family they picked, and two of the three names assume the observations are neural. It renames them all to `observations`, the name the three most recently added families already use.

Every `deviance` call site inside nemos, in `src` and in `tests`, passes the argument positionally, so nothing internal breaks. The risk is external callers using the old keyword.

Read straight out of git, the first argument of `deviance` across the 7 classes:

```
BEFORE (main):
   2   neural_activity: jnp.ndarray,
   3   observations: jnp.ndarray,
   2   spike_counts: jnp.ndarray,

AFTER:
   7   observations: jnp.ndarray,
```

A fail before check does not apply to a docstring sweep, and I am not going to manufacture one. The rename has the checkable before state shown above.

Verification: doctests for both edited modules pass at 17. `tests/test_observation_models.py` plus `tests/test_glm.py` give 8190 passed and 581 skipped. The full non solver suite, run to completion, gives 40379 passed and 1236 skipped. black, isort, flake8, ruff and `check_parameter_naming.py` are all clean.

### DIFF
--- a/src/nemos/glm/glm.py
+++ b/src/nemos/glm/glm.py
@@ -489,10 +489,10 @@ class GLM(BaseRegressor[GLMUserParams, GLMParams, GLMValidator]):
         See Also
         --------
         :meth:`nemos.glm.GLM.score`
-            Score predicted rates against target spike counts.
+            Score predicted rates against the observations.
 
         :meth:`nemos.glm.GLM.simulate`
-            Simulate neural activity in response to a feed-forward input (feed-forward only).
+            Simulate observations in response to a feed-forward input (feed-forward only).
 
         :func:`nemos.simulation.simulate_recurrent`
             Simulate neural activity in response to a feed-forward input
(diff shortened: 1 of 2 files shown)


BalzaniEdoardo (approved):
Thanks for this one too!

PR #607, opened 2026-08-08
Partial param spec

BalzaniEdoardo:
## Summary and Motivation
With this PR we add support for partial parameter specifications in GLMs. This feature has been requested by multiple users, including in this discussion #485. 

This PR addresses that. 

### GLM

- **new** `fix_params` parameter. This is of type `UserGLMParams`, which is a `tuple[PyTree[NDArray | None] | None, NDArray|None]` coef and intercept. Any concrete array in the tree will be fixed, the None will be learned.
- Extended initialization logic to account for `fix_params`: similar to the `fit_intercept=False` initialization logic, when the intercept is frozen, either by flag or by `fix_params` the active coefficients are initialized to a constant minimizing $\Vert g(y) - X_\text{fix} \cdot b_\text{fix}- X_\text{active} \cdot \mathbf{1} b \Vert _2^2$.
- Resolution order for fixing the intercept: if it is provided by `fix_params`, the pinned value is used. Otherwise, the `fit_intercept` flag determines if the intercept is pinned to zero or not.
- DOF computation is now based on the active parameters only.
- If all params are fixed a NoOpSolver is defined matching the interface.

### Validator
- **validate_param_specs**: method of the base validator. This checks that the `fix_params` has the expected structure, and that its leaves are None or arrays.
- **additional_validation_param_specs**: hook to validate additional model specific structure.
- BaseValidator validation chain for fix parameter specifications.

### Solvers
- **new** convenience No-op solver, can be used for trivial model fits (no params)
- changed the interface of self.fun in optimistix adapter to match the nemos loss. This can be safely done because the function that optimistix minimise receives is self.fun_with_aux, which follows their interface (`fun(params, args=None)`).

### BaseRegressor
- convenience method for no-op optimizer setup.

### Parameters

- Typing of model parameters was generalized, allowing more precise `Params[LeafType]` specification. this is useful because we can annotate if a method requires the fully specified parameters `GLMParams[NDArray]` or if it can accept partial parameters specs: `GLMParams[NDArray | None]` .

### Repr

- `format_repr` now filters out tree with all None leaves from the repr. This is the default for `fix_params`.

### DIFF
--- a/tests/test_partial_param_spec.py
+++ b/tests/test_partial_param_spec.py
@@ -5,10 +5,9 @@ import jax
 import jax.numpy as jnp
 import numpy as np
 import pytest
-
-import nemos as nmo
 from sklearn.base import clone
 
+import nemos as nmo
 from nemos.batching import ArrayDataLoader
 from nemos.callbacks import Callback
 from nemos.glm.params import GLMParams
--- a/tests/test_partial_param_spec.py
+++ b/tests/test_partial_param_spec.py
@@ -1,7 +1,4 @@
-"""Tests for the partial parameter specification (``fix_params``) machinery.
-
-Tree algebra only: no optimization runs here.
-"""
+"""Tests for the partial parameter specification (``fix_params``) machinery."""
 
 import equinox as eqx
 import jax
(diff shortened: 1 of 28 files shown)


tests/test_partial_param_spec.py
```diff
@@ -0,0 +1,611 @@
...
+    public surface catches that when the interface grows, e.g. a Hessian mixin.
+    """
+    expected = {
+        name
+        for name in dir(AbstractSolver)
+        if not name.startswith("_") and callable(getattr(AbstractSolver, name, None))
+    }
+
+    missing = sorted(name for name in expected if not hasattr(NoOpSolver, name))
+    assert missing == []
```
wulfdewolf: feels like it should be in a different test file

wulfdewolf (commented):
LGTM!

Issue #580, opened 2026-06-10
Newton solvers: `newton-cholesky` (PD/PSD) and `newton` (general) — one modified-Hessian Newton step


BalzaniEdoardo:
## Objective

Make Newton maximally efficient for GLMs while ensuring convergence by following stable algorithmic paths.

We will need two algorithms: Newton and Newton-Cholesky solvers compute the same kind of step — a Newton step $\Delta x = -B^{-1} g$ where $B$ is a positive-definite surrogate of the Hessian $H$, obtained by a linear solve and globalized by line search (Nocedal & Wright, *Numerical Optimization* 2nd ed., §3.4, Algorithm 3.2).

They differ only in how $B$ is formed. `Newton-Cholesky` uses $B = H + \tau I$ with a closed-form $\tau$ resolved from the certified `HessianTag.property` (the cases where the needed shift is known a priori: $\tau=0$ for PD, $\tau=\varepsilon$ for PSD).

`Newton` forms $B$ with the most efficient in-family modification for the general/indefinite case (candidates below), still a linear solve, still line-search globalized.

## Current state and why it needs changing

The existing `src/nemos/solvers/_newton.py` selects `lx.Cholesky()` **only** when the tag is `PositiveDefinite`; every other property (PSD, indefinite, …) falls into the `AutoLinearSolver(well_posed=False)` (SVD) branch. Problems:

1. On an indefinite Hessian the SVD/pseudoinverse step $H^{+}(-g)$ is not guaranteed to be a descent direction ($g^\top p = -\sum_i (1/\lambda_i)(q_i^\top g)^2$ flips sign for negative $\lambda_i$); the `descent < 0` guard in `_apply_or_reject` then rejects the step, so iterations stall at `maxiter` with no progress and no error — confirmed by the N&W example $\nabla^2 f = \mathrm{diag}(10,3,-1)$, $g = (1,-3,2)$ → $p^{\mathrm{N}} = (-0.1,1,2)$, not a descent direction.
2. There is no fast PSD path: a `PositiveSemiDefinite` Hessian, which a single shifted Cholesky would solve cheaply, instead pays for the full SVD. (The PD branch annotates its operator `positive_semidefinite_tag` — a harmless mislabel, since `lx.Cholesky()` is chosen explicitly and is never handed a non-PD matrix.)
3. The `Diagonal` structure tag is computed but never used — diagonal systems pay $O(n^3)$ instead of $O(n)$.
[...]

Issue #20, opened 2023-06-22
Project Structure


BalzaniEdoardo:
Define sub-modules, for example basis.py can go in a separate sub-module

```
generalized-linear-model
├── src                     ┐
│   └── neurostatslib       │
│       ├── __init__.py     │ 
│       ├── glm.py          │   Package source code
│       └── utils.py        |
|       └──── basis         |
|.         └──__init__ .py  |
|.         └── basis.py     ┘
└── tests                   ┐
|  └── ...                  ┘ Package tests
|
...
```

PR #612, opened 2026-08-14
Newton solver generalisation

wulfdewolf:
## Summary
This PR generalises the `Newton` solver into multiple solving branches based on Hessian properties:

- [x] `PositiveDefinite` → Cholesky
- [x] `PositiveSemiDefinite` → Cholesky with Tikhonov damping
- [x] Otherwise → modified Hessian using eigendecomposition (`"eigh"`) or an adaptive identity shift (`"identity_shift"`)

Users can select the strategy through the `hessian_solver` argument:

- `"auto"` selects a strategy from the Hessian tag
- `"cholesky"` forces Cholesky
- `"eigh"` applies eigenvalue modification
- `"identity_shift"` increases a diagonal shift until Cholesky succeeds

Solver-specific arguments:

- `shift_const` controls damping for the positive-semidefinite Cholesky branch
- `identity_shift_beta` controls the minimum shift for `"identity_shift"`

A forced Cholesky strategy with an incompatible Hessian tag emits a warning but still runs.

## Testing
- Positive-definite Hessians:
  - Quadratic problems converge in one Newton step (reported as two due to the Cauchy termination criterion)
  - Exact solution recovery
  - Scale equivariance
  - Analytic and autodiff Hessians
- Positive-semidefinite Hessians:
  - Singular quadratics preserve the initial null-space component
  - Scale equivariance
- Modified-Hessian strategies:
  - Produce finite descent directions
  - Match standard Newton on positive-definite Hessians
  - Support block-diagonal and PyTree Hessians
  - Converge from indefinite regions on quartic and Rosenbrock problems
- Strategy-specific behavior:
  - Eigenvalue modification and flooring
  - Initial identity shift and tenfold Cholesky retries
  - Automatic strategy resolution
  - User overrides, warnings, and invalid arguments

## Documentation
TODO

## Questions for reviewers
Should users be able to select the positive-definite versus damped positive-semidefinite Cholesky approach explicitly, or should the Hessian tag continue to select between them?

## Related
#580


wulfdewolf:
@BalzaniEdoardo so I'm a bit stuck on the scale-equivariance test for PSD.
If you look at `test_newton_cholesky.py::test_psd_quadratic_scale_equivariance`, it fails:
- for `1e-6` because the gradient norm becomes tiny, and `NewtonCholesky` immediately assumes convergence, never taking a step
- for `1e6` because the gradient norm becomes huge, and `NewtonCholesky` doesn't converge (I tried with up to 1000 iterations)

Not sure whether this is an issue or not, in practice. 
The solver itself is scale-equivariant, but the convergence criterion is not?

wulfdewolf:
@BalzaniEdoardo then for the more general `Newton` solver I am still thinking about things. 
As noted in the PR summary, `lineax` raises for indefinite problems, so a retry approach becomes hard with jitted code.
For the eigenvalue modification approach we will have to think about a nice way to flatten our pytree hessians, because `jnp.linalg.eigh` expects a dense matrix.
This is less urgent, I need to think about it more.

tests/test_newton_solver.py
```diff
@@ -0,0 +1,627 @@
...
+def _quadratic_loss_and_hessian(A, b):
+    """f(x) = 1/2 x'Ax - b'x."""
+
+    def loss(params, *args):
+        return _quadratic_loss(params, *args)
+
+    def hess(params, *args):
+        return A
+
+    return loss, hess
```
BalzaniEdoardo, line 98:
```suggestion
  def _quadratic_loss_and_hessian(A, b):                                                                                                                              
      """f(x) = 1/2 x'Ax - b'x, paired with its analytic Hessian."""                                                                                                  
                                                                                                                                                                      
      def hess(params, *args):                                                                                                                                        
          return A                                                                                                                                                    
                                                                                                                                                                      
      return _quadratic_loss(A, b), hess                                                                                                                              
                                             
```

The signature of `_quadratic_loss` was off.

BalzaniEdoardo: with this helper fix the only case failing is the 1e-6 case, with exactly the pathology you identify (one step and stop). And the reason is what you flagged, so the plan for the stopping criteria was off

BalzaniEdoardo:
@wulfdewolf here is my take on this:

### 1. Convergence criterion: use `cauchy_termination`

`‖g‖ ≤ tol` cannot be scale-invariant, you are correct on your assessment.

The PD path has the same problem, that was hidden by a bug in the test helper: `_quadratic_loss_and_hessian` calls `_quadratic_loss(params)` instead of `_quadratic_loss(A, b)(params)`, so 13 of the 16 failures are a `TypeError`. See the comment in the code. Once fixed, `test_pd_quadratic_scale_equivariance[1e-06-float32]` fails with `steps=0`.

`cauchy_termination` is already used in `_fista.py:180` and wrapped as `_params_only_cauchy_criterion` in `_stochastic_mixins.py:21`. The `prox-newton` branch already has the pieces:

- `_converged(self, params, state, grad, fval)` as an overridable hook: `Newton` implements `‖g‖ ≤ tol`, `ProximalNewton` overrides it with Cauchy.
- `NewtonState.y_diff`, "initialized to inf so the first iteration never counts as converged". That removes the `steps=0` failure.
- `ProximalNewton._converged`'s docstring gives this failure mode as its reason for dropping a gradient test: "on badly conditioned data it never falls below `tol` even once the iterate has stopped moving".

Running `NewtonCholesky`'s step machinery with only the convergence test replaced by Cauchy, all 24 combinations of {PD, PSD} × {1e-6, 1, 1e6} × {f32, f64} × {rtol=0, 1e-6} converge to precision (no failures).

The Cauchy comes with the following costs:

- `steps` goes from 1 to 2 on a quadratic. The optimum is still reached on the first step, but two iterates are needed before they can be compared.
- `_converged_fn(lambda_sq, gnorm)` cannot express Cauchy since it does not receive the params, the step, or the previous `fval`. Look at what I am doing in the `prox-newton`'s branch in term of the convergence etc. Consider if you want to branch of it (or merge that branch into this one), because it already defines an HessianMixin etc.

This removes the PD/PSD criterion split from #580 instead of repairing it. I think `λ²` is still worth reporting, since on the PD path it estimates `f - f*` and is affine-invariant. Up to you, it may go in the solver stats.

### 2. Detecting indefiniteness: lineax is not what blocks the retry

lineax does not raise on an indefinite matrix at runtime. `Cholesky.init` raises only when the operator carries neither `positive_semidefinite_tag` nor `negative_semidefinite_tag`, a check on the tag at trace time, not on the values. Once the operator is tagged PSD, which `_solve` already does via `self._operator_tags`, `init` calls `jsp.linalg.cho_factor` and `compute` returns `RESULTS.successful` every time without looking at the factor. An indefinite Hessian gives NaNs and no exception, and `throw=False` cannot help because no failure is reported.

So the N&W Algorithm 3.3 retry loop can be jitted: detect the failure with `jnp.isnan` on the factor, by calling `jax.scipy.linalg.cho_factor` directly inside an `eqx.internal.while_loop` instead of going through `lx.linear_solve`.

**Other Comments:**

- `diverged` is computed in `update` but is not part of `OptimizationInfo` (`function_val`, `num_steps`, `converged`, `reached_max_steps`), so it stops the loop and is never reported. Either surface it or drop it.
- `eigh` needs no new flattening code. lineax already flattens on every Cholesky solve this branch does: `init` calls `operator.as_matrix()`, `compute` ravels the right-hand side with `jfu.ravel_pytree` and unravels the result. `lx.PyTreeLinearOperator(H, out_struct).as_matrix()` plus `ravel_pytree(grad)` gives the dense pair. `BlockDiagonal` is unaffected: `_newton_direction` already vmaps `_solve` per block, so the densification happens inside one block and `eigh` vmaps over blocks.

BalzaniEdoardo (commented):
My high level concern is the re-evaluation of needing 2 distinct classes.

The two differ in one method (`_solve`); #580 says as much ("`Newton` is `Newton-Cholesky` with a different solve step"). The tag already decides which step to take, and with `tagging-system-link-dependent` it decides reliably, so the switch belongs inside one solver: PD/PSD gets the shifted Cholesky, everything else gets the modified Hessian. That is one more case where `init_state` already resolves `_shift_fn`, not a second class.

The split as committed requires a user-facing rename with no deprecation, plus the registry, `__init__.pyi`, the `list_available_algorithms` doctest, the `conftest` solver fixture, the docstring tables in `glm.py`, and the `isinstance(solver, NewtonCholesky)` check in `base_regressor`. None of that buys anything algorithmically.

If you merge in prox-newton or cherry pick the mixin machinery and the convergence handling:

- Use `HessianMixin` from `prox-newton` (`_hessian_mixins.py`). Its `_uses_hessian: ClassVar[bool] = True` is the capability flag `BaseRegressor` should check, mirroring `_supports_stochastic`, instead of `isinstance(solver, NewtonCholesky)`.
- Keep `Newton` as the only registered name. 
- Report which step was taken in the solver stats. A Cholesky and an `eigh` are different costs per iteration, and the user should be able to see which one they paid for rather than infer it from the tag.

**Question:**

Do we want a mechanism for the user to force an algo vs another (bypassing tags)?

docs/assets/ashwood_targets.svg
BalzaniEdoardo, line 1: why?

BalzaniEdoardo:
Here a proposed list of tests for the general case:


- **A positive definite matrix behaves exactly as before.** With all eigenvalues well above the floor, the general path gives the same step as the Cholesky path and solves a quadratic in one step.
- **All three tags agree.** `General`, `Symmetric` and `NegativeDefinite` go down the same branch, so they must produce identical steps on the same problem.
- **The step matches an independent reference.** Compare against a step computed with `scipy.linalg.sqrtm(H @ H)`, which returns the same matrix but via a different algorithm than the eigendecomposition in the source.
- **The sign of the curvature does not matter.** `H` and `−H` give the same step, and so does flipping the sign of any subset of eigenvalues. Enumerate all 16 sign patterns for a 4×4.
- **The step always points downhill.** Over many random spectra — positive, negative, mixed, singular — check the step goes downhill, and that its slope falls inside the two bounds the spectrum predicts.
- **The floor kicks in at 1e-6.** With a zero eigenvalue and a gradient component along it, that part of the step is the gradient component divided by `1e-6`. Sweep eigenvalues just below, at, and just above the floor.
- **It escapes a saddle.** On `f(x) = x⁴/4 − x²/2`, started in the concave middle, the solver leaves the saddle and lands on ±1. Plain Newton would sit on the saddle.
- **It solves Rosenbrock.** Starting far out, where the Hessian is indefinite, it reaches `(1, 1)`. Cross-check with `scipy.optimize.minimize`.
- **It works block by block.** With a block-diagonal Hessian whose blocks are indefinite, the result matches solving each block separately in numpy.
- **It works on pytree parameters.** With parameters split across a tree, the step matches the reference. This is what catches the flatten/unflatten order getting out of sync with the matrix.
- **Same answer with jit on and off.**


for the Rosenbock test, limit the number of steps. A quadratic solver should take a relatively small number of steps to solve it.

src/nemos/solvers/_newton.py
```diff
@@ -105,44 +131,112 @@ def init_state(self, init_params, *args):
...
+                iteration, tau, _ = carry
+                new_tau = jnp.maximum(10.0 * tau, beta)
+                return iteration + 1, new_tau, factorize(new_tau)
+
+            _, _, factor = eqx.internal.while_loop(
+                cond,
+                body,
+                (jnp.asarray(0), tau0, factor0),
+                kind="lax",
+            )
```
BalzaniEdoardo: I wonder if you can do all of this in lineax:



- getting the diagonal and composed operator
```
diag = lx.diagonal(operator)
min_diag = jnp.min(diag)
tau0 = jnp.where(
                min_diag > 0,
                jnp.zeros((), dtype=dtype),
                -min_diag + beta,
            )
shift = lx.DiagonalLinearOperator(jnp.full(diag.shape[0], tau0))
opearator = hessian_op + shift

```

- Iterative cholesky decomposition:
```python
# Pre-compute Cholesky factorization
state = lx.Cholesky().init(operator)

# Reuse it for multiple solves efficiently
sol1 = lx.linear_solve(operator, vector1, solver=lx.Cholesky(), state=state)
sol2 = lx.linear_solve(operator, vector2, solver=lx.Cholesky(), state=state)
```


In this way we never materialize the operator. it will come in handy when I'll make sure that the model return directly a lineax operator.

src/nemos/solvers/_newton.py
```diff
@@ -105,44 +131,112 @@ def init_state(self, init_params, *args):
...
+        # Modify the Hessian eigenvalues, then solve the resulting positive-definite system
+        # This handles symmetric indefinite Hessians in one decomposition
+        if self._resolved_hessian_solver == "eigh":
+            g_flat, unravel = jax.flatten_util.ravel_pytree(grad)
+            H_dense = operator.as_matrix()
+
+            eigvals, Q = jnp.linalg.eigh(H_dense)
+            lam_mod = jnp.maximum(jnp.abs(eigvals), self._delta)
+            direction = Q @ ((Q.T @ (-g_flat)) / lam_mod)
+            return unravel(direction)
```
BalzaniEdoardo: isn't this handled by

```
lx.SVD(rcond=self._delta)?
```

BalzaniEdoardo (changes requested):
overall this looks correct, I think we should not leave lineax if we can

src/nemos/solvers/_newton.py
wulfdewolf: So, I can definitely use more `lineax`, but it adds other complexity:
```
            diag = lx.diagonal(operator)
            dtype = diag.dtype

            # Floor for beta
            eps = jnp.asarray(jnp.finfo(dtype).eps, dtype=dtype)
            beta = jnp.maximum(
                jnp.asarray(self._identity_shift_beta, dtype=dtype),
                eps,
            )

            min_diag = jnp.min(diag)
            tau0 = jnp.where(
                min_diag > 0,
                jnp.zeros((), dtype=dtype),
                -min_diag + beta,
            )

            cholesky = lx.Cholesky()
            identity = lx.IdentityLinearOperator(operator.in_structure())
            neg_grad = jax.tree.map(jnp.negative, grad)

            def _shifted_operator(tau):
                shifted = operator + tau * identity
                return lx.TaggedLinearOperator(
                    shifted,
                    lx.positive_semidefinite_tag,
                )

            def _solve(tau):
                result = lx.linear_solve(
                    _shifted_operator(tau),
                    neg_grad,
                    solver=cholesky,
                    throw=False,
                )
                failed = (
                    result.result != lx.RESULTS.successful
                ) | ~tree_utils.tree_all_finite(result.value)
                return result.value, failed

            direction0, failed0 = _solve(tau0)

            def cond(carry):
                iteration, _, _, failed = carry
                return failed & (iteration < self._identity_shift_max_steps)

            def body(carry):
                iteration, tau, _, _ = carry
                new_tau = jnp.maximum(
                    jnp.asarray(10.0, dtype=dtype) * tau,
                    beta,
                )
                direction, failed = _solve(new_tau)
                return iteration + 1, new_tau, direction, failed

            _, _, direction, _ = eqx.internal.while_loop(
                cond,
                body,
                (
                    jnp.asarray(0),
                    tau0,
                    direction0,
                    failed0,
                ),
                kind="lax",
            )

            return direction
```

- we need the finite check in tree space (I added a tree util function)
- we tree map for -grad

I think that's fine.

I don't think we can initialise the state and reuse it like in your example.
The state is dependent on the operator, lineax requires the operator to stay the same if you pass the state:
https://docs.kidger.site/lineax/api/linear_solve/#lineax.linear_solve
In our case the operator changes (tau).

src/nemos/solvers/_newton.py
wulfdewolf: It could be, but I'm not sure.

From my understanding of the docs I think that `rcond` drops the eigenvalues below self._delta, it does not replace them by their absolute value, nor floors them to delta like I was doing here:
```
lam_mod = jnp.maximum(jnp.abs(eigvals), self._delta)
```
The dropping might be fine (?) but the `abs` is necessary no?

BalzaniEdoardo: fair, then we keep this part as is!

src/nemos/solvers/_newton.py
BalzaniEdoardo: make sense, the repeated solve is for when you don't change the matrix,and you solve for different output vec.

BalzaniEdoardo: if you want to inspect the cholevsky decomposition before calling solve, you can grab it from the state

src/nemos/solvers/_newton.py
```diff
@@ -45,6 +99,10 @@ def __init__(
         maxiter: int = DEFAULT_MAX_STEPS,
         tol: float = DEFAULT_ATOL,
         rtol: float = DEFAULT_RTOL,
+        shift_const: float = 1.0,
```
BalzaniEdoardo: why defaulting to 1?

src/nemos/solvers/_newton.py
```diff
@@ -33,6 +33,60 @@ class NewtonState(eqx.Module):
...
+    throw :
+        Whether Lineax should raise an exception when the solve fails.
+
+    Returns
+    -------
+    :
+        The result returned by :func:`lineax.linear_solve`.
+    """
+    identity = lx.IdentityLinearOperator(operator.in_structure())
+    shifted_operator = operator + shift * identity
```
BalzaniEdoardo: Shifting for PD system doesn't make sense: we can save a vec * const multiplication and adding the diagonal to the operator. see if this is possible.
I was the one proposing the shift = 0 computing in the same way but I think we should optimize for speed. So I would say, try to route statically based on if the shift happens or not.

src/nemos/solvers/_newton.py
```diff
@@ -45,6 +99,10 @@ def __init__(
         maxiter: int = DEFAULT_MAX_STEPS,
         tol: float = DEFAULT_ATOL,
         rtol: float = DEFAULT_RTOL,
+        shift_const: float = 1.0,
+        identity_shift_beta: float = 0.0,
+        identity_shift_max_steps: int = 20,
+        hessian_solver: HessianSolver = "auto",
```
BalzaniEdoardo: rename to `linear_solver`. The fact that we solve the hessian is true, but the name should reflect the solver not the specific application of it in my opinion

src/nemos/solvers/_newton.py
```diff
@@ -105,44 +184,124 @@ def init_state(self, init_params, *args):
...
-            tags=self._operator_tags,
-        )
 
-        return lx.linear_solve(
+        operator = lx.PyTreeLinearOperator(H, jax.eval_shape(lambda: grad))
+
+        # Modify the Hessian eigenvalues, then solve the resulting positive-definite system
+        # This handles symmetric indefinite Hessians in one decomposition
+        if self._resolved_hessian_solver == "eigh":
+            g_flat, unravel = jax.flatten_util.ravel_pytree(grad)
```
BalzaniEdoardo:
```python
import jax
jax.flatten_util.ravel_pytree
```

raises, it is not exposed publicly. you get it to pass because lx imports flaten utils.

the correct way to use this is:

```python
import jax.flatten_utils as jfu
```
Or directly importing rave_pytree

src/nemos/tree_utils.py
```diff
@@ -7,7 +7,6 @@
 import jax
 import jax.numpy as jnp
 from jax.core import Tracer
-from jax.flatten_util import ravel_pytree
```
BalzaniEdoardo, line 10: also not good to remove this. you can't just import jax and call flatten utils. it may currently work as a side effect of importing optimistix and lineax

src/nemos/solvers/_newton.py
```diff
@@ -89,6 +149,25 @@ def __init__(
 
         # Cache
         self._gradient: Callable | None = None
+        self._hessian: Callable | None = None
+
+        self._linear_solver = lx.Cholesky()
```
BalzaniEdoardo: this gets selected in the mixin, so set it to None at this point

src/nemos/solvers/_newton.py
```diff
@@ -105,44 +184,124 @@ def init_state(self, init_params, *args):
...
+        # Catch use before init_state has resolved the requested strategy.
+        if self._resolved_hessian_solver != "cholesky":
+            raise RuntimeError(
+                "The Hessian solver has not been resolved. "
+                "Call init_state before update."
+            )
+
+        # Solve a positive Hessian with Lineax Cholesky.
+        # For a PSD tag, _shift_fn adds the small numerical shift selected by the tag
+        return _solve_shifted_system(
```
BalzaniEdoardo, line 288: on a requested cholesky, this throws if non-pd, not just warn. I think this is the right behavior. If I understand correctly the summary says the solver proceeds after warning, which is not what happens.

src/nemos/solvers/_hessian_mixins.py
```diff
@@ -139,12 +164,74 @@ def _resolve_linear_solver(self, init_params) -> None:
...
+                    operator.as_matrix(),
+                    self._shift_const,
+                )
+            else:
+                self._shift_fn = lambda _: 0.0
+        elif resolved == "eigh":
+            self._linear_solver = None
+            self._operator_tags = ()
+            self._shift_fn = lambda _: 0.0
+            self._delta = 1e-6
```
BalzaniEdoardo:
```suggestion
            dtype = jnp.result_type(*jax.tree_util.tree_leaves(init_params))
            self._delta = jnp.sqrt(jnp.finfo(dtype).eps)
```

Note: `_EIGH` in `tests/test_newton_glm.py` asserts `"_delta": 1e-6` and will need updating together with this.

BalzaniEdoardo: so that this is adaptive to machine precision

src/nemos/solvers/_newton.py
```diff
@@ -45,6 +99,10 @@ def __init__(
         maxiter: int = DEFAULT_MAX_STEPS,
         tol: float = DEFAULT_ATOL,
         rtol: float = DEFAULT_RTOL,
+        shift_const: float = 1.0,
+        identity_shift_beta: float = 0.0,
```
BalzaniEdoardo, line 104: N&W give the value for this on p. 52: "The choice of β is heuristic; a typical value is β = 10⁻³." The default of `0.0` gets floored at `finfo(dtype).eps`, which is far below that and makes the retry loop do a lot of extra work. 

```suggestion
        identity_shift_beta: float = 1e-3,
```
**But 1e-3 must be applied relative to the matrix, not absolutely** — see the `_solve` comment on line 228 below. I checked this after writing the paragraph above, and a plain absolute 1e-3 violates the scale equivariance assumptions.

src/nemos/solvers/_newton.py
```diff
@@ -105,44 +184,124 @@ def init_state(self, init_params, *args):
...
+        # Solve a positive Hessian with Lineax Cholesky.
+        # For a PSD tag, _shift_fn adds the small numerical shift selected by the tag
+        return _solve_shifted_system(
             operator,
-            jax.tree.map(lambda x: -x, grad),
-            self._linear_solver,
+            grad,
+            self._shift_fn(operator),
+            solver=self._linear_solver,
+            tags=self._operator_tags,
```
BalzaniEdoardo, line 293: here if you do want to warn and not throw when the user forces cholesky, 


```suggestion
            tags=self._operator_tags,
            throw=self._cholesky_certified,
```
and then in the mixin you can set a flag that testifies if the pd was certified internally by nemos, and in that case the failure should be loud, or by a user. I think independently a failure should just throw, so that the user knows things are not working

BalzaniEdoardo: self._cholesky_certified this should be true if PD certificate comes from us, false otherwise.

tests/test_newton_glm.py
BalzaniEdoardo, line 144:
```suggestion
]


def test_every_block_diagonal_model_has_fixtures():
    """Registry and discovery must agree, so no block-diagonal model goes unchecked.

    ``test_newton_block_diagonal_matches_full_autodiff_update`` is parametrized from
    ``_BLOCK_MODEL_FIXTURES``, so a model missing from it would be skipped rather than fail.
    This test is what turns that silence into a failure.
    """
    discovered = {cls.__name__ for cls in _block_diagonal_models()}
    registered = {cls.__name__ for cls in _BLOCK_MODEL_FIXTURES}
    assert discovered == registered, (
        f"declare a block-diagonal Hessian but are absent from _BLOCK_MODEL_FIXTURES, so "
        f"they are never checked against the full Hessian: {sorted(discovered - registered)}. "
        f"Registered but no longer block-diagonal: {sorted(registered - discovered)}."
    )
```

tests/test_newton_glm.py
```diff
@@ -1235,104 +1157,147 @@ def _installed_newton(model, X, y):
...
+        _CHOLESKY_PSD,
+        id="ClassifierGLM-UnRegularized-PSD",
+    ),
+    pytest.param(
+        "population_classifierGLM_model_instantiation",
+        "UnRegularized",
+        _CHOLESKY_PSD,
+        id="ClassifierPopulationGLM-UnRegularized-PSD",
+    ),
 ]
```
BalzaniEdoardo, line 1284: All eight cases here resolve to `cholesky`, and `_HESSIAN_SOLVER_CASES` only covers Ridge and UnRegularized. Nothing in the suite fits a GLM through `eigh` or `identity_shift` end to end, so the four new public `solver_kwargs` (`hessian_solver`, `shift_const`, `identity_shift_beta`, `identity_shift_max_steps`) are never checked against the argument validation, and the three strategies are never checked against each other on real data.

I ran it by hand on a Poisson GLM and they agree, which is the assertion worth having:

```
{'hessian_solver': 'eigh'}            -> coef=[0.0103 -0.1685 0.2490]
{'hessian_solver': 'identity_shift'}  -> coef=[0.0103 -0.1685 0.2490]
{'hessian_solver': 'cholesky'}        -> coef=[0.0103 -0.1685 0.2490]
{'hessian_solver': 'bogus'}           -> ValueError
```

```suggestion
]


@pytest.mark.requires_x64
@pytest.mark.parametrize("hessian_solver", ["auto", "cholesky", "eigh", "identity_shift"])
def test_every_hessian_solver_reaches_the_same_fit(
    poissonGLM_model_instantiation, hessian_solver
):
    """The strategies differ in cost per step, not in the optimum they reach."""
    X, y, model, *_ = poissonGLM_model_instantiation
    model.regularizer = "Ridge"
    model.regularizer_strength = 0.1
    model.solver_name = "Newton"

    model.solver_kwargs = {"hessian_solver": "auto"}
    reference = model.fit(X, y).coef_

    model.solver_kwargs = {"hessian_solver": hessian_solver}
    np.testing.assert_allclose(model.fit(X, y).coef_, reference, atol=1e-5)
```

Also worth pairing `_HESSIAN_SOLVER_CASES` with a coverage guard, since it is hand-written with a different reference per entry.

tests/test_newton_solver.py
```diff
@@ -0,0 +1,1193 @@
...
+from nemos.tree_utils import pytree_map_and_reduce
+
+N = 8
+
+# The two second-order solvers. Everything ``Newton`` converges to, ``ProximalNewton``
+# converges to as well: they differ in how the penalty is reached (a proximal operator
+# rather than the penalized loss) and in the convergence test, not in the optimum. So the
+# solver-agnostic tests below run for both, and only the genuinely divergent behaviour gets
+# a dedicated test.
+_NEWTON_SOLVERS = ("Newton", "ProximalNewton")
```
BalzaniEdoardo, line 31: The comment says "the solver-agnostic tests below run for both", but they don't. `_SOLVERS` (line #32) and `_SOLVER_CLASSES` (line #34) have zero uses in this file — every test builds `Newton` directly through `_make_solver` — and the `ProximalNewton` import on line #20 is unused too.

tests/test_newton_solver.py
```diff
@@ -0,0 +1,1193 @@
...
+        H,
+        grad,
+        tag,
+        jit=jit,
+        hessian_solver=hessian_solver,
+        identity_shift_beta=1e-3,
+    )
+
+    assert bool(jnp.all(jnp.isfinite(direction)))
+    assert float(jnp.vdot(grad, direction)) < 0.0
```
BalzaniEdoardo, line 646: This checks only the sign of the slope, so a step that points downhill but is wildly wrong in magnitude passes. That matters most for the `[1e-8, -1e-7, 1.0, -3.0]` case in the parametrization, where the eigenvalues sit below the floor and the step length is set entirely by δ — which is the value I'm suggesting we change.

For `eigh` the modified spectrum is known exactly, so the slope is bracketed and cheap to assert:

$$g^\top d = -\sum_i \frac{(q_i^\top g)^2}{\hat\lambda_i} \quad\Longrightarrow\quad -\frac{\|g\|^2}{\hat\lambda_{\min}} \le g^\top d \le -\frac{\|g\|^2}{\hat\lambda_{\max}}$$

```suggestion
    assert bool(jnp.all(jnp.isfinite(direction)))
    slope = float(jnp.vdot(grad, direction))
    assert slope < 0.0

    # For eigh the modified spectrum is known, so the slope is bracketed by it.
    if hessian_solver == "eigh":
        lam = np.maximum(np.abs(np.asarray(eigenvalues)), _EIGH_DELTA)
        grad_sq = float(jnp.vdot(grad, grad))
        assert -grad_sq / lam.min() <= slope <= -grad_sq / lam.max()
```

This needs `_EIGH_DELTA` as one module-level constant shared with src, rather than `1e-6` written out in three places (here, `test_eigh_direction_applies_eigenvalue_floor`, and `_EIGH` in `test_newton_glm.py`).

tests/test_newton_solver.py
```diff
@@ -0,0 +1,1193 @@
...
+        shift_const=0.0,
+        jit=jit,
+    )
+
+    x_opt, state, _ = solver.run(x0)
+
+    assert bool(state.stats.converged)
+    assert state.stats.num_steps == 2
+
+    np.testing.assert_allclose(x_opt, x_star, atol=solver.tol, rtol=solver.rtol)
```
BalzaniEdoardo, line 171: `solver.tol` is 1e-4 and `solver.rtol` is 0, but Newton reaches the optimum of a quadratic on the first step, so the answer is exact to machine precision. Across all six `(dtype, scale)` combinations:

```
float64   steps=2   max|x - x*| = 2.2e-16 .. 4.4e-16    (asserted: 1e-4)
float32   steps=2   max|x - x*| = 6.0e-08 .. 1.2e-07    (asserted: 1e-4)
```

That is about 11 orders of margin in float64. 

Since `x_star` has no near-zero components here (the smallest is 0.249), this wants a relative tolerance and no `atol` at all:

```suggestion
    np.testing.assert_allclose(x_opt, x_star, rtol=20 * np.finfo(dtype).eps, atol=0.0)
```

A similar check should be applied to all sites in which we are currently using the solver parameters instead. we should be more conservative when we know the solution needs  to be extremely close to optimal

tests/test_newton_solver.py
```diff
@@ -0,0 +1,1193 @@
...
+        hessian_solver="cholesky",
+    )
+
+    with pytest.warns(
+        RuntimeWarning,
+        match="Cholesky generally requires",
+    ):
+        solver.init_state(params)
+
+    assert solver._resolved_hessian_solver == "cholesky"
```
BalzaniEdoardo: Add a case with a genuinely indefinite Hessian pins the documented behaviour — warns, runs, reports failure (if that's the behavior we choose, i am fine throwing more often):

```suggestion
    assert solver._resolved_hessian_solver == "cholesky"


@pytest.mark.requires_x64
def test_forced_cholesky_on_indefinite_hessian_does_not_raise():
    """A forced Cholesky warns and reports failure rather than raising."""
    H = jnp.asarray([[1.0, 3.0], [3.0, 1.0]])  # eigenvalues 4 and -2
    params = jnp.asarray([0.3, -0.1])

    solver = _make_solver(
        _quadratic_loss(H, jnp.asarray([1.0, 0.5])),
        lambda params, *args: H,
        _make_tag(params, property=MatrixProperty.SYMMETRIC),
        params,
        hessian_solver="cholesky",
    )

    with pytest.warns(RuntimeWarning, match="Cholesky generally requires"):
        _, state, _ = solver.run(params)

    assert not bool(state.stats.converged)
```

tests/test_newton_solver.py
```diff
@@ -0,0 +1,1193 @@
...
+    assert state.stats.num_steps == 2
+
+    np.testing.assert_allclose(x_opt, x_star, atol=solver.tol, rtol=solver.rtol)
+
+
+@pytest.mark.requires_x64
+@pytest.mark.parametrize("dtype", [jnp.float32, jnp.float64])
+@pytest.mark.parametrize("scale", [1e-6, 1.0, 1e6])
+@pytest.mark.parametrize("jit", [False, True])
+def test_psd_quadratic_scale_equivariance(dtype, scale, jit):
```
BalzaniEdoardo, line 350: Both scale-equivariance tests sit on the Cholesky path. Neither `eigh` nor `identity_shift` is tested under rescaling, which is what hides the fixed-δ issue.

Scaling the objective by `s` sends `g -> s*g` and `H -> s*H`, so the exact Newton step is unchanged — the `s` cancels between numerator and denominator. The floor breaks that cancellation: `max(|s*lam|, delta)` stops depending on `s` once the eigenvalue falls below δ, while the numerator keeps its `s`. In plain numpy:

```ipython
>>> import numpy as np
>>> H = np.array([[1.0, 3.0], [3.0, 1.0]])  # eigenvalues 4 and -2
>>> g = np.array([1.0, 0.5])
>>> delta = 1e-6
>>> def eigh_direction(s):
...     lam, Q = np.linalg.eigh(s * H)
...     return Q @ ((Q.T @ (-s * g)) / np.maximum(np.abs(lam), delta))
>>> 
>>> 
>>> ref = eigh_direction(1.0)
>>> for s in (1e2, 1.0, 1e-4, 1e-6, 1e-8, 1e-10):
>>>     d = eigh_direction(s)
>>>     lam_min = np.abs(np.linalg.eigvalsh(s * H)).min()
>>>     print(
>>>         f"s={s:7.0e}  min|lam|={lam_min:8.1e}  floored={lam_min < delta!s:5s}"
>>>         f"  d={np.round(d, 6)}  rel.dev={np.linalg.norm(d - ref) / np.linalg.norm(ref):.3f}"
>>>     )
s=  1e+02  min|lam|= 2.0e+02  floored=False  d=[-0.3125 -0.0625]  rel.dev=0.000
s=  1e+00  min|lam|= 2.0e+00  floored=False  d=[-0.3125 -0.0625]  rel.dev=0.000
s=  1e-04  min|lam|= 2.0e-04  floored=False  d=[-0.3125 -0.0625]  rel.dev=0.000
s=  1e-06  min|lam|= 2.0e-06  floored=False  d=[-0.3125 -0.0625]  rel.dev=0.000
s=  1e-08  min|lam|= 2.0e-08  floored=True   d=[-0.01   -0.005 ]  rel.dev=0.966
s=  1e-10  min|lam|= 2.0e-10  floored=True   d=[-1.e-04 -5.e-05]  rel.dev=1.000
```


Exact while the eigenvalues clear δ, then the step collapses to `-(s/delta) * g`: a gradient descent step of fixed length, curvature discarded. It still points downhill, which is why the sign-only descent test on line #651 does not notice.

The test should assert the crossover rather than invariance everywhere, since both sides have closed forms — equal to the `s=1` step while `min|s*lam| >= delta`, and equal to `-(s/delta) * q_i (q_i.g)` in the floored directions once below.

src/nemos/solvers/_newton.py
```diff
@@ -105,44 +184,124 @@ def init_state(self, init_params, *args):
...
+        if self._resolved_hessian_solver == "identity_shift":
+            diag = lx.diagonal(operator)
+            dtype = diag.dtype
+
+            # Floor for beta
+            eps = jnp.asarray(jnp.finfo(dtype).eps, dtype=dtype)
+            beta = jnp.maximum(
+                jnp.asarray(self._identity_shift_beta, dtype=dtype),
+                eps,
+            )
```
BalzaniEdoardo, line 226: Paired with line 103. Making β relative to the matrix is what restores scale-equivariance;

```suggestion
            # Beta scales with the matrix so the shift ladder is scale-equivariant;
            # N&W p.52 give 1e-3 as the typical magnitude, here relative to the diagonal.
            eps = jnp.asarray(jnp.finfo(dtype).eps, dtype=dtype)
            beta = jnp.maximum(
                jnp.asarray(self._identity_shift_beta, dtype=dtype)
                * jnp.max(jnp.abs(diag)),
                eps,
            )
```

This changes one existing test's expected value, which needs adapting rather than working around. I had claude checking all three:

| test | max\|diag\| | τ absolute | τ relative | status |
| --- | --- | --- | --- | --- |
| `test_identity_shift_uses_initial_diagonal_shift` | 3.0 | 2.25 | 2.75 | **expected value changes** |
| `test_identity_shift_retries_until_cholesky_succeeds` | 1.0 | 2.00 | 2.00 | unchanged |
| `test_identity_shift_beta_is_floored_at_epsilon` | 1.0 | 1.00 | 1.00 | unchanged |

docs/developers_notes/08-hessian_tagging.md
```diff
@@ -640,6 +646,160 @@ Two cases are easy, because the diagonal partition refines every other: diagonal
...
+$$
+
+for every non-zero $g$.
+
+This strategy may perform several dense factorizations.
+It trades the deterministic cost of an eigendecomposition for the possibility that one or a small number of Cholesky attempts will suffice.
+
+### Eigenvalue modification
+
+The `eigh` strategy follows the eigenvalue modification in Nocedal and Wright, equation 3.43. For a symmetric Hessian,
```
BalzaniEdoardo, line 752: Suggest citing p. 50 and (3.49) instead of (3.43), no??

BalzaniEdoardo: The table's `cholesky` row lists the required information as "H certified positive (semi)definite as required by the resolved tag", but the code lets a user force Cholesky on an uncertified tag with a warning. The table should say what that path does (returns non-finite, reports non-convergence).

Also, one thing to decide rather than leave implicit: with a fixed absolute δ the method is **not** scale-equivariant, but it does still satisfy N&W's global-convergence hypothesis. The bounded modified factorization property (3.39) needs `κ(B_k) ≤ C` for a bounded Hessian sequence, and

$$\kappa(B) = \frac{\max(|\lambda|_{\max}, \delta)}{\delta}$$

is bounded whenever the Hessians are, so Theorem 3.8 applies. (The "bounded relative to the norm of the actual Hessian" language on p. 52 is about Gill–Murray modified Cholesky, not about the eigenvalue modification.) So the dtype fix is the one that matters; full scale-equivariance would need `δ = √u · max(1, ||H||)`, which is a separate decision worth stating in the note either way.

BalzaniEdoardo: N&W p. 52 also suggest a warm start: "We could choose the first nonzero shift τ₀ to be proportional to the final value of τ used in the latest Hessian modification."

The implementation recomputes τ₀ from the diagonal every outer iteration and throws away the τ that actually worked, even though the required shift is strongly correlated between consecutive iterates. Carrying the accepted τ in `NewtonState` and seeding τ₀ from it would remove most of the repeated factorizations. This one needs a new state field, so it is a design change rather than a constant

BalzaniEdoardo (changes requested):
here my latest comments. I think we are almost done with a few edits here and there

src/nemos/solvers/_newton.py
wulfdewolf: I solved this by doing this:
```
    if shift is not None:
        identity = lx.IdentityLinearOperator(operator.in_structure())
        shifted_operator = operator + shift * identity
        operator = lx.TaggedLinearOperator(
            shifted_operator,
            tags=tags,
        )

    return lx.linear_solve(
        operator,
        jax.tree.map(jnp.negative, grad),
        solver=solver,
        throw=throw,
    )
```
Let me know if that's not good enough, but it comes down to routing, since I also set:
```
        if resolved == "cholesky":
            self._linear_solver = lx.Cholesky()
            self._operator_tags = lx.positive_semidefinite_tag
            # Continue using the tag to distinguish the PSD and PD branches
            if matrix_property is MatrixProperty.POSITIVE_SEMI_DEFINITE:

                def _compute_shift(operator):
                    diagonal = lx.diagonal(operator)
                    return (
                        diagonal.size * jnp.finfo(diagonal.dtype).eps * diagonal.max()
                    )

                self._shift_fn = _compute_shift
            else:
                self._shift_fn = lambda _: None
```

src/nemos/solvers/_newton.py
wulfdewolf: I renamed it to linear_solver, and I renamed things internally to `_linear_solver_tag`, since we also have the actual `_linear_solver` object.

src/nemos/solvers/_newton.py
wulfdewolf: So, what I have was:
```
        if resolved == "cholesky" and matrix_property not in positive_properties:
            warnings.warn(
                "linear_solver='cholesky' was requested, but the Hessian tag "
                f"reports {matrix_property}. Cholesky generally requires a "
                "positive-definite or positive-semidefinite Hessian. Proceeding "
                "with Cholesky as requested; the solve may fail or return "
                "non-finite values.",
                RuntimeWarning,
                stacklevel=2,
            )
```

So if you explicitly ask for `"cholesky"`, we check the tag, and if not PD or PSD, we thrown this warning, but go ahead anyways. The solver might then throw an error.
If the tag is PD/PSD, we apply the shift or not.

docs/developers_notes/08-hessian_tagging.md
```diff
@@ -640,6 +646,160 @@ Two cases are easy, because the diagonal partition refines every other: diagonal
...
+$$
+
+for every non-zero $g$.
+
+This strategy may perform several dense factorizations.
+It trades the deterministic cost of an eigendecomposition for the possibility that one or a small number of Cholesky attempts will suffice.
+
+### Eigenvalue modification
+
+The `eigh` strategy follows the eigenvalue modification in Nocedal and Wright, equation 3.50 (p. 50). For a symmetric Hessian,
```
BalzaniEdoardo: I was mentioning 3.49, but I might have been wrong, double check to make sure that the ref is correct here.

docs/developers_notes/08-hessian_tagging.md
```diff
@@ -640,6 +646,160 @@ Two cases are easy, because the diagonal partition refines every other: diagonal
...
+0, & \min_i H_{ii}>0,\\
+-\min_i H_{ii}+\beta, & \text{otherwise},
+\end{cases}
+$$
+
+where the implementation floors $\beta$ at machine precision,
+
+$$
+\beta \leftarrow \max(\beta,\epsilon_{\mathrm{dtype}}).
+$$
```
BalzaniEdoardo, line 737: this seems out of date with the current code that uses: beta * max |diag|

tests/test_newton_glm.py
```diff
@@ -1235,112 +1168,236 @@ def _installed_newton(model, X, y):
...
+        "population_poissonGLM_model_instantiation",
+        "classifierGLM_model_instantiation",
+        "population_classifierGLM_model_instantiation",
+    }
+    expected = {
+        (fixture_name, regularizer_name)
+        for fixture_name in fixtures
+        for regularizer_name in ("Ridge", "UnRegularized")
+    }
+    actual = {(case.values[0], case.values[1]) for case in _LINEAR_SOLVER_CASES}
```
BalzaniEdoardo, line 1345: this compares two hand written lists, can we use a discovery like in the test below

BalzaniEdoardo: also mark every test that checks something like this (that all variants are tested) with the `@pytest.mark.metatest`

src/nemos/solvers/_newton.py
```diff
@@ -28,11 +29,86 @@ class NewtonState(eqx.Module):
...
+        solution = eqx.tree_at(
+            lambda result: result.value,
+            solution,
+            checked_value,
+        )
+
+    return solution
+
+
 class Newton(HessianMixin):
```
BalzaniEdoardo, line 112: we should add a class-level docstring with a comprehensive description of the variants.

BalzaniEdoardo (approved):
This basically is a done pr. 

I'll start approving this since the algorithm is solid, the implementation is clean.


The only substantial thing I would suggest to do is related to an issue I introduced :(

## On the shift update 

I asked for the accepted shift to be carried between iterations, following the remark on of Nocedal and Wright. The implementation turns the shift into a value that can only grow, that is worse than not carrying it at all, at least on my tests. The problem with it is that if tau grows large enough the algorithm converges to gradient descent (and a slow one with a linear solve...).

We can fix this in `Newton._solve`and it should cut total iterations. Across my benchmark by 31%, so I would implement it but the algo works as is already.

The proposed fix is

```python
min_diag = jnp.min(diag)
# N&W p.52 seed the ladder from the shift the last iteration accepted. Decaying it by
# the factor the ladder climbs by costs at most one extra factorization when the
# required shift is stable, and lets tau fall back to zero once the iterates reach a
# region where the unshifted Cholesky succeeds.
warm_start = jnp.where(
    previous_shift > beta,
    jnp.asarray(0.1, dtype=dtype) * previous_shift,  # scale it down of 1/10
    jnp.zeros((), dtype=dtype),
)
tau0 = jnp.maximum(
    jnp.where(min_diag > 0, jnp.zeros((), dtype=dtype), -min_diag + beta),
    warm_start,
)
```
Note that multiplying by 0.1 compensate for the growth of the tau due to the retry.

In the retry body

```
and in the retry body:

```python
new_tau = jnp.maximum(jnp.asarray(10.0, dtype=dtype) * tau, beta)
```


Try benchmark this yourself.

## Tolerance in absolute terms in tests

You applied the tightening to the three positive-definite tests. These are the ones left, I asked claude to make a list of them.

| file:line | test | asserted now | measured floor, f32 / f64 | suggested |
| --- | --- | --- | --- | --- |
| `test_newton_solver.py:267` and `:309` | `test_psd_singular_quadratic_preserves_null_space` and its autodiff twin, `x_opt` against `expected` | `atol=1e-4, rtol=0` | 6.9e-07 = 5.8 eps / 1.3e-15 = 5.8 eps | `rtol=100 * eps, atol=0.0` |
| `test_newton_solver.py:261` and `:303` | same tests, null component preserved | `atol=1e-4, rtol=0` | exactly 0 / exactly 0 | `rtol=100 * eps, atol=0.0` |
| `test_newton_solver.py:381` | `test_psd_quadratic_scale_equivariance` | `atol=1e-4, rtol=0` | 6.9e-07 = 5.8 eps / 4.5e-15 = 20.3 eps | `rtol=100 * eps, atol=0.0` |
| `test_newton_solver.py:256` and `:298` | same tests, residual against zero | `atol=1e-4, rtol=0` | 3.6e-07 / 4.4e-16 absolute | stays absolute, scaled like the PD tests: `atol=20 * eps * norm(b, inf), rtol=0.0` |
| `test_newton_solver.py:407` | `test_newton_linear_or_ridge_regression` | `np.allclose(atol=1e-5, rtol=0)` | x64 only: 3.0e-15 = 13.4 eps | `rtol=100 * eps, atol=0.0` |
| `test_newton_solver.py:1404` | `test_identity_shift_is_equivariant` | `atol=1e-10, rtol=1e-10` | absolute 3.5e-11, relative 1.0e-13 | `rtol=1e-12, atol=0.0` |
| `test_newton_glm.py:1347` and `:1348` | `test_every_linear_solver_reaches_the_same_fit` | `atol=1e-5` | 2.4e-16 = 1.1 eps | `rtol=1e-12, atol=0.0` |

Claude also says: 

`100 * eps` rather than the `20 * eps` used on the positive-definite tests because the `scale=1e-6` equivariance case sits at 20.3 eps and would fail at 20. It leaves a factor of 5 there and a factor of 17 everywhere else. Dropping `atol` to zero is safe at all of these: the smallest component of the semidefinite reference is 0.090 and of the regression references 0.25.

BalzaniEdoardo:
Great job @wulfdewolf

PR #615, opened 2026-08-20
Improve tagging & combine system for hessian.

BalzaniEdoardo:
Propose a tagging system that is leaf dependent. The tag includes:

- information about the leaves in which the loss curves (principal block of the hessian corresponding to leaf  strictly definite and constant in sign).
- information about the leaves in which the loss is flat (principal block of the hessian corresponding to leaf = 0).

With this tag it is possible to infer that a GLM that do not regularize an intercept, has a PositiveDefinite principal block corresponding to it intercept **if the link function guarantees convexity**. This is crucial to claiming positive definiteness when combined with a ridge that do not penalize the intercept.

The Regularizer Tagging system now requires both the parameters and the strength to create the tag. This because:

-`regularizable_subtree` defines which leaves are regularized, and which skipped. Every skipped leaf is a flat principal block in the hessian. 
- Any 0 in the regularization strength result in a non-defined block.  

Additionally, I added to the observation model an attribute specifying which links are guaranteed to keep the glm likelihood convex. This information is then read-off at the GLM class level, when the tag is requested, and used to assess if the intercept principal block is defined or not. 

The derivation of the tagging system properties can be found in the developer notes: `08-hessian_tagging.md`.

The tagging system cannot be refined without spectral information about the model design, which requires O(n^3) operation, the same as a general solve. So for, our purposes, this system is optimal, see `08-hessian_tagging.md`.

## Changes to Tagging System:

- `_hess.py` is moved to up in the folder structure since regularizer, solver and glm module needs it.
- `MatrixStructure` and `MatrixProperty` classes are replaced with Enums. 
- `MatrixProperty` dropped General, and has as floor, Symmetric (this is ok if we can assume our losses are at least C^2).
- `MatrixProperty` includes two new fields: `flat_on: PyTree[bool]`, `definite_on: PyTree[bool]`. This is marks which principal blocks are either flat (has the ridge regularization for the GLMParams where the intercept is not regularized). `definite_on` marks blocks that are definite (for example the coef block of the ridge penalization of the GLMParams).
- The combination logic follows the math described in `08-hessian_tagging.md`
- Model can implements `_resolve_hess_property` to infer on the fly the matrix property characterization for the hessian.
- `Regularizer` also implements `_resolve_hess_property`, used to determine the tag based on the reg strength and the model parameters (which informs the regularizer about which leaves are actually affected.
- `BaseRegressor`:  call `_resolve_hess_property` as part of the solver instantiation if solver is newton, and has a default implementation that provides a general tag.

## Question

Currently the leaf of the ridge is assumed non-definite if a zero is in the strength for that leaf. Should we set that the leaf is completely flat (null hessian) if the strength for that leaf is a zeros? Currently we are not checking that property.


docs/developers_notes/08-hessian_tagging.md
```diff
@@ -0,0 +1,664 @@
+(developers-hessian-tagging)=
+# Hessian tagging
+
+## Background
+
+A Newton step solves `H d = -g`, and how that linear system is best solved depends on what is known about `H` beforehand. A Cholesky factorization is the cheapest option but needs a positive definite matrix and fails on a singular one. A pseudo-inverse handles a singular matrix but costs more. A diagonal Hessian can be solved in time linear in the number of parameters. And if `H` has no sign guarantee at all, `-H^{-1} g` need not even point downhill, in which case a Newton step is the wrong thing to take and a first-order or quasi-Newton solver is a better choice.
+
+Checking any of this at run time means factorizing the matrix, which costs as much as the Newton solve itself. Under `jit` a failed Cholesky cannot be caught and retried either. So the information is carried alongside the Hessian instead, as a [`HessianTag`](nemos._hess.HessianTag) that each model and each regularizer declares about the term it contributes. The tags are combined when the terms are added, and the result picks the linear solver.
```
wulfdewolf: I think you found that the `jit` part isn't true no? maybe just drop it here.

PR #619, opened 2026-09-03
Move docpages

BalzaniEdoardo:
Start moving the docpages around with minimal change to the content.

The structure I am proposing is the following:
- A `User Guide` page, which overviews the package: each subpage will have with the background, then some worked out usage example with synthetic data.
- `Install`, `quickstart` and `getting help` are grouped in `getting started`
- How-to should have the short text + code-block copy-pastable examples.
- Tutorials are unchanged so far, but could be grouped by model.

I am opening a PR to a docs-upgrades branch that will be merged into development so that we can do the reviewing & merging of work-in-progress documentation in a non-blocking way. When the `docs-upgrades` are completed we can merge into `development`.


docs/getting_started.md
```diff
@@ -0,0 +1,45 @@
...
+:maxdepth: 3
+
+installation.md
+```
+
+:::
+
+## Quickstart
+
+Fit single-neuron, population and classifier GLMs on simulated data, and construct their features with the basis module.
```
billbrod, line 21:
```suggestion
Get an overview of nemos's features: how to fit models to data, how to use basis to construct features, and our compatibility with pynapple and scikit-learn. Then see the User Guide for more details! 
```

I think we want this to be a bit more "what's the goal of this page" and less specific about its contents.

docs/index.md
```diff
@@ -30,89 +28,263 @@ It streamlines the process of defining and selecting models, through a collectio
 
 The core of NeMoS includes GPU-accelerated, well-tested implementations of standard statistical models for systems neuroscience.
```
billbrod, line 29:
```suggestion
The core of NeMoS includes GPU-accelerated, well-tested implementations of standard statistical models for systems neuroscience. Our models are compliant with [scikit-learn's API](https://scikit-learn.org/stable/) and so are compatible with their pipelining and cross-validation machinery. Data can be represented as numpy array or [pynapple](https://scikit-learn.org/stable/) objects, to make working with neural data even easier.
```

Something to point this out front-and-center. maybe should even be more obvious, like adding cards with their logos? (we'd have to check if scikit-learn has a policy about that)

docs/user_guide/README.md
```diff
@@ -0,0 +1,122 @@
+# User Guide
+
+NeMoS has two core modules: `basis`, which builds model features from inputs such as position, phase, stimuli or spike counts, and the model classes, which relate those features to a measured response — spike counts, calcium traces, behavioral choices. Those two cover most analyses, and the chapters on them come first.
```
billbrod, line 3:
```suggestion
NeMoS has two core modules: 
- `basis`, which builds model features from inputs such as position, phase, stimuli or spike counts
- the model classes, which relate those features to a measured response — spike counts, calcium traces, behavioral choices.

Those two cover most analyses, and the chapters on them come first.
```

and add links to basis and model classes to their relevant headers.

I'm not sure what you mean by "cover most analyses"? Will be used by most users of nemos?

docs/user_guide/README.md
```diff
@@ -0,0 +1,122 @@
...
+
+For more advanced use, NeMoS exposes a set of components — observation models, regularizers, solvers — that you can combine to build your own model. Each component defines an interface, so you can supply your own if the ones implemented do not cover your use case.
```
billbrod, line 5:
```suggestion
For more advanced use, NeMoS exposes a set of components — observation models, regularizers, solvers — that you can use to customize our models and build your own. Each component defines an interface, so you can supply your own if the ones implemented do not cover your use case.
```

docs/user_guide/README.md
```diff
@@ -0,0 +1,122 @@
...
+
+The last chapters cover saving a fitted model, hyperparameter search with scikit-learn, and recordings too large to fit in memory.
```
billbrod, line 7:
```suggestion
The remaining sections cover saving a fitted model, hyperparameter search with scikit-learn, and recordings too large to fit in memory.
```

But I don't think we want an explicit list here, in which case we'll have to keep updating it. Something like "cover other useful topics, such as saving and interaction with other packages". to make it clear it's not an exhaustive list

docs/user_guide/README.md
```diff
@@ -0,0 +1,122 @@
...
+
+## Saving and Loading
+
+Writing a model to disk and reading it back.
+
+:::{card}
+
+```{toctree}
+:maxdepth: 2
+:titlesonly:
```
billbrod, line 83: need to add caption to show up in left sidebar

docs/user_guide/README.md
```diff
@@ -0,0 +1,122 @@
...
+
+## Stochastic optimization
+
+Fitting when the data does not fit in memory.
+
+:::{card}
+
+```{toctree}
+:maxdepth: 2
+:titlesonly:
```
billbrod, line 117: need to add caption to show up in left sidebar

docs/how_to_guide/README.md
```diff
@@ -1,64 +1,24 @@
 
 # How-To Guide
 
-Familiarize with NeMoS modules and learn how to take advantage of the `pynapple` and `scikit-learn` compatibility.
+Task-oriented recipes: each page answers a single "how do I ...?" question. The concepts behind them are covered in the [user guide](../user_guide/README.md).
 
 :::{dropdown} Additional requirements
```
billbrod, line 6: shojld be expanded by default

billbrod (changes requested):
Focusing just on the structure and various landing pages here, rather than content of the sub-pages
 
- I don't think the buttons should be the same as the top-level categories. For plenoptic, they're pages within "getting started" that I think people will need, plus the citation guide.
- On the landing page:
    - I like this layout!
    - Your model and examples cards aren't clickable -- they should be, right?
    - Would probably look nicer to turn Observation models and Regularizers into cards too, with a small visualization on top and then listing all the options
    - I think something on this page should refer to scikit-learn and pynapple. Maybe in the first paragraph at the top? I've added a suggestion.
- Getting started:
    - Either here or on the landing page (or both), I'd show that nemos can be installed simply with pip: `pip install nemos`, and then point to the install page for more details
    - probably not for right now, but install page should use tabs to separate Windows and Mac/Linux
    - Getting help is so short, I think it can probably just be included on the "getting started" page
- How to guide:
    - If you can, I'd change structure so that the headers and sub-pages show up in the left sidebar (using toctrees with captions) and hide the right sidebar. otherwise, the shift in structure when you go to any of these sub-pages is confusing, and the hierarchy isn't represented in their left sidebars.
    - Simulating spike trains is not about fitting models. Maybe change the header to "Using models"? or just "Models" and then have "Model selection" as a subtree within it?
    - I'm not sure the last three in "Performance and Scaling" belong there. They seem more like "Extending NeMoS" or something like that
- Tutorials
    - admonition about extra dependencies should be expanded by default
    - might be worth looking through the extra dependencies and seeing if they're all required for each of them? could split it into how-to and tutorials, or do it on a page-by-page basis? 
    - names for all of these can probably be reworked. might be useful to also tell people what their learning objectives are, e.g., would they learn something other than "how to fit nemos to this particular type of data"
- API guide: should have link to "developer notes" where it's referenced. In general, probably worth going through the docs and seeing if there are places where cross-references belong, because I've noticed several where I think they'd be helpful. (but probably a separate PR)
- I'm not sure "benchmarking" belongs as its own top-level header. Plenoptic has a [reference section](https://docs.plenoptic.org/docs/tags/2.1.0/reference/index.html), and in nemos's case, that could include benchmarking and citation. (and then link benchmarking in a button on the landing page, if you want to call it out)

docs/user_guide/README.md
BalzaniEdoardo: ok, makes sense

docs/user_guide/README.md
gviejo, line 1: I think the section `Interaction with other packages` is not very informative. People want to know how to do cross-validation or how to interact with temporal features. I propose this new way of sectioning the user guide by putting the part of pynapple in the feature design : 
Feature Design
- Spatial features
  - Simple stuff
  - Complicated stuff
  - More complicated stuff with Pynapple
- Temporal features
  - Convolutions
  - Multi-epochs
    - Pynapple compatibility (Pynapple handles multi-epochs automatically)
    - PyTree
- Basis addition
- Basis multiplication

and the part of cross-validation in a new section called cross-validation:
 
- Cross-validation and model selection
  - Transformers
  - Scikit-learn
  - Etc.

Issue #605, opened 2026-08-06
Guidance on solver choice for ElasticNet Poisson GLMs that hit `maxiter`


Akseli-Ilmanen:
## What I'm trying to do

I'm fitting per-session Poisson encoding GLMs to relate 3D kinematics and behavioural syllable labels to spike counts in a premotor/motor area in a carrion crow doing a tool-use task. The target analysis is essentially Fig. 4c–d of [Dhawale et al. 2021, *Nat Neurosci* 24:1256–1269](https://www.nature.com/articles/s41593-021-00889-3) — comparing held-out pseudo-R² across nested predictor sets (kinematics alone, behavioural label alone, both) to ask how much each feature set explains per unit.

<img width="600" height="400" alt="Image" src="https://github.com/user-attachments/assets/f1e5d689-5eeb-427c-a393-3bc3632ec70e" />

Screenshot of Fig. 4c–d of Dhawale et al. 2021

Concretely, an example session looks like:

- 25,000 bins @ 25 ms, 177 trials (each trial ~4-10s), 15-50 units
- predictors: position (3 axes), speed, head direction
- ~10 model "variants" (kinematics alone, syllable alone, kin+syl, and each kinematic signal/axis alone), all fit on one shared trial-grouped train/test split so their pseudo-R² are comparable
- `PopulationGLM` + `PoissonObservations`, ElasticNet (`l1_ratio=0.9`) like Dhawale et al. 2021, lambda chosen by `GridSearchCV` (8 lambdas × 3 folds)

Originally I binned the kinematic variables directly (stacking the sub-samples within each 25 ms bin as separate columns), similar to Dhawale et al. 2021. But after the NeMoS workshop in Barcelona, I moved to basis functions, which I much prefer conceptually:

```python
BASIS_CFG = {
    "position": {"kind": "bspline", "n": 15},   # one basis per axis
    "speed":    {"kind": "bspline", "n": 25},
    "angles":   {"kind": "cyclic",  "n": 25},   # heading, CyclicBSplineEval
}
```

## The problem: fits routinely hit `maxiter`

With `maxiter=5000`, `tol=1e-6`, `jax_enable_x64=True`, and `lambdas = np.logspace(-4, 1, 8)`:

- **8 of 8 variants** ended with the *CV-selected* fit having hit `maxiter` — so the reported coefficients and pseudo-R² were not at the optimum
- worse, because the unconverged fits still produce a CV score, the lambda **selection itself** was driven by fits that never converged
[...]

PR #622, opened 2026-09-09
Prox newton

wulfdewolf:


BalzaniEdoardo:
## Summary
This PR implements a proximal version of the Newton algorithm in response to issue #605. In particular, our `ProximalGradientDescent` was not performant on real dataset of medium size and was outperformed by C++ implementation of a coordinate ascent for elastic-net regression of glmnet. 

With this PR we propose a outer-inner loop proximal-Newton for non-smooth optimization, where in the inner loop:

- The smooth part of the loss is approximated by a quadratic Taylor expansion.
- The quadratic + non-smooth approximate loss is efficiently solved by FISTA.
- The proposed step is backtracked via Armijo within the FISTA search
- The steps are repeated until convergence. 

The proposed algorithm on data regime of issue #605 strongly outperformed glmnet.

## What's changed

**New `HessianMixin` class** that includes the hessian specific machinery, some of which previously lived under `Newton` some is new, and adds some convenience methods that both `Newton` and `ProxNewton` re-use. 

Moved machinery: `setup_hessian`, `_penalize_hessian`
New convenience methods: 
- `_resolve_linear_solver`: figures out the solver to use based on the tag.
- `_block_apply`: apply a function (linear solve or hessian-vector product) batching over vmappable dimension if needed.

**Edits to `Newton`** include moving methods to mixin, storing the unpenalized or penalized loss based on the `_proximal` tag, setting `y_diff` in the solver state forcing at least 1 iteration, extending the `_solve` signature to match that of `ProximalNewton`, added a `_converged` and `_line_search_input` method that must be re-implemented in `ProxNewton`.


**New `ProxNewton` class **.  This is the bulk of the implementation but it is not much more than gluing different pieces together of what we already had. In particular:

- `_hvp_block`: it is the pytree linear operator applying the hessian-vector product (hvp). 

- `_newton_direction`: this is the inner problem and minimizes, $\nabla f^\top (z - \beta) + \frac12 (z - \beta)^\top H (z - \beta) + P(z)$

  First, defines the unpenalized quadratic approx of the smooth part of the loss, vmapping the hvp. Then run a FISTA solve on the problem and return the proposed step.

- `_converged`: relies on the cauchy termination.

- `_line_search_inputs`: this method is remaps the inputs for the standard backtracking algorithm to  Tseng & Yun (2009) backtracking which works for losses of the form $F(z) = f(z) + P(z)$ where $f$ is the smooth part and $P$ is convex and potentially non smooth. If $d$ is the candidate direction, $a$ is the step size, the modified acceptance condition for the backtracking is:

    $$F(\beta + a \cdot d) \le F(\beta) + c \cdot a \cdot \Big[ \nabla f(\beta)^\top d + P(\beta + d) - P(\beta) \Big], \qquad c \in (0,1).$$

    However, optax's search takes three inputs — call them $v$ (`value`), $g$ (`grad`) and $\Phi$ (`value_fn`) — and internally does exactly

    $$\sigma = \langle d, g \rangle, \qquad \text{accept} \iff \Phi(\beta + a \cdot d) \le v + c \cdot a \cdot \sigma$$

    So by passing to opax the following:

    $$v = f(\beta) + P(\beta), \qquad \Phi = f + P, \qquad g = \nabla f + \frac{P(\beta+d) - P(\beta)}{\lVert d\rVert^2} \cdot d$$

    We recover exactly Tieng & Yun (2009).

- `lx.internal` instead of `optx._misc`. Replaced all the refs to `_misc` with the lineax public equivalents. 
- `BaseRegressor` ducktypes the hessian setup. based on the `_uses_hessian` flag.
- Added `penalty_fn` method to Regularizer which returns the penalty function applied to the filtered parameters.


## Question for reviewer

The descent direction is based on the assumption that the $P(z)$ non-smooth part is convex. The current backtracking it is very fast to compute, which makes it desirable. For all current regularizers $P$ is convex, so we can use this line search. 

In the future it may not work if we implement regularizers with non-convex $P$. The alternative we have in-house is to use the `fista_line_search` which is robust to non-convex `P` but is more costly. What should we do? Merge this PR as is, or already setup a mechanism for swapping line-searches based on the regularizer convexity property?

wulfdewolf (commented):
This LGTM!

In my opinion, we can hold off on adding support for swapping the linesearch till we actually have a non-convex regularizer.

PR #623, opened 2026-09-09
Classifier GLM-HMM

sjvenditto:
This PR adds a new class, `ClassifierGLMHMM`, for running the GLM-HMM on multi-class classification problems, similar to the GLM class `ClassifierGLM`.

Summary of changes:
- The `ClassifierGLMHMM` class was added in `glm_hmm/classifier_glm_hmm.py`. Similar to `ClassifierGLM`, it hardcodes the `Categorical` observation model and redefines relevant methods for label handling.
- A custom validator class `ClassifierGLMHMMValidator` class was added in `glm_hmm/validation.py` with updates to expected model parameter shapes and use of `ClassifierGLMValidator` as the internal `_glm_validator`
- the validator class parameter `n_states` was moved to the `extra_params` dictionary of `HMMValidator` to match how the `n_classes` parameter is defined in the `ClassifierGLMValidator`
- the `ClassifierMixin` class in `classifier_glm.py` was split into two classes, where methods specific to the GLM/PopulationGLM were moved into `ClassifierGLMMixin`, and model agnostic methods were kept in `ClassifierMixin`
- relevant methods in GLM-HMM's `algorithm_configs.py` have been given an additional input `is_categorical_glm` to differentiate how additional dimensions are vmaped (i.e. the categorical dimension is summed over, but the population axis is not)
- some tests in `test_glm_hmm_class_init.py` and `test_glm_hmm_validator.py` were parameterized to additionally check `ClassifierGLMHMM`
- new tests specific to `ClassifierGLMHMM` were added to `test_glm_hmm_classifier.py`


src/nemos/glm/classifier_glm.py
```diff
@@ -33,7 +33,7 @@
 
 
 class ClassifierMixin:
-    """GLM for classification."""
+    """Additional methods for classification models."""
```
BalzaniEdoardo, line 36: This class seems more general than GLM, why don't we move it one level up, in a new `_regressor_mixins.py` module?

src/nemos/glm/classifier_glm.py
BalzaniEdoardo, line 44: This mentions GLMs

src/nemos/glm_hmm/glm_hmm.py
```diff
@@ -622,6 +639,9 @@ def inverse_link_function(self, inverse_link_function: Callable):
         self._inverse_link_function = resolve_inverse_link_function(
             inverse_link_function, self._observation_model
         )
+        # if self._is_categorical_glm:
+        #     inverse_link = partial(inverse_link, axis=-2)
+        # self._inverse_link_function = inverse_link
```
BalzaniEdoardo, line 644: leftover

src/nemos/inverse_link_function_utils.py
```diff
@@ -14,18 +14,22 @@
...
+    if isinstance(func_exec, functools.partial):
+        func = func_exec.func
+    else:
+        func = func_exec
 
     @functools.wraps(func)
-    def wrapper(x):
-        return func(x)
+    def wrapper(x, **kwargs):
+        return func_exec(x, **kwargs)
```
BalzaniEdoardo, line 26: why not directly this?

```suggestion
    def wrapper(*args, **kwargs):
        return func_exec(*args, **kwargs)
```

tests/test_glm_hmm_classifier.py
```diff
@@ -0,0 +1,297 @@
...
+import jax
+import jax.numpy as jnp
+import numpy as np
+import pytest
+
+from nemos.glm import ClassifierGLM
+from nemos.glm_hmm import GLMHMM, ClassifierGLMHMM
+
+
+class TestClassifierGLMHMM:
```
BalzaniEdoardo, line 15: should we add a multiple epoch case to these tests?

BalzaniEdoardo (changes requested):
In general this is looking good. I have been playing around with the class and I found out the following open issues:

- The model cannot be re-loaded after saving

    ```python
    import nemos as nmo
    import numpy as np
    import jax.numpy as jnp
    import jax

    T = 400
    X = np.random.randn(T, 2)
    y = np.random.choice(["a", "b", "c"], T)
    model =  nmo.glm_hmm.ClassifierGLMHMM(n_states=4, n_classes=3)
    model.set_classes(y)
    params = model.initialize_params(X, y)
    model._set_model_params(model._validator.to_model_params(params))
    ysim, logp, gt_state = model.simulate(jax.random.PRNGKey(0), X)

    model.fit(X, ysim)
    model.save_params("cl_glm_hmm.npz")

    recovered = nmo.load_model("cl_glm_hmm.npz")
    ```

    results in a `KeyError` because the model is not in the model registry of the IO. `MODEL_REGISTRY` in `io.py` is hand-written and has no coverage guard, so a new regressor is silently absent rather than failing. This means we are missing an io test for it, and a metatest asserting the registry covers every concrete `BaseRegressor` subclass (derivable exactly — `ClassifierGLMHMM` is the only current miss). Everything else round-trips correctly once the entry is added, `classes_` included. Probably is worth coding a self-discovery of models in the io.py, so that the registry is auto populated.

- `compute_loss` raises

    ```python
    model = nmo.glm_hmm.ClassifierGLMHMM(n_states=4, n_classes=3).fit(X, y)
    params = (model.coef_, model.intercept_, model.scale_, model.initial_prob_, model.transition_prob_)
    model.compute_loss(params, X, y, np.zeros(T, dtype=bool))
    ```

    `BaseRegressor.compute_loss` preprocesses (`base_regressor.py:810`) and `BaseHMM._compute_loss` preprocesses again (`hmm/hmm.py:688`). That was harmless while `_preprocess_inputs` only dropped NaNs; `ClassifierMixin._preprocess_inputs` one-hots, so `y` goes `(T,)` → `(T,3)` → `(T,3,3)` and `is_population_glm` flips True. `compute_loss` is missing from both the `test_must_set_classes_before_calling` list and the label-invariance tests.

- A GLM-HMM accepts `Categorical` and dies in the vmap

    ```python
    nmo.glm_hmm.GLMHMM(n_states=2, observation_model="Categorical").fit(X, y)
    ```

    `GLM` rejects this through `_invalid_observation_types` / `_validate_observation_class` with a message pointing at `ClassifierGLM`; the GLM-HMM setter never calls it. The reverse direction is silent:

    ```python
    model = nmo.glm_hmm.ClassifierGLMHMM(n_states=2)
    model.observation_model = "Poisson"
    ```

    gives a `PoissonObservations` with `_is_categorical_glm=False`. Rather than a new check, we can reuse what `GLM` already has: set `_invalid_observation_types` on `GLMHMM` and `ClassifierGLMHMM`, and call `_validate_observation_class` from the GLM-HMM setter. No test either way.

- Any `kmeans` initializer fails on the classifier

    ```python
    model = nmo.glm_hmm.ClassifierGLMHMM(n_states=4, n_classes=3)
    model.setup(glm_params_init="kmeans")
    model.fit(X, y)
    ```

    raises `TypeError: The CategoricalObservations observation type is not supported for PopulationGLM models`. The kmeans setup builds one `PopulationGLM` per state (`glm_hmm/initialize_parameters.py:182`) with the model's own observation model, and `PopulationGLM` rejects `Categorical`. All four of `glm_params_init`, `scale_init`, `initial_proba_init` and `transition_proba_init` go through that same setup and fail identically; the same call works on `GLMHMM`. Untested, since `TestGLMHMMSetup` and the kmeans tests are GLMHMM-only.

- `dof_resid_` is wrong for the classifier

    ```python
    nmo.glm_hmm.ClassifierGLMHMM(n_states=4, n_classes=3).fit(X, y).dof_resid_
    ```

    returns `[365., 365., 365.]` — length `n_classes`, because `_estimate_resid_degrees_of_freedom` reads `coef_leaf.shape[1]` as `n_neurons` (`glm_hmm.py:881`), and it counts `X.shape[1] * n_states` coefficients, ignoring the class axis and the `n_classes - 1` simplex correction that `ClassifierGLMMixin` applies. Nothing consumes `dof_resid_` in the HMM path, so this is a wrong public attribute rather than a wrong fit.

- The base `GLMHMMValidator` dimensionality message now reads `scale must be of shape (n_classes, n_states)` — copy-paste leak .

- `ClassifierGLMHMMValidator.get_empty_params` returns `jnp.empty(n_states)` for scale, while `expected_param_dims` demands ndim 2 and `fit` produces `(n_classes, n_states)`. Harmless today since `Categorical` is in `_NO_SCALE`, but `get_empty_params` is the GroupLasso mask path, and the new `test_log_scale_shape` asserts `(N_STATES,)` for both branches, which pins the inconsistency.

- `_make_wrapper`'s `isinstance(func_exec, functools.partial)` branch has no caller (it is the one line codecov flags). It can go along with the `*args` simplification.

## Missing tests

- io: save/load round-trip for `ClassifierGLMHMM` (`TestGLMHMMSaveLoad` is GLMHMM-only), plus the registry metatest above.

- pynapple: untested. It does work — `fit`/`score`/`smooth_proba`/`filter_proba`/`decode_state` accept `TsdFrame`/`Tsd`, the last three return `TsdFrame`, `simulate` returns raw arrays as `GLMHMM` does.

- pytree `X`: untested. Also works — `fit`, `smooth_proba`, `simulate`, `initialize_params`.

    Both are invisible because `MODEL_REGISTRY` / `instantiate_base_regressor_subclass` in `conftest.py` don't know the class and `test_glm_hmm_class.py` hardcodes `{"model": "GLMHMM"}`. `ClassifierPopulationGLM` is missing from that registry too. Nothing guards it, which is the same hole as the io one.

- `test_glm_hmm_validator.py`: flipping the five `@pytest.mark.parametrize("is_classifier", [False])` classes to `[False, True]` gives 114 passing and 2 failures, both in `TestValidateConsistency`, which hardcodes GLMHMM shapes — `bad_coef` wants `(NF+1, NC, NS)` and `bad_scale` wants `(NC, NS+1)`, and both then raise as expected. `TestValidateAndCastFeatureMask` is the one worth having, since it now delegates to `ClassifierGLMValidator`.

- `test_glm_hmm_class_init.py`: `TestGLMHMMInit`, `TestGLMHMMSetup`, `TestGLMHMMCheckModelIsFit`, `TestGLMHMMSklearnClone` are not parametrized over the classifier. `clone`, `set_params(n_states=, n_classes=)` and `repr` are all correct by hand, but the `_set_n_states`-before-`n_classes` ordering in `__init__` is matters (the `n_classes` setter reaches `_get_validator_extra_params`, which needs `_n_states`).

- `_preprocess_inputs` moved above the initialization block in `fit`, so the initializers now get the NaN-filtered data instead of the raw input. The classifier needs this, because that call is also where `y` gets one-hot encoded — under the old order its initialization fails with a broadcast error. But it changes plain `GLMHMM` too: the default initializers return different values under the two orders, and only `update` has NaN tests. One assertion covers it — `fit(X_with_nan_rows, y)` must give exactly the same result as `fit(X[k:], y[k:])`, which it now does.

- Solvers: `Newton`, `ProximalGradient`+Lasso and `GroupLasso` all run with `ClassifierGLMHMM`, none is tested. `_hess_leaf_claims` sits on `ClassifierMixin` now and is reachable through Newton.

- `is_categorical_glm=True` never appears in an `algorithm_configs` unit test — every call site in the diff passes `False`, so the new axis arithmetic is only exercised end-to-end. `state_axes = 3` is unreachable without a `ClassifierPopulationGLMHMM`, so `1 + True + True` is dead arithmetic worth a note.

PR #624, opened 2026-09-14
Associative Forward-Backward and Max-Sum

BalzaniEdoardo:
## Summary

In this PR the forward-backward and max-sum algorithms are fully parallelized via associative scan.

The algorithm is described in full in `docs/developers_notes/09-associative_estep_hmm.md` including the math. 

The signature of the associative scan function are matched to their sequential counterpart, and the algorithm can be selected by indexing into a dictionary pointing to the function defs.

## New Functions

- `_forward_pass_assoc`: implements the algorithm described in `The Scan, Step by Step` section of the `09-associative_estep_hmm.md` note but in log space.
- `_backward_pass_assoc`:  implements the algorithm described in `The Backward Pass` section of the `09-associative_estep_hmm.md` note but in log space.
- `forward_backward_assoc`: similar the forward-backward sequential just calling the associative func instead.
- `max_sum`: implements the algorithm described in `The MAP Path in the Same Shape` section of the `09-associative_estep_hmm.md`.
- Dictionaries defined in hmm.py map between the algorithms:
    ```python
    FORWARD_PASS = {
        "sequential": forward_pass,
        "associative": forward_pass_assoc,
    }
    
    FORWARD_BACKWARD = {
        "sequential": forward_backward,
        "associative": forward_backward_assoc,
    }
    
    MAX_SUM = {
        "sequential": max_sum,
        "associative": max_sum_assoc,
    }
    ```
- New parameter and attribuite `estep_type` with value "sequential" or "associative" that switches between algorithms.

## Tests
Added the algorithm type as a parametrization to existing tests to verify correctness, jit compilation behavior, pynapple and pytree compatibility.

Added tests that make sure that the `BaseHMM` methods are calling the correct algorithm for a set flag, that the flag setter raises correctly for unrecognized settings, and that the `GLMHMM` class calls the mixin super method for the estep.


src/nemos/hmm/hmm.py
sjvenditto, line 122: Missing docs for `estep_type` here

src/nemos/hmm/parallel_expectation.py
```diff
@@ -0,0 +1,710 @@
...
+        Normalized log forward messages, shape ``(n_time_bins, n_states)``.
+        Entry ``[t, k]`` is the log filtered probability log p(z_t=k | y_1:t).
+        Each row is normalized: ``exp(log_alphas[t]).sum() == 1``.
+    log_normalizers :
+        Array of shape ``(n_time_bins,)`` containing the log-normalization constants at each
+        time step. The sum of these values gives the log-likelihood of the sequence.
+
+    See Also
+    --------
+    :func:`~nemos.hmm.forward_backward` : Computes both forward and backward messages for smoothing.
```
sjvenditto, line 305:
```suggestion
    :func:`~nemos.hmm.forward_backward_assoc` : Computes both forward and backward messages for smoothing.
```

src/nemos/hmm/parallel_expectation.py
```diff
@@ -0,0 +1,710 @@
...
+    and
+
+    .. math::
+        \hat{\beta}_{t-1} = N_t N_{t+1} \cdots N_{T-1} \mathbf{1},
+
+    a suffix product of matrices, which is the scan map of section **Scan Map** of the
+    note run in reverse.
+
+    Two things differ from the forward pass, both noted in the closing paragraph of the
+    note, which does not derive this pass. The messages are not normalized, so the
```
sjvenditto, line 394: What note is this referring to? The page for the developers notes has a section on the backward pass, is that not deriving this?

src/nemos/hmm/parallel_expectation.py
```diff
@@ -0,0 +1,710 @@
+r"""Parallel implementation of the forward-backward algorithm.
+
+The recursions of :mod:`nemos.hmm.expectation_maximization` are prefix problems over a
+sequence of per-step transfer matrices, so they can be evaluated by
+``jax.lax.associative_scan`` at depth :math:`O(\\log T)` rather than by a scan of length
```
sjvenditto, line 5:
```suggestion
``jax.lax.associative_scan`` at depth :math:`O(\log T)` rather than by a scan of length
```

src/nemos/hmm/parallel_expectation.py
```diff
@@ -0,0 +1,710 @@
...
+
+
+def log_matmul(log_earlier: Array, log_later: Array) -> Array:
+    r"""Log-semiring product of two segments, earlier on the left.
+
+    :math:`\text{out}[i,j] = \text{logsumexp}_k(\text{earlier}[i,k] + \text{later}[k,j])`,
+    so ``i`` indexes the entry state of the earlier segment and ``j`` the exit state
+    of the later one, with the shared boundary state ``k`` contracted away.
+
+    Shifted per row of ``log_A`` and per column of ``log_B``, then run as a GEMM. The
```
sjvenditto, line 60:
```suggestion
    Shifted per row of ``log_earlier`` and per column of ``log_later``, then run as a GEMM. The
```

src/nemos/hmm/parallel_expectation.py
```diff
@@ -0,0 +1,710 @@
...
+    Conditioning step described in the notes docs/developers_notes/09-associative_estep_hmm.md
+
+    Parameters
+    ----------
+    log_exit_weights:
+        Log of either :math:`p(y_t | z_t=i)` (when computing initial elements) or
+        :math:`p(y_{v:t} | z_{v-1}=k)` (during the scan). Shape (n_samples, n_states).
+    log_row_stochastic:
+        Log of either :math:`p(z_t=i | z_{t-1}=j)` (when computing initial elements) or
+        :math:`p(z_{v-1}=k | z_{u-1}=j, y_{u:v-1})$` (during the scan). Shape (n_samples, n_states, n_states).
```
sjvenditto, line 87:
```suggestion
        :math:`p(z_{v-1}=k | z_{u-1}=j, y_{u:v-1})` (during the scan). Shape (n_samples, n_states, n_states).
```

src/nemos/hmm/parallel_expectation.py
```diff
@@ -0,0 +1,710 @@
...
+
+
+def combine_forward(
+    earlier: Tuple[Array, Array], later: Tuple[Array, Array]
+) -> Tuple[Array, Array]:
+    r"""Combine in the associative scan.
+
+    The combination implements the :math:`\oplus` operator described
+    in section **Get (log(l), L) via scan**  of the note
+    the notes docs/developers_notes/09-associative_estep_hmm.md.
```
sjvenditto, line 123:
```suggestion
    docs/developers_notes/09-associative_estep_hmm.md.
```

src/nemos/hmm/parallel_expectation.py
```diff
@@ -0,0 +1,710 @@
...
+    return log_l, log_L_unnorm - log_l[..., jnp.newaxis]
+
+
+def combine_forward(
+    earlier: Tuple[Array, Array], later: Tuple[Array, Array]
+) -> Tuple[Array, Array]:
+    r"""Combine in the associative scan.
+
+    The combination implements the :math:`\oplus` operator described
+    in section **Get (log(l), L) via scan**  of the note
```
sjvenditto, line 122:
```suggestion
    in section **Get (log(l), L) via scan** of the note
```

src/nemos/hmm/parallel_expectation.py
```diff
@@ -0,0 +1,710 @@
...
+        `log_initial_prob` and `log_transition_prob`, as well as the attribute `model_params`, also a
+        `ModelParams` subclass containing model-dependent parameters used in the log-likelihood function.
+
+    log_likelihood_func :
+        Function computing the elementwise log-likelihood of observations.
+        Must return an array of shape ``(n_time_bins, n_states)``.
+
+    session_starts :
+        Boolean array marking the start of a new session.
+        If unspecified or empty, treats the full set of trials as a single session.
```
sjvenditto, line 481: do we allow session starts to be empty?

src/nemos/hmm/parallel_expectation.py
```diff
@@ -0,0 +1,710 @@
...
+        Marginal log-posterior distribution over latent states, shape ``(n_time_bins, n_states)``.
+
+    log_joint_posterior :
+        Joint log-posterior distribution between consecutive time steps summed
+        over samples, shape ``(n_states, n_states)``.
+
+    log_likelihood :
+        Total log-likelihood of the observation sequence under the model.
+
+    log_likelihood_norm :
```
sjvenditto, line 495: the returned variable for `log_likelihood_norm` is not a log

src/nemos/hmm/parallel_expectation.py
```diff
@@ -0,0 +1,710 @@
...
+
+import jax
+import jax.numpy as jnp
+
+from ..typing import EStepOutput, LogLikelihoodFn, ModelParamsT
+from .expectation_maximization import compute_xi_log
+from .utils import Array
+
+
+def log_matmul(log_earlier: Array, log_later: Array) -> Array:
```
sjvenditto, line 53: This function (as well as `combine_forward` and `max_plus_matmul`) are never exported. should they be formatted as private functions?

sjvenditto (changes requested):
I've left some minor comments throughout the code. A "larger" bug (which I have in quotes because I don't know the likelihood of running into this in practice) is that the implementation of `log_matmul` returns `nan` in cases it should return `-inf` when a column is all 0s, e.g.
```
A = jnp.array([[0.9, 0.1, 0.0],
               [0.1, 0.9, 0.0],
               [0.1, 0.9, 0.0]], dtype=jnp.float32)
logA = jnp.log(A)

ours = log_matmul(logA, logA)
ref = jax.scipy.special.logsumexp(logA[:, :, None] + logA[None, :, :], axis=-2)

print("log_matmul:\n", np.asarray(ours))
print("\nlogsumexp reference:\n", np.asarray(ref))
```
returns:
```
log_matmul:
 [[-0.1985 -1.7148     nan]
 [-1.7148 -0.1985     nan]
 [-1.7148 -0.1985     nan]]

logsumexp reference:
 [[-0.1985 -1.7148    -inf]
 [-1.7148 -0.1985    -inf]
 [-1.7148 -0.1985    -inf]]
```
This leads to nan coefficients in the model fit (where the sequential fit does not fail). We will need to clamp the non-finite shifts (which is what [jax does in `logsumexp`](https://github.com/jax-ml/jax/blob/78745436233777179cd4bda8378ca33e52c52902/jax/_src/ops/special.py#L87)). something like:
```
row_max = jax.lax.stop_gradient(jnp.where(jnp.isfinite(row_max), row_max, 0.0))
col_max = jax.lax.stop_gradient(jnp.where(jnp.isfinite(col_max), col_max, 0.0))
```

src/nemos/hmm/parallel_expectation.py
BalzaniEdoardo: this was stale, i added the derivaiton so nothing is missing

src/nemos/hmm/parallel_expectation.py
BalzaniEdoardo: I think we always set it in the model, so the guard is probably useless at the moment. 

I kept it in, since it would be compiled away but let me know if you think i should drop this entirely

sjvenditto (approved):
Looks good!

PR #625, opened 2026-09-17
Ruff is all you need

wulfdewolf:
This switches out `black`, `isort`, and `flake8` for `ruff`.

I tried to keep the old settings as much as possible. 
I updated `pyproject.toml`, `.pre-commit-config.yaml`, and `tox.ini` to use `ruff`.

I then ran the new setup on all files.
Initially, `ruff` complained about a lot, so I kept ignoring rules until I got to a point where the only complaints where auto-fixable.
Those auto-fixes are the changes you can see in the files.

I copied most settings from Plenoptic


BalzaniEdoardo (commented):
this looks sensible to  me, I would have @billbrod take a look before merging

pyproject.toml
```diff
@@ -183,20 +151,87 @@ exclude_lines = [
...
+    "C901",  # Complex tests and fixtures.
+    "E731",  # Lambda assignments as compact test callables.
+    "E741",  # Short mathematical variable names.
+]
+"scripts/**" = [
+    "E501",  # Long parametrizations and expected messages.
+    "C901",  # Complex tests and fixtures.
+    "E731",  # Lambda assignments as compact test callables.
+    "E741",  # Short mathematical variable names.
+]
```
BalzaniEdoardo, line 223: why C901 is in the scripts?

wulfdewolf: because there are scripts with complexity > 10, the comment behind it isn't accuarte, I copied it from the one above.
C901 checks for function complexity, I just pushed a more correct comment for it in the scripts block

pyproject.toml
```diff
@@ -183,20 +151,87 @@ exclude_lines = [
...
+    # complexity check
+    "C90"
+]
+ignore = ["D420", "D421", "D401", "F722"]
 
+[tool.ruff.lint.mccabe]
+max-complexity = 10
+
+[tool.ruff.lint.pycodestyle]
+max-line-length = 120
```
billbrod, line 205: how does this interact with `line-length = 88` set under `tool.ruff`?

pyproject.toml
```diff
@@ -183,20 +151,87 @@ exclude_lines = [
...
+
+[tool.ruff.lint.flake8-annotations]
+allow-star-arg-any = true
+suppress-none-returning = true
+
```
billbrod:
```suggestion
```

you don't have `ANN` activated, so these aren't doing anything

pyproject.toml
billbrod: I think it's worth thinking about whether those scripts should be simplified. Same with tests

pyproject.toml
```diff
@@ -183,20 +151,87 @@ exclude_lines = [
...
+]
+"scripts/**" = [
+    "E501",  # Long parametrizations and expected messages.
+    "C901",  # Complex scripts.
+    "E731",  # Lambda assignments as compact test callables.
+    "E741",  # Short mathematical variable names.
+]
+
+# Exercise notebooks may reference names defined in skipped cells.
+"exercises*.ipynb" = ["F821"]
```
billbrod:
```suggestion
```

doesn't do anything for you -- this is for a specific subsection of plenoptic's documentation

billbrod (approved):
Looks good to me, though I'd suggest adding EVEN MORE CHECKS. Here are the ruff rules plenoptic has that you're aren't using: 
- [pyupgrade](https://docs.astral.sh/ruff/rules/#pyupgrade-up), to make sure code is using modern features
- [flake8-simplify](https://docs.astral.sh/ruff/rules/#flake8-simplify-sim), lots of rules to simplify code
- various documentation-related ones: [flake8-annotate](https://docs.astral.sh/ruff/rules/#flake8-annotations-ann), [pydoclint](https://docs.astral.sh/ruff/rules/#pydoclint-doc): DOC201, DOC202, DOC402, DOC403, DOC501 (those all have to be explicitly added)
- [RUF027](https://docs.astral.sh/ruff/rules/missing-f-string-syntax/) for f-strings

based on what Wolf said, those will require more effort to fix, so if you want them, you could add them one at a time in PRs that focus on fixing any failures that arise.

I have several other checks as part of my pre-commit that I'd recommend considering as well (though that should be in extra PRs, like above). I think the first 2 should definitely be done, the others are more if useful for you.
- [numpydoc](https://numpydoc.readthedocs.io/en/latest/validation.html). Ruff's documentation rules alone don't ensure well-formatted docstrings, I had to add numpydoc as well.
- ruff can lint/format ipynb files, but not myst markdown notebooks. I have a [small custom script](https://github.com/plenoptic-org/plenoptic/blob/main/linting/ruff_notebooks.py) that uses jupytext to convert the markdown notebooks to ipynb, runs ruff on them, and converts them back / raises errors if there any problems. (which is part of my pre commit) 
- I have a pair (`linting/check_magic_functions.py` and `linting/fix_tutorial_semicolons.py`) which together ensure that all plots in all notebooks only show up once, avoiding the duplicate plot issue.
- `linting/check_return_literal.py` makes sure there are no Literal return annotations, which prevents tab-completion (see https://github.com/plenoptic-org/plenoptic/pull/463).
- I have a couple that check for likely misformatted rest / markdown directive: `linting/check_markdown.py`, `linting/check_sphinx_directives.py`

happy to talk through any of those. the ones I wrote myself are less flexible/general, but should be usable without too much work.

pyproject.toml
wulfdewolf: Good question, so I had to add this because ruff doesn't format docstring lines.
So what this does:
- ruff formats everything it usually formats to 88
- ruff doesn't complain if lines are longer than 120

This matches the setup NeMoS had before.
I agree we'd ideally make everything 88, and address the long docstrings/comments.

pyproject.toml
wulfdewolf: got it will remove, my bad

PR #627, opened 2026-09-18
add flag to build full api locally

sjvenditto:
Add a flag in `conf.py` that lets you build docs locally that includes the entire API. This lets us check the formatting of docstrings for private functions that have useful details for developers, but never gets rendered in the full docs.

For example, this can be built like:
```
NEMOS_FULL_API=true EXCLUDE_TUTORIALS=true sphinx-build -b html docs docs/_build/fullapi
```


BalzaniEdoardo (approved):
great, thanks!

